Issue #22, opened 2017-05-04
Syncs should return CompletetionStages as results instead of blocking.


heshamMassoud:
Syncs should return CompletetionStages as results instead of blocking.

- [x] for services
- [x] for sync results

PR #37, opened 2017-05-10
Use CompletionStage instead of blocking in BaseSync and CategoryService

heshamMassoud:
Addresses issue #22:
1.  Use CompletionStage instead of blocking in BaseSync and CategoryService...
2. Gracefully handle Futures containing Exceptions.
3. Add some missing error messages as enums.


src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -89,12 +96,13 @@ protected void processDrafts(@Nonnull final List<CategoryDraft> categoryDrafts)
...
         }
+        return CompletableFuture.completedFuture(this.statistics);
     }
 
     @Override
-    protected void process(@Nonnull final List<Category> resources) {
+    protected CompletionStage<CategorySyncStatistics> process(@Nonnull final List<Category> resources) {
         //TODO: SEE GITHUB ISSUE#12
-
+        return CompletableFuture.completedFuture(this.statistics);
```
jakubchrobak: The `CategorySync` would be adjusted further in another PR right?

heshamMassoud: yes

src/main/java/com/commercetools/sync/services/CategoryService.java
```diff
@@ -7,16 +7,51 @@
...
-    Category createCategory(@Nonnull final CategoryDraft categoryDraft);
+    /**
+     * Given a {@link CategoryDraft}, this method creates a {@link Category} based on it in the CTP project defined in
+     * a potentially injected {@link io.sphere.sdk.client.SphereClient}. This method returns
+     * {@link CompletionStage&lt;Category&gt;} in which the result of it's completion contains an instance of the
+     * {@link Category} which was created in the CTP project.
+     *
+     * @param categoryDraft the {@link CategoryDraft} to create a {@link Category} based off of.
+     * @return {@link CompletionStage&lt;Category&gt;}containing as a result of it's completion an instance of the
+     *      {@link Category} which was created in the CTP project.
```
heshamMassoud: 0dec077

src/main/java/com/commercetools/sync/commons/enums/Error.java
```diff
@@ -4,15 +4,26 @@
 import javax.annotation.Nonnull;
 
 /**
- * Represents all error messages across the sync update action build modules.
+ * Represents all error messages across the sync modules.
  */
 public enum Error {
```
jakubchrobak: What is the gain from having all update messages in one place as public enum anyway? They are used always as ENUM.getDescritpion() inside format for calllback message. Why they're not kept as `private final static String` in context they belong to? Am I missing something?

heshamMassoud:
@andrii-kovalenko-ct if you have time, can you please take a look at this PR.

src/main/java/com/commercetools/sync/commons/enums/Error.java
heshamMassoud: Actually, when I come to think of it again, it's actually better to use constant `Strings` instead of Enums in this case. Because most of them are _format_ `Strings` and if they are used as enums (`ENUM.getDescription()`) then the IDE will not be able to show warnings if the number of arguments don't match as opposed to directly using the constant String. 

Also it's not the best idea to keep all strings in one class (enum in this case), so if it get's really big and confusing. So I refactored them to be at the Utils where they are only needed. 	

29a1098

Issue #44, opened 2017-05-19
Remove unneeded info log message at start of sync 


heshamMassoud:
### Description
Remove unneeded info log message at start of sync.

- As a user of the sync libraries, I don't want the sync to internally forcibly log the start. I can log (or not) the start of it from where I call the sync from within my application.

### Expected Behavior
The sync library should not log anything out. The users can depend on the callbacks (warning and error for now) for logging.

### Current Behavior
The sync library currently logs the start of the sync. It's not only needed, but it will also cause this info log to be printed out on every sync batch, which doesn't make sense.

### Possible Solution
Remove the info log at all or replace with a new `infoCallBack`. For now the former solution, is enough.

PR #52, opened 2017-05-22
44 remove info log, project key and CtpClient Wrapper

heshamMassoud:
#### Summary
The sync library should not log anything. The user of the library should be responsible for logging the start of sync. The log message at the start was also wrong because if the user streams data and syncs in batches, it will log the start on every batch processing which is wrong.

#### Description
- Removes the log message at the start of the sync.
- Adds constants for error messages at the inventory sync.
- Removes all project keys in the error messages. This can be appended from outside the library using the callbacks from the user.
- Removes the CtpClient wrapper which was used to combine SphereClient and ClientConfig in one class. The client config was needed to get the project key. Now not needed anymore. 

#### Relevant Issues
#44 

#### Hints for Review
- `BaseSync.java` contains the removal of the log message at the start of sync.
- `InventorySync.java` contains the change of removal of project keys from error messages and also using constant strings for messages. 
- `BaseSyncOptions.java` removes the helper method to get the projectkey from the client.
- Rest of changes are changes because of replacing `CtpClient` to just JVM SDK `SphereClient`.


src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -410,24 +417,19 @@ private InventoryEntryDraft withSupplyChannel(@Nonnull final InventoryEntryDraft
...
-     * callback specified in the {@code syncOptions}. <br/>
-     * <strong>NOTE:</strong> to the end of {@code message} following phrase will be joined:
-     * <pre> " in CTP project with key '_KEY_'"</pre>
-     * where _KEY_ is replaced by CTP project key, taken from CTP client configuration.
+     * callback specified in the {@code syncOptions}.
      *
      * @param message the reason of failure
      * @param exception the exception that occurred, if any
      */
     private void handleFailure(@Nonnull final String message, @Nullable final Throwable exception) {
```
jakubchrobak: Then this method is not necessary anymore.

heshamMassoud: true

heshamMassoud: @jakubchrobak 4dfa3ba

Issue #39, opened 2017-05-11
null CategoryDrafts are not counted as `processed`


heshamMassoud:
#### Current Behaviour
Currently, if one of the input CategoryDrafts is `null` it won't be counted as `processed`.

This can be found here: https://github.com/commercetools/commercetools-sync-java/blob/develop/src/main/java/com/commercetools/sync/categories/CategorySync.java#L81-L82

#### Expected Behaviour
It, however, should be counted as `processed`. 

#### Proposed Solution
Either count it as `failed` or have a new counter `invalidResource`.

heshamMassoud:
@jakubchrobak Please make sure the inventory sync also handles this.

butenkor:
The idea is that summary report on: `sync([cat1,null,cat2,cat3,cat5])` looks like that:
```
{"processed": 5, "created": 2, "updated": 2, "failed": 1}
```

PR #53, opened 2017-05-22
39 null categories are processed

heshamMassoud:
#### Summary
Currently, if one of the input CategoryDrafts is `null` it won't be counted as processed. However, should be counted as processed.

#### Description
The idea is that summary report on: sync([cat1,null,cat2,cat3,cat5]) looks like that:
`{"processed": 5, "created": 2, "updated": 2, "failed": 1}`

#### Relevant Issues
#39 

#### Todo

- Tests
    - [x] Unit


heshamMassoud:
@piobra IMO, what is important is, if there is a `null` input the sync process shouldn't stop. It also shouldn't continue on error. However, that's where callbacks come into play -> It is up to the user of the library how he wants to proceed if such an incident occurs, the library should notify him with this event, and accordingly he reacts with his desire callback. We could instead of passing a null exception and this message, just try and catch the NPE and then pass the NPE to the errorCallback. However, in such case, just the simple message `CategoryDraft is null.` is more than enough. It's also removes the need to add the try and catch NPE code. WDYT?

piobra:
@heshamMassoud I see. I don't like NPE try-catch either so I think it's ok.
It's rather an architectural problem. Whether to accept those null's or not in that case.
Preventing entire process to stop on one null argument seems reasonable.

It's all about the contract with library's client.
For instance if null's were not allowed and it was clearly stated I'd add validation against it before proceeding with entire sync - it'd fail before start. That's the alternative I see but as I said it's mainly about this contract.
Anyway I'd be happy to see a remark about it in docu :)

heshamMassoud:
@piobra I added it now as a note to the README in the docu fixes pr: 1ef9f60

Issue #46, opened 2017-05-19
Sync should only accept drafts.


heshamMassoud:
### Description
The library should only accept a list of resource drafts as opposed to accepting both list of resources and list of resource drafts. The user of the library is expected to always convert resources to their drafts.

### Expected Behavior
The sync should now only accept drafts of the resources. The user should do the conversion of a resource to its draft using the JVM SDK before hand.

### Current Behavior
Currently the library provides the user with the option to also sync drafts as well as resources.

### Context
JVM SDK now provides the functionality to convert a resource to it's draft. Therefore, it should be the responsibility of the library to do the conversion, but actually the user of the library. 

### Related Issues
#12 

### Hints
Please upgrade the JVM SDK version to the one that contains the draft transformation functionality.

PR #54, opened 2017-05-23
Accept drafts only

piobra:
#### Summary
Remove implementation of processing and syncing non-draft resources.

#### Description
Until now library exposed methods to process and synchronize non-draft resources as well as drafts.
As of version 1.17.0 of ct-jvm-sdk provides draft builders which make it convenient enough for this library user to build drafts based on resources. Thus it is no longer needed to provide similar functionality in the library itself.

#### Relevant Issues
[#46](https://github.com/commercetools/commercetools-sync-java/issues/46)

#### Todo
- Upgrade sdk dependency
- Remove sync and process methods for non-draft resources
- Change _syncDrafts_ and _processDrafts_ names to _sync_ and _process_ since now these are the only entry points
- Remove no longer used transformer and other utils
- Clean up and adjust names of unit tests
- Update documentation


src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -11,7 +11,6 @@
 
 import static java.lang.String.format;
 
-
 public abstract class BaseSync<T, S extends Resource<S>, U extends BaseSyncStatistics, V extends BaseSyncOptions> {
```
heshamMassoud: @piobra I guess generic type`S` is not needed anymore

heshamMassoud: @piobra I will change it in PR #52

heshamMassoud: cfce296

PR #58, opened 2017-05-24
55 fetch inventory without expansion

### LINKED ISSUE #55: Fetch old InventoryEntries without reference expansion
### Description
As @heshamMassoud mentioned, reference expansion is quite inefficient. Currently query of old inventories requires reference expansion so that supply channel `key` could be extracted. It could be easily replaced by id-key mapping, so that we could obtain `key` having only `id` in not expanded reference. At the beginning of inventory sync we cache opposite mapping of key-to-id so we could also store id-to-key.
Efficiency of the syncer should increase then.

### Expected Behaviour
Fetching old inventory entries from CTP requires supply channels reference expansion.

### Current Behaviour
Fetching old inventory entries from CTP should not requiressupply channels reference expansion.

### Possible Solution
I would create a wrapper that would represents existing channels.
It would be backed on two HashMaps. 
It would allow to get id for key and key for id of supply channels. 
The wrapper would be more convenient than separate HashMaps, because as @piobra suggested we should return actual results from sync methods, instead of modify global sync variables.

jakubchrobak:
#### Summary

PR resolves #55. 

#### Description
- `InventorySync` - little refactor was done, mainly concerning not using global variables and increasing the readability.
	- Updated javadoc.
	- Extracted constants to a separate `InventorySyncMessages` class.
	- Refactored the `process` method: the logic of validating, error handling, accumulating and running the batch is not interleaved anymore. Introduced helpers methods: `validateDraft`, `calculateAmountOfBatches` and `getBatch`.
	- Refactored the `processBatch` method: deleted an error handling (moved to relevant fetching method).
	- Refactored the `fetchExistingInventories` method: returns result instead of saving it to a global variable. Added an error handling.  
	- Refactored the `resolveReferencesAndSync`: splitted to separate methods with a single responsibility. `resolveReferences` is now responsible for resolving references. `syncDraft` is now responsible for deciding whether to create or update an inventory.
- `InventoryServiceImpl` - deleted an expansion path.
- `InventorySyncTest` - minor changes.

#### Relevant Issues
https://github.com/commercetools/commercetools-sync-java/issues/55

#### Hints for Review


jakubchrobak:
@heshamMassoud @piobra please review when you find some time.

src/main/java/com/commercetools/sync/inventories/ChannelKeyExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+     *     </li>
+     *     <li>{@code null} - otherwise</li>
+     * </ul>
+     * @param newEntryDraft non-null {@link InventoryEntryDraft} instance
+     * @return {@link String} that is supposed to be supply channel key or {@code null} if draft doesn't belong to
+     *      any supply channel
+     */
+    @Nullable
+    static String extractChannelKey(@Nonnull final InventoryEntryDraft newEntryDraft) {
+        String key = null;
```
piobra: replace this with return statements

src/main/java/com/commercetools/sync/inventories/ChannelKeyExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+        if (newEntryDraft.getSupplyChannel() != null) {
```
piobra: extract to a variable and reuse inf these four places

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -44,9 +45,6 @@
         + "and supply channel key '%s'.";
     private static final String CHANNEL_KEY_MAPPING_DOESNT_EXIST = "Failed to find supply channel of key '%s'.";
```
piobra: in general handling exceptions in this class is very complex, split among many methods
some methods producing completion stages do not have .exceptionally() handlers, some of they have
i'm quite confused about the logic behind it

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -44,9 +45,6 @@
...
```
piobra: extract messages to separate class

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -101,42 +99,49 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
      * @return {@link CompletionStage} with {@link InventorySyncStatistics} holding statistics of all sync
      *                                           processes performed by this sync instance
      */
     @Nonnull
     private CompletionStage<InventorySyncStatistics> splitToBatchesAndProcess(@Nonnull final List<InventoryEntryDraft>
-                                                                                      inventories) {
+                                                                                      drafts,
+                                                                              @Nonnull final Map<String, String>
+                                                                                  supplyChannelKeyToId) {
         List<InventoryEntryDraft> accumulator = new ArrayList<>(syncOptions.getBatchSize());
```
piobra: this method is too complex
extract into several smaller methods

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -151,7 +156,11 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
             }
         }
         if (!accumulator.isEmpty()) {
-            completableFutures.add(processBatch(accumulator).toCompletableFuture());
+            final CompletableFuture<Void> batchProcessingFuture =
+                processBatch(accumulator, supplyChannelKeyToId)
+                    .exceptionally(exception  -> null)
```
piobra: so to me it looks like this
if processBatch completes exceptionally you throw SyncProblemException
and here you ignore it - what's the logic behind it?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -168,119 +177,186 @@ public InventorySyncStatistics getStatistics() {
...
+     *                             CTP project.
+     * @return {@link CompletionStage} of {@link Map} that contains updated {@code supplyChannelKeyToId}
      */
     @SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION") // https://github.com/findbugsproject/findbugs/issues/79
-    private CompletionStage<Void> createMissingSupplyChannels(@Nonnull final List<InventoryEntryDraft> drafts) {
+    private CompletionStage<Map<String, String>> createMissingSupplyChannels(@Nonnull final List<InventoryEntryDraft>
+                                                                                     drafts,
+                                                                             @Nonnull final Map<String, String>
+                                                                                 supplyChannelKeyToId) {
         if (syncOptions.shouldEnsureChannels()) {
```
piobra: method is too complex
split into several smaller methods

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -168,119 +177,186 @@ public InventorySyncStatistics getStatistics() {
...
+                .map(ChannelKeyExtractor::extractChannelKey)
                 .distinct()
                 .filter(Objects::nonNull)
                 .filter(key -> !supplyChannelKeyToId.containsKey(key))
                 .collect(toList());
-            final List<CompletableFuture<Void>> creationStages = missingChannelsKeys.stream()
+            final List<CompletableFuture<Channel>> creationStages = missingChannelsKeys.stream()
                 .map(this::createMissingSupplyChannel)
+                .map(newChannelStage -> newChannelStage
+                    .handle((newChannel, exception) -> {
```
piobra: looks extremely confusing, are you expecting newChannel not to be null
if everything went well?
I'm also wondering about whole inventory sync process integrity and consistency
if some channel failed to be created then some inventories might fail to sync after that,
right?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -168,119 +177,186 @@ public InventorySyncStatistics getStatistics() {
...
+     * @param supplyChannelKeyToId mapping of supply channel key to supply channel Id for supply channels existing in
+     *                             CTP project.
      * @return {@link CompletionStage} of {@link Void} that indicates all possible creation/update attempts progress.
      */
-    private CompletionStage<Void> compareAndSync(final Map<SkuChannelKeyTuple, InventoryEntry> existingInventories,
-                                                 final List<InventoryEntryDraft> drafts) {
+    private CompletionStage<Void> compareAndSync(@Nonnull final List<InventoryEntry> oldInventories,
+                                                 @Nonnull final List<InventoryEntryDraft> drafts,
+                                                 @Nonnull final Map<String, String> supplyChannelKeyToId) {
         final List<CompletableFuture<Void>> futures = new ArrayList<>(drafts.size());
```
piobra: method is too complex
split into several smaller methods

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -168,119 +177,186 @@ public InventorySyncStatistics getStatistics() {
...
+     * Attempts to find in an {@code oldInventories} list an {@link InventoryEntry} of the same {@code sku} and
+     * {@code supplyChannel} as {@code draft} instance. Supply channels, if not null, are compared by id.
+     *
+     * @param oldInventories list of old {@link InventoryEntry} instances
+     * @param draft new {@link InventoryEntryDraft} instance where supply channel reference, if present, points to
+     *              channel in target CTP project
+     * @return {@link Optional} that may contain {@link InventoryEntry} found in {@code oldInventories} that correspond
+     *      to {@code draft} or empty {@link Optional} if no such entry was found
+     */
+    private Optional<InventoryEntry> getCorrespondingEntry(@Nonnull final List<InventoryEntry> oldInventories,
```
piobra: since it's optional's producer follow the convention and name it find*

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -298,64 +374,24 @@ public InventorySyncStatistics getStatistics() {
...
-                        SkuChannelKeyTuple.of(draft).getKey()), exception);
-                    return null;
-                });
-        } else {
-            statistics.incrementFailed();
+    private CompletionStage<InventoryEntry> attemptUpdate(final InventoryEntry entry, final InventoryEntryDraft draft) {
+        final List<UpdateAction<InventoryEntry>> updateActions =
+            InventorySyncUtils.buildActions(entry, draft, syncOptions, typeService);
+        if (!updateActions.isEmpty()) {
+            return inventoryService.updateInventoryEntry(entry, updateActions);
```
piobra: move exception handling here if it is the same concern, don't wrap it in caller method

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -168,119 +177,186 @@ public InventorySyncStatistics getStatistics() {
...
-    private CompletionStage<Map<SkuChannelKeyTuple, InventoryEntry>> fetchExistingInventories(final
-                                                                                              List<InventoryEntryDraft>
-                                                                                                  drafts) {
+    private CompletionStage<List<InventoryEntry>> fetchExistingInventories(@Nonnull final List<InventoryEntryDraft>
+                                                                               drafts) {
         final Set<String> skus = extractSkus(drafts);
-        return inventoryService.fetchInventoryEntriesBySkus(skus)
-            .thenApply(existingEntries -> existingEntries.stream().collect(toMap(SkuChannelKeyTuple::of,
-                entry -> entry)));
+        return inventoryService.fetchInventoryEntriesBySkus(skus);
```
piobra: move exception handling here if it is the same concern, don't wrap it in caller method

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -298,64 +374,24 @@ public InventorySyncStatistics getStatistics() {
...
-                .thenAccept(createdEntry -> statistics.incrementCreated())
-                .exceptionally(exception -> {
-                    statistics.incrementFailed();
-                    syncOptions.applyErrorCallback(format(CTP_INVENTORY_ENTRY_CREATE_FAILED, draft.getSku(),
-                        SkuChannelKeyTuple.of(draft).getKey()), exception);
-                    return null;
-                });
-        } else {
-            statistics.incrementFailed();
+    private CompletionStage<InventoryEntry> attemptUpdate(final InventoryEntry entry, final InventoryEntryDraft draft) {
```
piobra: actually the method with name starting with attempt is kind of wrong
it implies that operation might fail but failure is exceptional situation so it should not
be indicated by method's name. Just 'update' is enough

src/main/java/com/commercetools/sync/inventories/InventorySync.java
piobra: move .exceptionally() handling here since it's the same concern

src/main/java/com/commercetools/sync/inventories/SyncProblemException.java
```diff
@@ -0,0 +1,10 @@
+package com.commercetools.sync.inventories;
+
+/**
+ * Indicates whether sync meets problem which makes further sync unavailable. It may refer to different scope of data,
+ * that are unable to be synced.
+ * e.g. May occur during fetching supply channels what makes all drafts unable to be synced.
+ * e.g.2. May occur during fetching batch of old inventory entries what makes batch of draft unable to be synced.
+ */
+class SyncProblemException extends RuntimeException {
```
piobra: I don't really see any purpose of that if library provides error callback for client
which can act accordingly
plus by looking at the usages of this exception I can tell that it will never leave the scope of any
completion stage - it is swallowed inside other .exceptionally() handlers
regardin name of this class:
every exception indicates a Problem. Don't provide it in the name

piobra: One more thing:
Handling exceptions in this library is not a trivial concern and most of all it's a separate concern so I wouldn't touch it in this PR. A separate issue should be created for it for the whole library.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: Yes, if some channel failed to be created then some inventories might fail to sync after that.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -444,18 +513,19 @@ private InventoryEntryDraft withSupplyChannel(@Nonnull final InventoryEntryDraft
...
+     * returned, otherwise error callback function is executed and {@link CompletionStage} with empty {@link Optional}
+     * is returned.
      *
      * @param supplyChannelKey key of supply channel that seems to not exists in a system
-     * @return {@link CompletionStage} instance that contains either created {@link Channel} instance when succeed
-     *      or {@link SyncProblemException} instance otherwise
+     * @return {@link CompletionStage} instance with {@link Optional} that may contain created {@link Channel}
      */
-    private CompletionStage<Channel> createMissingSupplyChannel(@Nonnull final String supplyChannelKey) {
+    private CompletionStage<Optional<Channel>> createMissingSupplyChannel(@Nonnull final String supplyChannelKey) {
```
piobra: this method is far away of the group of methods that it's logically connected to

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -217,32 +254,57 @@ public InventorySyncStatistics getStatistics() {
...
-                        }
-                        return newChannel;
-                    }))
-                .map(CompletionStage::toCompletableFuture)
+            final List<String> missingChannelsKeys = findMissingChannelsKeys(drafts, supplyChannelKeyToId);
+            final List<CompletableFuture<Void>> newChannelsFutures = createSupplyChannelsOfKeys(missingChannelsKeys)
+                .stream()
+                .map(newChannelFuture -> newChannelFuture
+                    .thenAccept(newChannelOptional -> newChannelOptional.ifPresent(
+                        newChannel -> supplyChannelKeyToId.put(newChannel.getKey(), newChannel.getId()))))
```
piobra: looks like this modifies the map as a side effect in every future. Despite being hard to read,
potentially it could also cause concurrency problems if the code would be parallel.
I'd suggest to modify map AFTER all channels are created and ready in a single shot.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -181,23 +213,28 @@ public InventorySyncStatistics getStatistics() {
...
+     * keys to ids. An empty optional indicates that fetching existing channels failed.
      *
      * @param drafts {@link List} containing {@link InventoryEntryDraft} objects where missing supply channels can occur
-     * @return {@link CompletionStage} of {@link Map} that contain mapping of supply channels' keys to ids, or
-     *      {@link SyncProblemException} when fetching supply channels results in exception
+     * @return {@link CompletionStage} of {@link Optional} that may contain mapping of supply channels' keys to ids.
      */
-    private CompletionStage<Map<String, String>> populateSupplyChannels(@Nonnull final List<InventoryEntryDraft>
-                                                                            drafts) {
+    private CompletionStage<Optional<Map<String, String>>> findSupplyChannels(@Nonnull final List<InventoryEntryDraft>
```
piobra: since it actually creates supply channels I would not use name which implies it only reads some entities

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -181,23 +213,28 @@ public InventorySyncStatistics getStatistics() {
...
-                            .collect(toMap(Channel::getKey, Channel::getId)))
-                .thenCompose(supplyChannelKeyToId -> createMissingSupplyChannels(drafts, supplyChannelKeyToId));
+            .thenApply(existingSupplyChannels -> existingSupplyChannels.stream()
+                .collect(toMap(Channel::getKey, Channel::getId)))
+            .exceptionally(exception -> {
+                syncOptions.applyErrorCallback(CTP_CHANNEL_FETCH_FAILED, exception);
+                return null;
+            })
+            .thenCompose(supplyChannelKeyToId -> {
+                if (supplyChannelKeyToId != null) {
```
piobra: move this if-clause inside create method

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -251,25 +313,24 @@ public InventorySyncStatistics getStatistics() {
...
         return fetchExistingInventories(batchOfDrafts)
-            .exceptionally(exception -> {
-                syncOptions.applyErrorCallback(format(CTP_INVENTORY_FETCH_FAILED, extractSkus(batchOfDrafts)),
-                    exception);
-                throw new SyncProblemException();
-            })
-            .thenCompose(existingInventories -> compareAndSync(existingInventories, batchOfDrafts,
-                supplyChannelKeyToId));
+            .thenCompose(oldInventoriesOptional -> {
+                if (oldInventoriesOptional.isPresent()) {
```
piobra: use Optional mapping

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -331,8 +371,8 @@ public InventorySyncStatistics getStatistics() {
      * @return {@link Optional} that may contain {@link InventoryEntry} found in {@code oldInventories} that correspond
      *      to {@code draft} or empty {@link Optional} if no such entry was found
      */
-    private Optional<InventoryEntry> getCorrespondingEntry(@Nonnull final List<InventoryEntry> oldInventories,
-                                                           @Nonnull final InventoryEntryDraft draft) {
+    private Optional<InventoryEntry> findCorrespondingEntry(@Nonnull final List<InventoryEntry> oldInventories,
+                                                            @Nonnull final InventoryEntryDraft draft) {
         return oldInventories.stream()
             .filter(oldInventory -> oldInventory.getSku().equals(draft.getSku()))
             .filter(oldInventory -> draft.getSupplyChannel() == null
```
piobra: extract this predicate to a method since it's hard to grasp

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -343,20 +383,23 @@ public InventorySyncStatistics getStatistics() {
...
+     * @return {@link CompletionStage} instance with {@link Optional} that may contain list of inventory entries
      */
-    private CompletionStage<List<InventoryEntry>> fetchExistingInventories(@Nonnull final List<InventoryEntryDraft>
-                                                                               drafts) {
+    private CompletionStage<Optional<List<InventoryEntry>>> fetchExistingInventories(@Nonnull final
+                                                                                     List<InventoryEntryDraft> drafts) {
         final Set<String> skus = extractSkus(drafts);
-        return inventoryService.fetchInventoryEntriesBySkus(skus);
+        return inventoryService.fetchInventoryEntriesBySkus(skus)
+            .thenApply(Optional::of)
```
piobra: rather than wrapping in Optional I'd just return emptyList in case of failure and check for emptiness in the line where it's invoked.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -374,24 +417,50 @@ public InventorySyncStatistics getStatistics() {
...
+    private CompletionStage<Void> update(final InventoryEntry entry,
+                                         final InventoryEntryDraft draft) {
         final List<UpdateAction<InventoryEntry>> updateActions =
             InventorySyncUtils.buildActions(entry, draft, syncOptions, typeService);
         if (!updateActions.isEmpty()) {
-            return inventoryService.updateInventoryEntry(entry, updateActions);
+            return inventoryService.updateInventoryEntry(entry, updateActions)
+                .thenAccept(updatedInventory -> statistics.incrementUpdated())
+                .exceptionally(exception -> {
+                    statistics.incrementFailed();
```
piobra: is incrementProcessed done somewhere else along the line?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: I need a way to express that error occured, what can't be accomplished via empty list.
- empty `Optional` indicates that error occured -> so new inventories should not be created (because the state of corresponding inventories within CTP is unknown)
- `Optional` with empty `List` indicates that fetch was successful but didn't find any corresponding inventories within CTP -> so new inventories should be created.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: Inside `compareAndSync`

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -107,9 +107,14 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
     @Override
     protected CompletionStage<InventorySyncStatistics> process(@Nonnull final List<InventoryEntryDraft>
                                                                        inventoryDrafts) {
-        return populateSupplyChannels(inventoryDrafts)
-            .thenCompose(supplyChannelKeyToId -> splitToBatchesAndProcess(inventoryDrafts, supplyChannelKeyToId))
-            .exceptionally(exception -> statistics);
+        return findSupplyChannels(inventoryDrafts)
+            .thenCompose(supplyChannelKeyToIdOptional -> {
+                if (supplyChannelKeyToIdOptional.isPresent()) {
```
jakubchrobak: 7daa78b99c770670ed098205c0f18dbc3fc87f12

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -133,40 +138,67 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
-                    statistics.incrementFailed();
-                    syncOptions.applyErrorCallback(INVENTORY_DRAFT_HAS_NO_SKU, null);
-                }
-            } else {
-                statistics.incrementProcessed();
-                statistics.incrementFailed();
-                syncOptions.applyErrorCallback(INVENTORY_DRAFT_IS_NULL, null);
+            if (shouldProceedWithDraft(draft)) {
+                accumulator.add(draft);
+                accumulator = syncAccumulator(accumulator, completableFutures, supplyChannelKeyToId, true);
```
jakubchrobak: Not sure, if this is what you've asked for? I believe it's less confusing 9d38a3ff4c88df26cf70156c77f96b4a32639009

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -133,40 +138,67 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
 
+    /**
+     * Checks if a draft is valid for further processing. If so, then returns {@code true}.
+     * Otherwise applies an {@code errorCallback}, increments "failed" and "proceessed" statistics and returns
+     * {@code false}.
+     *
+     * @param draft nullable draft, requested for sync
+     * @return boolean that indicate if given {@code draft} is valid for sync
+     */
+    private boolean shouldProceedWithDraft(@Nullable final InventoryEntryDraft draft) {
```
jakubchrobak: 88171a01ab3e5a3686e2bba53a21e6a5e7b9401e

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 88171a01ab3e5a3686e2bba53a21e6a5e7b9401e

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 6cbdf38895da5422f9a303389df3c9dd46408a87 and then 21724e8735237ccdf6d359dc1ae8f2d15f4e5b83 (otherwise `findBugs` breaks build)

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 6cbdf38895da5422f9a303389df3c9dd46408a87

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 3d66b05bff67b638522b80821465412a3ac08031

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 22b5bb9b9a5d3e1a3f3fbb5d014f67201cf16dd5

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 89a441170314dc5206915c942016e314a5d56410

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -25,34 +25,36 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
+import static com.commercetools.sync.inventories.InventorySyncMessages.CTP_INVENTORY_ENTRY_CREATE_FAILED;
```
heshamMassoud: why extract the messages to a separate class, when they are only used by this class? If it is because too many messages, then we need a clear strategy about when do we decide when to separate messages into separate classes and when not? A lot of other classes in the project have messages in them but are not separated. Shall we also separate when it's one message?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -88,143 +88,184 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
-            .allOf(completableFutures.toArray(new CompletableFuture[completableFutures.size()]))
-            .thenApply(result -> {
+                                                                       inventories) {
+        final List<InventoryEntryDraft> validInventories = inventories.stream()
+            .filter(this::validateDraft)
+            .collect(toList());
+        final List<CompletableFuture<Void>> completableFutures = IntStream
+            .range(0, calculateAmountOfBatches(validInventories.size()))
+            .mapToObj(batchIndex -> getBatch(batchIndex, validInventories))
+            .map(batch -> processBatch(batch))
```
heshamMassoud: could be replaced by method reference

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -88,143 +88,184 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
      * @return {@link CompletionStage} with {@link InventorySyncStatistics} holding statistics of all sync
      *                                           processes performed by this sync instance
      */
-    @Nonnull
+    @Override
```
heshamMassoud: why remove the `@Nonnull` annotation.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -77,9 +79,7 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
 
     /**
      * Iterates through the whole {@code inventories} list and accumulates its valid drafts to batches. Every batch
-     * is then processed by {@link InventorySync#processBatch(List)}. For invalid drafts from {@code inventories}
-     * "processed" and "failed" counters from statistics are incremented and error callback is executed. A valid draft
-     * is a {@link InventoryEntryDraft} object that is not {@code null} and its SKU is not empty.
+     * is then processed by {@link InventorySync#processBatch(List)}.
```
heshamMassoud: why remove the doc about the stats being updated? It is also a side effect of calling this method, no?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -88,143 +88,184 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
+        return false;
+    }
+
+    /**
+     * Calculates amount of batches, available in a list of a given size. The batch size is specified by a sync options.
+     *
+     * @param listSize size of a list
+     * @return amount of batches, available in a list of a given size
+     */
+    int calculateAmountOfBatches(int listSize) {
```
heshamMassoud: could be private?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -88,143 +88,184 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
+
+    /**
+     * Returns a batch of drafts, extracted from a given list. The batch size is specified by a sync options.
+     *
+     * @param batchIndex zero-based index of a batch
+     * @param drafts list of drafts
+     * @return an n-th batch of drafts, where n is specified by {@code batchIndex}
+     */
+    @Nonnull
+    List<InventoryEntryDraft> getBatch(int batchIndex, @Nonnull final List<InventoryEntryDraft> drafts) {
```
heshamMassoud: could be private?

heshamMassoud: `final int batchIndex`

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -25,34 +25,36 @@
...
-    private static final String CTP_INVENTORY_ENTRY_CREATE_FAILED = "Failed to create inventory entry of sku '%s' "
-        + "and supply channel id '%s'.";
-    private static final String FAILED_TO_RESOLVE_CUSTOM_TYPE = "Failed to resolve custom type reference on "
-        + "InventoryEntryDraft with sku:'%s'. Reason: %s";
-    private static final String FAILED_TO_RESOLVE_SUPPLY_CHANNEL = "Failed to resolve supply channel reference on "
-        + "InventoryEntryDraft with sku:'%s'. Reason: %s";
 
     private final InventoryService inventoryService;
     private final InventoryReferenceResolver referenceResolver;
-    private Map<InventoryEntryIdentifier, InventoryEntry> existingInventories;
```
heshamMassoud: 👍

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -88,143 +88,184 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
+                                                                                     List<InventoryEntryDraft> drafts) {
         final Set<String> skus = extractSkus(drafts);
         return inventoryService.fetchInventoryEntriesBySkus(skus)
-                               .thenAccept(
-                                   existingEntries -> {
-                                       existingInventories = existingEntries
-                                           .stream().collect(toMap(InventoryEntryIdentifier::of,
-                                               entry -> entry));
-                                   });
+            .thenApply(Optional::of)
```
heshamMassoud: why wrap the result list in an optional?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -88,143 +88,184 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
+     * resolution was successful this method returns a future which contains an {@link Optional} containing resolved
+     * inventory entry draft. Otherwise an error is handled and a future containing an empty optional is returned.
      *
-     * @param inventoryEntryDraft draft that need to be synced
-     * @return a future which contains an empty result after execution of the update.
+     * @param inventoryEntryDraft an inventory entry draft which references has to be resolved
+     * @return a future which may contain the resolved inventory entry draft
      */
-    private CompletableFuture<Void> resolveReferencesAndSync(@Nonnull final InventoryEntryDraft inventoryEntryDraft) {
+    private CompletableFuture<Optional<InventoryEntryDraft>> resolveReferences(@Nonnull final InventoryEntryDraft
```
heshamMassoud: 👍

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -88,143 +88,184 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
+                    inventoryEntryDraft.getSku(), exception.getMessage());
+                handleError(errorMessage, exception, 1);
+                return Optional.empty();
+            })
+            .toCompletableFuture();
     }
 
     /**
-     * Returns a distinct set of SKUs from the supplied list of inventory entry drafts.
+     * Tries to create the new inventory entry or update an old one, depending whether an entry
```
heshamMassoud: `Checks if the {@code resolvedDraft} matches with an old existing inventory entry. If it does, it tries to update it. If it doesn't, it creates it.`

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -257,30 +298,31 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
-     * out successfully or not. If an exception was thrown on executing the request to CTP, the optional error callback
-     * specified in the {@code syncOptions} is called.
+     * out successfully or not. If an exception was thrown on executing the request to CTP, the error handling method
+     * is called.
      *
      * @param draft the inventory entry draft to create the inventory entry from.
      * @return a future which contains an empty result after execution of the create.
      */
-    private CompletionStage<Void> attemptCreate(@Nonnull final InventoryEntryDraft draft) {
+    @SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION") // https://github.com/findbugsproject/findbugs/issues/79
```
heshamMassoud: Why add this annotation here?

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -257,30 +298,31 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
         return inventoryService.createInventoryEntry(draft)
-                               .thenAccept(createdEntry -> statistics.incrementCreated())
-                               .exceptionally(exception -> {
-                                   final Reference<Channel> supplyChannel = draft.getSupplyChannel();
-                                   final String errorMessage = format(CTP_INVENTORY_ENTRY_CREATE_FAILED, draft.getSku(),
-                                       supplyChannel != null ? supplyChannel.getId() : null);
-                                   handleError(errorMessage, exception, 1);
-                                   return null;
-                               });
+            .thenAccept(createdInventory -> statistics.incrementCreated())
```
heshamMassoud: Wrong indentation.

src/main/java/com/commercetools/sync/services/impl/InventoryServiceImpl.java
```diff
@@ -31,8 +30,6 @@ public InventoryServiceImpl(@Nonnull final SphereClient ctpClient) {
         final InventoryEntryQuery query = InventoryEntryQueryBuilder.of()
                                                                     .plusPredicates(
                                                                         queryModel -> queryModel.sku().isIn(skus))
-                                                                    .plusExpansionPaths(
```
heshamMassoud: 🎉

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: d02e9467ce48f53cbfd13f97dce82a6cdeec8565

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 624c6aaa10b2a8d2c844fa48a595f67d769bb815

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 7ab1eb1e051dc1b92d6f23d2088f97aec8bd5ef8

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 7ab1eb1e051dc1b92d6f23d2088f97aec8bd5ef8

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 8606849ef40ef39bec173bc3ace1c09254b57edf

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: 70c1416fa932f6f105b8aad59bf7f5c85166ccd8

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: How is this wrong? Checkstyle reported nothing.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: I already answered the reason here https://github.com/commercetools/commercetools-sync-java/pull/58#discussion_r120898605

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: Yes it is a side effect but you couldn't tell that looking at the method body, could you? I found a  javadoc too broad. It describes more than the method actually do, so it generates a redundancy, and makes that javadoc could become out of date if something has changed in dependent functions.

> "processed" and "failed" counters from statistics are incremented and error callback is executed

This is what the `handleError` method do, so I think it is enough to keep this description in a javadoc of `handleError`.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: Because of conflicted review https://github.com/commercetools/commercetools-sync-java/pull/58#discussion_r119598293 Actually I should have asked about the motivation.

But I agree with you. Constants belong to a `Sync` only, so they should stay inside the class. I reverted the changes  d78891654a39fbe5f308d8ba3bf78a9b80e43b24

src/main/java/com/commercetools/sync/inventories/InventorySync.java
heshamMassoud: @jakubchrobak agreed

src/main/java/com/commercetools/sync/inventories/InventorySync.java
heshamMassoud: @jakubchrobak why not depend on the `exceptionally` block to catch an error occurring, this way it would not continue to create a new inventory, since the exception would not be completed successfully.

jakubchrobak: @heshamMassoud An `exceptionally` block is already present inside `fetchExistingInventories` - right after invocation of service's method. Exception handling should be close to the source of possible errors (service calls) so I wouldn't move the `exceptionally` block outside this function. Since the error is already handled, then method is expected to return something. The `Optional` fits into my needs here. The alternative is to return a nullable resul, but I think `Optional` is better than that.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
heshamMassoud: I honestly don't know how we can enforce a checkstyle rule for enforcing method chaining indentation correctness. But by looking at, the dots of chaining should have the same indentation level.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
heshamMassoud: exactly, I don't mean to change the position of the `exceptionally` block, it's in the correct place :)

Not sure how is `optional` better in this case. 

This is what I mean:

````java
// I would even name this fetchOldEntriesAndSync()
private CompletionStage<Void> processBatch(@Nonnull final List<InventoryEntryDraft> batchOfDrafts) {
    return fetchExistingInventories(batchOfDrafts)
        .thenCompose(oldInventories -> syncBatch(oldInventories, batchOfDrafts));
     }

private CompletionStage<List<InventoryEntry>> fetchExistingInventories(@Nonnull final List<InventoryEntryDraft> drafts) {
    final Set<String> skus = extractSkus(drafts);
    return inventoryService.fetchInventoryEntriesBySkus(skus)
        .exceptionally(exception -> {
            final String errorMessage = format(CTP_INVENTORY_FETCH_FAILED, extractSkus(drafts));
            handleError(errorMessage, exception, drafts.size());
            return null;
            });
    }

````

Why not do it this way?

jakubchrobak: The returned `null` has to be checked inside `processBatch` anyway. So above example should looks more like

````java
private CompletionStage<Void> processBatch(@Nonnull final List<InventoryEntryDraft> batchOfDrafts) {
    return fetchExistingInventories(batchOfDrafts)
        .thenCompose(oldInventories -> oldInventories != null 
             ? syncBatch(oldInventories, batchOfDrafts))
             : completedFuture(null);
     }
````

It does the same as the current version. An `Optional` is just a syntactic sugar (it may be more readable I think, however its argueable).

heshamMassoud: hmm, I see. The thing I was missing, was that exceptionally returns a new exception that is ok actually not. Its ok for now, but I am not sure if this is the correct way to handle errors that were caaught by an exceptionally in a different method. Because Optional<List> doesn't look very right too me..

src/main/java/com/commercetools/sync/inventories/InventorySync.java
jakubchrobak: Yeah, I don't know neither. That's why a common code formatter should be shared, I mentioned it once. I've created an issue with a proposition, to make code style consistent among all developers: https://github.com/commercetools/commercetools-sync-java/issues/66

Issue #38, opened 2017-05-11
Add "already synced" counter to statistics


heshamMassoud:
Currently a resource(e.g. category, inventory entry) which wasn't required to be updated/created will not be counted as processed. 

It however, would be better if there is another counter for example `alreadySynced` which accumulates the number of these unaffected resources.

jakubchrobak:
I thought that this issue was a part of https://github.com/commercetools/commercetools-sync-java/issues/23 so I implemented it in PR https://github.com/commercetools/commercetools-sync-java/pull/59

Issue #23, opened 2017-05-04
Refactor Statistics classes


heshamMassoud:
Processing time implementation could be improved.

Please read discussions about it in this pr: https://github.com/commercetools/commercetools-sync-java/pull/10#discussion_r112173242

andrii-kovalenko-ct:
Also see this suggestion with date/time formatting examples: https://github.com/commercetools/commercetools-sync-java/pull/10#discussion_r112247446

lojzatran:
Also make the statistics class thread-safe or ensure that only one thread is calling the statistics instance at a time.

jakubchrobak:
**Proposed solution:**

- Statistics, similairly to options, should be a pair of immutable object and dedicated builder.
- Statistics should have additional "up to date" counter
- Statistics should have simplified processing time storage.
- Builder should not be responsible for time calculation (because it's error prone)
- Builder should not have method for incrementing "processed" counter because it's redundand (can be calculated as sum of other counters)
- Builder should have method to join statistics object (e.g. batches processed separately returns statistics, that are merged into one then)
- Separate statistics instance should be returned for each sync process
- There could be one synchronized statistics object that concatenates statistics from all sync performed by one sync instance (would be returned by `getStatistics()` - otherwise that method would have no sense anymore).
- There should be separate `JsonFormatter` class instead of static method.

![statistics_diagram](https://cloud.githubusercontent.com/assets/24873948/26631743/d27c5016-460c-11e7-9565-3609941c7a89.png)

PR #59, opened 2017-05-30
23 refactor statistics classes

ghost:
#### Summary
Refactored sync statistics classes. Instance of statistics is unmodifable now, and builders are introduced for construction.

#### Description
- `BaseSyncStatistics` 
   - Changed to be immutable, hid the constructor.
   - Simplified processing time to be held as milliseconds only. Added customizable formatting method. 
   - Added field `upToDate` that indicates number of resources which was processed but not created/updated.
   - Added field with common report message structure.
   - Deleted `getStatisticsAsJsonString`.
- `BaseSyncStatisticsBuilder` - abstract builder of above. 
   - Released from responsibility of time measurement (calculated processing time should be set). 
   - Released from responsibility of `processed` incrementation (as processed = created + updated + upToDate + failed)
   - Added the `addAllStatistics` method which allows to increment all statistics by values present in another statistics instance.
- `CategorySyncStatistics`, `InventorySyncStatistics` - updated `getReportMessage`
- `CategorySyncStatisticsBuilder`, `InventorySyncStatisticsBuilder` - added.
- `BaseSync`
   - Changed `process()` to return statistic builder class. Process should returns builder of a single sync having its counters updated. Thanks to that, a sync time could be set in the `sync()` method and statistics could be built and returned.
   - Deleted `statistics` field and its `getter`. Every `sync` invocation returns its own statistics, so `getStatistics()` seems to be pointless. An user receive a statistics object right after sync finished. If he would like to keep tracking the total amount of resources processed by a single sync instance, he can use a builder to merge statistics received from few syncs.
- `CategorySync` and `InventorySync` were adjusted so that `statistics` builder object was passing around methods, not stored as instance variable. Thanks to that, the synchronization of a statistics object is not needed anymore.
- `SyncStatisticsFormatter` - added interface for formatting statistics objects
- Unit tests was adjusted to above changes

#### Relevant Issues
https://github.com/commercetools/commercetools-sync-java/issues/23
https://github.com/commercetools/commercetools-sync-java/issues/38

#### Todo
- [ ] Ensure if `getStatistics()` method from `BaseSync` is still necessary. It is deleted now (the motivation was described above) . If it's still needed then revert the commit c392b993e0b7c84f20a7b9807b2a9486fcc7147e


jakubchrobak:
@heshamMassoud @piobra please review when you find some time.

src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -22,44 +26,54 @@ protected BaseSync(@Nonnull final U statistics, @Nonnull final V syncOptions) {
...
+        }
+    }
+
+    /**
+     * Join given statistics to total stats of all sync processes performed by {@code this} sync instance.
+     *
+     * @param statisticsToJoin statistics of sync that should be joined to total stats
+     */
+    private void updateTotalStatistics(final U statisticsToJoin) {
+        synchronized (totalStatisticsLock) {
```
heshamMassoud: why use `synchronized` blocks for the `addAllStatistics`? Why not use `AtomicInteger` for all the counters being incremented inside?

src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -22,44 +26,54 @@ protected BaseSync(@Nonnull final U statistics, @Nonnull final V syncOptions) {
...
+     * Returns an instance of type U which is a subclass of {@link BaseSyncStatistics} containing the total stats of
+     * all sync processes performed by {@code this} sync instance.
      *
-     * @return a statistics object for the sync process.
+     * @return a statistics object for all sync processes performed by {@code this} sync instance.
      */
     @Nonnull
     public U getStatistics() {
-        return statistics;
+        synchronized (totalStatisticsLock) {
```
heshamMassoud: why is this block `synchronized` anyway? But if it is why not make the method itself `synchronized`, if the block covers the entire method, instead of having a monitor object/lock?

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -79,32 +81,33 @@ public CategorySync(@Nonnull final CategorySyncOptions syncOptions) {
...
-     * <p>The {@code statistics} instance is updated accordingly to whether the CTP request was carried out
+     * <p>The returned builder instance is updated accordingly to whether the CTP request was carried out
      * successfully or not. If an exception was thrown on executing the request to CTP, the optional error callback
      * specified in the {@code syncOptions} is called.
      *
      * @param categoryDrafts the list of new category drafts to sync to the CTP project.
      * @return an instance of {@link CompletionStage&lt;U&gt;} which contains as a result an instance of
-     *      {@link CategorySyncStatistics} representing the {@code statistics} instance attribute of
-     *      {@link this} {@link CategorySync}.
+     *      {@link CategorySyncStatisticsBuilder} representing the builder of {@code statistics} baked for
```
heshamMassoud: what do you mean `baked for sync process`?

src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -22,44 +26,54 @@ protected BaseSync(@Nonnull final U statistics, @Nonnull final V syncOptions) {
...
      * @param resourceDrafts the list of new resources as drafts.
-     * @return an instance of {@link CompletionStage&lt;U&gt;} which contains as a result an instance of {@link U} which
-     *      is a subclass of {@link BaseSyncStatistics} representing the {@code statistics} instance attribute of
-     *      {@link this} {@link BaseSync}.
+     * @return an instance of {@link CompletionStage&lt;U&gt;} which contains as a result an instance of {@link V} which
+     *      is a subclass of {@link BaseSyncStatisticsBuilder} representing the builder of {@code statistics} baked for
+     *      the sync process
      */
-    protected abstract CompletionStage<U> process(@Nonnull final List<T> resourceDrafts);
+    protected abstract CompletionStage<V> process(@Nonnull final List<T> resourceDrafts);
```
heshamMassoud: please add more explanation in the doc why the builder is returned as a result from this process (to set the processing time).. And how is it used, would be useful for the devs using it.

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatistics.java
```diff
@@ -212,28 +85,32 @@ public long getProcessingTimeInMillis() {
...
-        String result = null;
-        final ObjectMapper mapper = new ObjectMapper();
-        try {
-            result = mapper.writeValueAsString(statistics);
-        } catch (JsonProcessingException processingException) {
-            LOGGER.error("Failed to build JSON String of summary.", processingException);
-        }
-        return result;
-    }
+    @JsonIgnore
```
heshamMassoud: why?

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatisticsBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+    protected int failed = 0;
+    protected long processingTimeInMillis = 0;
+
+    /**
+     * Increments the total number of resource created.
+     *
+     * @return {@code this} instance of {@link T}
+     */
+    public T incrementCreated() {
+        this.created++;
```
heshamMassoud: I guess all the `this` usages in this class are redundant

src/main/java/com/commercetools/sync/commons/helpers/SyncStatisticsFormatter.java
```diff
@@ -0,0 +1,11 @@
+package com.commercetools.sync.commons.helpers;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Provides interface for different formatters of {@link BaseSyncStatistics}.
+ */
+public interface SyncStatisticsFormatter {
+
+    String format(@Nonnull final BaseSyncStatistics statistics);
```
heshamMassoud: please annotate the return of the function. Is it nullable or nonnull?

src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -22,44 +26,54 @@ protected BaseSync(@Nonnull final U statistics, @Nonnull final V syncOptions) {
...
+     *      is a subclass of {@link BaseSyncStatistics} representing the {@code statistics} of this sync process.
      */
     public CompletionStage<U> sync(@Nonnull final List<T> resourceDrafts) {
-        statistics.startTimer();
-        return process(resourceDrafts).thenApply(resultingStatistics -> {
-            resultingStatistics.calculateProcessingTime();
+        final long startTime = System.currentTimeMillis();
+        return process(resourceDrafts).thenApply(resultingStatisticsBuilder -> {
+            resultingStatisticsBuilder.setProcessingTimeInMillis(System.currentTimeMillis() - startTime);
+            final U resultingStatistics = resultingStatisticsBuilder.build();
```
heshamMassoud: you can directly use without storing in var

src/test/java/com/commercetools/sync/commons/helpers/JsonSyncStatisticsFormatterImplTest.java
```diff
@@ -0,0 +1,58 @@
...
+        inventorySyncStatisticsBuilder.setProcessingTimeInMillis(PROCESSING_TIME_IN_MILLIS);
+
+        final String statisticsAsJsonString = jsonFormatter.format(inventorySyncStatisticsBuilder.build());
+        assertThat(statisticsAsJsonString)
+            .isEqualTo("{\"updated\":1,\""
+                + "created\":1,\""
+                + "failed\":1,\""
+                + "upToDate\":1,\""
+                + "processed\":4,\""
+                + "processingTimeInMillis\":" + PROCESSING_TIME_IN_MILLIS + "}");
```
heshamMassoud: I guess now the `reportMessage` is not there anymore when serializing the statistics object into JSON because the instance attribute `reportMessage` was removed from the class.

src/main/java/com/commercetools/sync/commons/BaseSync.java
jakubchrobak: To prevent from access to `totalStatisticsBuilder` object when another thread is inside `updateTotalStatistics()`

src/main/java/com/commercetools/sync/commons/BaseSync.java
jakubchrobak: To avoid inconsistent state. 
When one thread is updating fields the other could call `getStatistics()` and obtain the statistics object with partially updated counters what would be wrong.
That's why I decided to use `lock` which prevent from concurrent access to `totalStatisticsBuilder`.

src/test/java/com/commercetools/sync/commons/helpers/JsonSyncStatisticsFormatterImplTest.java
jakubchrobak: Yes, I deleted a `reportMessage` variable  because it was not used anywhere.
Then I found serialization of report message confusing because there was no `reportMessage` variable.
Basically, do we want to have a report message serialized, right? If so then I'll revert these changes.

src/main/java/com/commercetools/sync/commons/BaseSync.java
jakubchrobak: But 2 objects have to be built then (one for joining to instance statistics and the second for return statement)

src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -1,19 +1,23 @@
...
 
 import javax.annotation.Nonnull;
 import java.util.List;
 import java.util.concurrent.CompletionStage;
 
 
-public abstract class BaseSync<T, U extends BaseSyncStatistics, V extends BaseSyncOptions> {
-    protected final U statistics;
-    protected final V syncOptions;
+public abstract class BaseSync<T, U extends BaseSyncStatistics, V extends BaseSyncStatisticsBuilder<V, U>,
```
jakubchrobak: 2f63bc0203cbeb22da6d5f87e92cd64b26ab7495

src/main/java/com/commercetools/sync/commons/BaseSync.java
jakubchrobak: d068f482c33aaf3e7f610304fbb6180476615eb0

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatistics.java
```diff
@@ -1,44 +1,33 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
-import java.util.concurrent.TimeUnit;
-
-import static java.lang.String.format;
 
 public abstract class BaseSyncStatistics {
     private static final Logger LOGGER = LoggerFactory.getLogger(BaseSyncStatistics.class);
```
jakubchrobak: fc94926a4e3456b4f23a2e930251f4c7318ab04e

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatistics.java
jakubchrobak: It's not needed anymore fc76703d1abb2f6ebc0c61cc6d3313c2ae03049a

src/main/java/com/commercetools/sync/commons/helpers/JsonSyncStatisticsFormatterImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Provides formatter which format {@link BaseSyncStatistics} as JSON string.
+ */
+public class JsonSyncStatisticsFormatterImpl implements SyncStatisticsFormatter {
```
jakubchrobak: True that. I removed it for now 8a7d18aaaa8ea39b7f31635b5ae2c43fffb044ba

src/main/java/com/commercetools/sync/categories/CategorySync.java
jakubchrobak: I meant something like it was not created per sync instance but per invocation of sync method.
Here I adjusted e04195b6ad76160582dd3c58ff407a9deb4e886d

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatisticsBuilder.java
jakubchrobak: 9bd87fa8d6452dd61fe9135e0ff25f5732f06580

src/main/java/com/commercetools/sync/commons/helpers/SyncStatisticsFormatter.java
jakubchrobak: 4986b3a5fac471fef1927cc9c0dbd5a4b3ea3074

src/test/java/com/commercetools/sync/commons/helpers/JsonSyncStatisticsFormatterImplTest.java
heshamMassoud: Seems like Jackson does serialization with getter visibility by default. So any getter should be serialized. So its weird for me why the getReportMessage method wasn't serialized in your JSON object.

jakubchrobak: > However, its weird for me why the getReportMessage method wasn't serialized in your JSON object

It wasn't serialized because of `@JsonIgnore` annotation on a `getReportMessage` method. I inserted`@JsonIgnore` there after I had removed a `reportMessage` field, because I was confused if `reportMessage` should be still serialized.
Finally I got rid of both `reportMessage` field, and `@JsonIgnore` annotation so it can be serialized.

heshamMassoud: aha, ok

piobra:
@heshamMassoud @jakubchrobak 
please see my comment above about requested changes.

I see a big problem with the code. There are many different concerns in the library's logic:
* executing sync
* handling errors
* gathering statistics
* parallel processing

And actually all of them are mixed together, there's lots of tight coupling, logic of those concerns is interlaced with each other.

As I wrote above IMO we should apply some kind of observer pattern implementation here that
would really make coupling a lot more loose and provide means of separation.

And as library is getting more and more complex I'd take it under serious consideration if we should not do it with some reactive library, like https://github.com/ReactiveX/RxJava as it already provides solutions for event-driven async programming (actually strongly related to listener-observer).

jakubchrobak:
@piobra 
Do you suggest that `Sync` classes should be redesigned? Could you provide problem description and your proposition of a solution in an issue or (confluence)[https://sphere.atlassian.net/wiki/spaces/PS/pages/145192925/commercetools+java+sync]?

src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -34,32 +42,38 @@ protected BaseSync(@Nonnull final U statistics, @Nonnull final V syncOptions) {
...
+     * Returns an instance of type S which is a subclass of {@link BaseSyncStatistics} containing the total stats of
+     * all sync processes performed by {@code this} sync instance.
      *
-     * @return a statistics object for the sync process.
+     * @return a statistics object for all sync processes performed by {@code this} sync instance.
      */
     @Nonnull
-    public U getStatistics() {
-        return statistics;
+    public synchronized S getStatistics() {
```
andrii-kovalenko-ct: since `totalStatisticsBuilder` is final  i'd rather lock on this instance, instead of whole `BaseSync` instance:
```java
public S getStatistics() {
    synchronized (totalStatisticsBuilder) {
        return totalStatisticsBuilder.build();
    }
}
```

Same for `BaseSync#updateTotalStatistics` below.

But, honestly, i don't like the whole concept of the statistic update/get:
1) do we really expect that statistic could be fetched before sync completed? If no, then we need some flag like _completed_ and return nothing (null, empty object) before completion.

2) do we expect concurrent `updateTotalStatistics()` access? As i see from `BaseSync#sync` - no. Then synchronization there is redundant (just set _completed_ flag only when the statistic is completely updated.

3) If we really should stick with concurrency here - pls provide a respective comments to the methods.

jakubchrobak: @andrii-kovalenko-ct 
`getStatistics()` was there since the beginning, when `sync()` returned the `void`. Now `sync()` returns `CompletionStage` containing a distinct statistics instance for every subsequent invocations. Because of that, keeping `getStatistics()` makes sense only if we want to hold a merged statistics of all `sync()` invocations performed by a single `BaseSync` instance. The question is do we want to do so? 
If the answer is no, then I'll delete both `update/get` functions. The user will be able to merge statistics by himself. 
Otherwise I think that concurrent `updateTotalStatistics()` is possible, cause user may invoke`sync()` multiple times concurrently e.g. for different chunks fetched from CSV. That's why we need a synchronization here.

jakubchrobak: They seem redundant so I deleted them c392b993e0b7c84f20a7b9807b2a9486fcc7147e

andrii-kovalenko-ct:
@heshamMassoud 
is this PR is still actual? Looks like we already completely re-factored statistic classes so these changes should de dropped.

heshamMassoud:
@andrii-kovalenko-ct The only reason I keep it open is there were some good ideas as far as I remember mentioned it. So I have it open just to remember especially when working on #23 which this PR addresses actually.

Issue #47, opened 2017-05-19
Reference resolution 


heshamMassoud:
### Description

Some resources have references to other resources like supply channels on inventory entries or custom types on categories. The sync will always use the `key` to identify the referenced resources. This is because 
- when syncing from an external feed, the internal CTP resource id is not known. 
- when syncing from another CTP project, the CTP resource id is different

To better describe the issue, I will use supply channels on inventory entries as an example. 

Currently, for the two scenarios:
**1. Syncing inventory entries from a data feed (XML, CSV)**
In order for the sync to be able to access this key, it needs to be accessible in one of the following ways:
 - Either provide the inventory entry with reference expansion for the supply channels. This way the key is directly accessible from the expanded channel.
- Or don't do reference expansion, but provide the key value of the channel in place of the reference id. For example: `Reference.of("Channel", "keyValue");`

**2. Syncing inventory entries from another CTP project.**
In order for the sync to be able to access this key, it needs to be accessible in only one way:
- The Inventory entry has to have the channel reference expanded in order to get the key. 

The problem with the current approach is:-
**Wrong channel creation**
In the first scenario, if ensureChannel is set to true, the inventory is not reference expanded and the key is not set in the id field, then the sync will see this as an Inventory Entry without a Channel and thus will wrongly create a new one.

In the second scenario, if ensureChannel is set to true and the reference is not expanded, then the sync will see this as an Inventory Entry without a Channel and thus will wrongly create a new one.

### Proposed Solution

![referenceresoltion](https://cloud.githubusercontent.com/assets/9512131/26248073/12f6549a-3ca1-11e7-926a-545996a0c78e.png)
For resources that have references to other resources on them, for example InventoryEntry has `Custom` which references a `Type` resource and `supplyChannel` which references a `Channel` resource. 

- This would be input in the library in a wrapper which has the `InventoryEntryDraft`, the `supplyChannelKey` and `customTypeKey`. Let's say this wrapper is called `CustomInventoryEntry` (open for better naming proposals

- The user of the library should care how to provide these keys.
- If the key is not set (`null`), then the library will consider this InventoryEntry without a Channel and will not create one for it. Even if ensureChannel is set.
- If the key is set, the library will check if this channel exists, if it doesn't and `ensureChannel` is true, it will create this Channel.

FYI: @jakubchrobak @butenkor @piobra

butenkor:
[...]

Issue #27, opened 2017-05-05
Building update action `changeParent` for category should use external id


heshamMassoud:
### Description
Input category drafts should be expected to have the `externalId` of the parent category as a value for the `parent` field.

### Expected Behaviour
On building update action `changeParent` for category, the given drafts should be expected to have the `externalId` of the parent category as a value, for the `parentField`, instead of the actual UUID of the parent category. This is because the parent of a draft could be a new category that is still not on CTP, so there is no UUID for it, that the user of the sync could actually attach to the `parent` field of the draft.

### Current Behaviour
Currently the sync expects that the actual UUID of the parent category is there in the `parent` field of every input draft. Exact location in code is [here](https://github.com/commercetools/commercetools-sync-java/blob/CTT-192-Category-Sync/src/main/java/com/commercetools/sync/categories/utils/CategoryUpdateActionUtils.java#L109-L116).

### Context
 This is because the parent of a draft could be a new category that is still not on CTP, so there is no UUID for it that the user of the sync could actually attach to the `parent` field of the draft.

### Possible Solution

1. On the first run of a category sync, fetch all categories and cache a map of `externalId` -> `UUID` of categories. (To avoid multiple GET requests).
2. On each category create request add the new Category UUID to the cache.
3. Right now, since current implementation expects the list of categories to be given ordered hierarchically. So every time, a parent assignment is done, the external id should be expected to be found in this cached map. Therefore,  [here](https://github.com/commercetools/commercetools-sync-java/blob/CTT-192-Category-Sync/src/main/java/com/commercetools/sync/categories/utils/CategoryUpdateActionUtils.java#L109-L116), would be changed to fetching the actual UUID from the map that corresponds to the `externalId` stored in the draft. However, this order would not be a requirement after the improvement in this issue #28 is applied.
4. It should be well documented as a contract of this library that it expects the given category drafts have the `externalId` of the parent category as a value for the `parent` field instead of the actual `UUID`.

Issue #14, opened 2017-03-31
CustomUpdateActionUtils should be decoupled from TypeService 


heshamMassoud:
The CustomUpdateActionUtils should not depend on a TypeService to get a cached key of the custom type.  https://github.com/commercetools/commercetools-sync-java/pull/10/files#diff-cea7e36a8d3c43856febd3c4d36f8145R122

But then otherwise, it should expect an already (reference-expanded Custom Type on the custom fields.)

heshamMassoud:
Also TypeService method should be 
````java
@Nonnull
    CompletionStage<Optional<String>> getCachedTypeKeyById(@Nullable final String id);
````

heshamMassoud:
and `fetchAllTypesKeysIntoCache` method should be `fetchKeyAndCache(id)` to also return `CompletionStage<Optional<String>>` instead of `CompletionStage<Void>`

heshamMassoud:
In the cases, where the cache of the Type keys is not provided, the sync would expect that reference expansion is already done and would try to get the key from the expanded object. The caching options are there to encourage the users of the sync module not to do reference expansion which degrades performance.

heshamMassoud:
update README after issue resolution.

PR #60, opened 2017-05-31
47 reference resolution

heshamMassoud:
#### Summary
This PR mainly addresses the implementation for reference resolution for syncing resources with references for example parent category reference on categories, supply channels on inventory entries, custom type references on any resource with custom fields. 

#### Description
The proposed solution of #47 does not prove as a good solution due to the fact that Product resources have several references on them, not only that, but these references have references to other resource. This means that the `wrapper` concept proposed in the solution will need to be implemented for almost every single resource (product, product variant, prices, etc..) This adds a whole layer of complexity that the user of the sync will need to understand in order to use the library which is not user friendly.

Therefore, the solution implemented instead of the proposed one is for the user to provide the referenced resource keys either via:
1. The `id` field on the `reference`.
2. The expanded reference with its accessible `key` field.

The sync library will always resolve the keys on the references before calculating update actions. So by the end reference comparison will always be compared by their actual ids (UUID). 

The sync will have the option `allowUUID` which if set to `true` will enable the user to place keys with a UUID format, this is only in the case if the external system uses keys with UUIDs. However, by default this option is set to `false`, therefore the sync will validate the key and check if it's in UUID format then it would trigger the error callback and fail syncing the draft under question.

This reference resolution implementation is made generic by having a super abstract class: `BaseReferenceResolver`, which is extended by any resource which needs reference resolution for example for categories `CategoryReferenceResolver` and for inventory entries `InventoryReferenceResolver`   

#### Relevant Issues
#47, #14, #27 

#### Todo
- Tests
    - [x] Unit 
- [x] Adapt README

#### Status
Ready for review @jakubchrobak @piobra please review when you have time.


src/main/java/com/commercetools/sync/services/impl/InventoryServiceImpl.java
```diff
@@ -19,14 +13,11 @@
...
 import javax.annotation.Nonnull;
-import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
 
-import static java.util.Collections.singleton;
-
-final class InventoryServiceImpl implements InventoryService {
+public class InventoryServiceImpl implements InventoryService {
```
jakubchrobak: Why don't `final`? It doesn't seem like designed for inheritance.

src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
```diff
@@ -10,23 +10,49 @@
...
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
 public class CategoryServiceImpl implements CategoryService {
     private final SphereClient ctpClient;
 
+    /**
+     * Cache of Categories' [externalId -> id].
+     */
+    private final Map<String, String> cache = new HashMap<>();
```
jakubchrobak: Also apply to other services.

src/test/java/com/commercetools/sync/commons/MockUtils.java
```diff
@@ -22,6 +40,91 @@ public static CustomFieldsDraft getMockCustomFieldsDraft() {
...
+
+    /**
+     * Creates a mock instance of {@link CustomFieldsDraft} with the supplied {@code customTypeId} and
+     * {@code customFields}.
+     *
+     * @return a mock instance of {@link CustomFieldsDraft} with supplied id and custom fields.
+     */
+    public static CustomFieldsDraft getMockCustomFieldsDraft(@Nullable final String customTypeId,
+                                                             @Nonnull final Map<String, JsonNode> customFields) {
+        final CustomFieldsDraft newCategoryCustomFieldsDraft = mock(CustomFieldsDraft.class);
```
jakubchrobak: Wouldn't be better to test with real draft objects? Unless you verify method calls.

src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java
```diff
@@ -143,20 +150,73 @@ public void sync_WithEnsuredChannels_ShouldCreateEntriesWithUnknownChannels() {
...
                 .join();
         assertThat(stats.getProcessed()).isEqualTo(1);
         assertThat(stats.getFailed()).isEqualTo(1);
         assertThat(stats.getCreated()).isEqualTo(0);
         assertThat(stats.getUpdated()).isEqualTo(0);
+        assertThat(errorCallBackMessages).hasSize(1);
+        assertThat(errorCallBackMessages.get(0)).isEqualTo(format("Failed to resolve supply channel reference on"
+            + " InventoryEntryDraft with sku:'%s'. Reason: "
+            + "com.commercetools.sync.commons.exceptions.ReferenceResolutionException: "
+            + "Channel with key '%s' does not exist.", SKU_3, KEY_3));
```
jakubchrobak: Shouldn't be constants resued somehow in such comparisions? Let's imagine someone change `ReferenceResolutionException` name or move it to another package, how many tests would this affect?

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Helper method that returns an {@link Optional} containing the value of the id field on the passed
+     * {@link Reference} object.
+     *
+     * @param reference the reference to get the value of it's id field.
+     * @return an optional containing the id field value of the passed reference.
+     */
+    @Nonnull
+    private static Optional<String> getKeyFromReference(@Nonnull final Reference reference) {
```
heshamMassoud: 0e0036c

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Given an id as {@link String}, this method checks whether if it is in UUID format or not.
+     *
+     * @param id to check if it is in UUID format.
+     * @return true if it is in UUID format, otherwise false.
+     */
+    private static boolean isUuid(@Nonnull final String id) {
+        final String uuidRegex = "[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}";
```
heshamMassoud: bce7c26

src/main/java/com/commercetools/sync/commons/BaseSyncOptions.java
```diff
@@ -144,4 +147,16 @@ public void applyErrorCallback(@Nonnull final String errorMessage, @Nullable fin
             this.errorCallBack.accept(errorMessage, exception);
         }
     }
+
+    /**
+     * The sync expects the user to pass the keys to references in the {@code id} field of References. If the key values
+     * are in UUID format, then this flag must be set to true, otherwise the sync will fail to resolve the reference.
+     * This flag, if set to true, enables the user to use keys with UUID format. By default, it is set to {@code false}.
```
heshamMassoud: c4d9b3b

src/main/java/com/commercetools/sync/commons/BaseSyncOptions.java
```diff
@@ -144,4 +147,16 @@ public void applyErrorCallback(@Nonnull final String errorMessage, @Nullable fin
...
+     *
+     * @return a {@code boolean} flag, if set to true, enables the user to use keys with UUID format for references.
+     *      By default, it is set to {@code false}.
+     */
+    public boolean isUuidAllowed() {
```
heshamMassoud: c4d9b3b

src/main/java/com/commercetools/sync/commons/utils/CustomUpdateActionUtils.java
```diff
@@ -149,22 +147,21 @@
...
+            if (isBlank(oldCustomTypeId)) {
+                throw new BuildUpdateActionException(format(CUSTOM_TYPE_IDS_NOT_SET,
                     resource.toReference().getTypeId()));
             }
             if (newCustomFieldsJsonMap == null) {
                 // New resource's custom fields are null/not set. So we should unset old custom fields.
                 final UpdateAction<T> updateAction = buildTypedSetCustomTypeUpdateAction(
-                    newCustomFieldsTypeKey, null, resource, syncOptions).orElse(null);
+                    newCustomTypeId, null, resource, syncOptions).orElse(null);
                 return updateAction != null ? Collections.singletonList(updateAction) : Collections.emptyList();
```
heshamMassoud: good catch! 	8ce2712

src/main/java/com/commercetools/sync/commons/utils/CustomUpdateActionUtils.java
```diff
@@ -174,7 +171,7 @@
         } else {
             final UpdateAction<T> updateAction =
                 buildTypedSetCustomTypeUpdateAction(
-                    newCustomFieldsTypeKey, newCustomFieldsJsonMap, resource, syncOptions)
+                    newCustomTypeId, newCustomFieldsJsonMap, resource, syncOptions)
                     .orElse(null);
```
heshamMassoud: 8ce2712

src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java
```diff
@@ -92,6 +93,20 @@ public T setRemoveOtherProperties(final boolean removeOtherProperties) {
     }
 
     /**
+     * The sync expects the user to pass the keys to references in the {@code id} field of References. If the key values
+     * are in UUID format, then this flag must be set to true, otherwise the sync will fail to resolve the reference.
+     * This flag, if set to true, enables the user to use keys with UUID format. By default, it is set to {@code false}.
+     * This method sets this flag's value with the new value.
```
heshamMassoud: 15eafa9

src/main/java/com/commercetools/sync/services/impl/TypeServiceImpl.java
```diff
@@ -10,34 +10,42 @@
...
 
     public TypeServiceImpl(@Nonnull final SphereClient ctpClient) {
         this.ctpClient = ctpClient;
     }
 
-    @Nullable
+    @Nonnull
     @Override
-    public String getCachedTypeKeyById(@Nullable final String id) {
+    public CompletionStage<Optional<String>> fetchCachedTypeId(@Nullable final String key) {
```
heshamMassoud: 3f671a8, 486d631

src/main/java/com/commercetools/sync/services/impl/TypeServiceImpl.java
```diff
@@ -10,34 +10,42 @@
...
         if (cache.isEmpty()) {
-            fetchAllTypesKeysIntoCache().toCompletableFuture().join();
+            cacheAndFetch(key);
```
heshamMassoud: 88c2284

src/main/java/com/commercetools/sync/services/impl/ChannelServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private final Map<String, String> cache = new HashMap<>();
+
+    public ChannelServiceImpl(@Nonnull final SphereClient ctpClient) {
+        this.ctpClient = ctpClient;
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Optional<String>> fetchCachedChannelIdByKeyAndRoles(@Nonnull final String key,
```
heshamMassoud: very good catch! 68f4b44 I'd stick now to keeping a separate implementation of the service for each usage of the channel reference.

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    @Nonnull
+    public CompletionStage<CategoryDraft> resolveCustomTypeReference(@Nonnull final CategoryDraft categoryDraft) {
+        final CustomFieldsDraft custom = categoryDraft.getCustom();
+        if (custom != null) {
+            return getCustomTypeId(custom)
+                .thenApply(resolvedTypeIdOptional ->
+                    resolvedTypeIdOptional.filter(StringUtils::isNotBlank)
```
heshamMassoud: yes, if the type doesn't exist.

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
```diff
@@ -0,0 +1,116 @@
...
+                                          .map(resolvedTypeId ->
+                                              CategoryDraftBuilder.of(categoryDraft)
+                                                                  .custom(CustomFieldsDraft.ofTypeIdAndJson(
+                                                                      resolvedTypeId, custom.getFields()))
+                                                                  .build())
+                                          .orElseGet(() -> CategoryDraftBuilder.of(categoryDraft).build()));
```
heshamMassoud: because here the map() returns an optional containing a `CategoryDraftDsl` so the orElseGet should return the same type and `categoryDraft` is passed as a `CategoryDraft` not `CategoryDraftDsl` (which actually implements it). If there is a better way? please suggest :)

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+        return CompletableFuture.completedFuture(CategoryDraftBuilder.of(categoryDraft).build());
```
heshamMassoud: 858cd61

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
```diff
@@ -0,0 +1,116 @@
...
+        @Nonnull final String parentCategoryExternalId) {
+        return categoryService.fetchCachedCategoryId(parentCategoryExternalId)
+                              .thenApply(resolvedParentIdOptional -> resolvedParentIdOptional
+                                  .filter(StringUtils::isNotBlank)
+                                  .map(resolvedParentId ->
+                                      CategoryDraftBuilder.of(categoryDraft)
+                                                          .parent(Reference.of(Category.referenceTypeId(),
+                                                              resolvedParentId))
+                                                          .build())
+                                  .orElseGet(() -> CategoryDraftBuilder.of(categoryDraft).build()));
```
heshamMassoud: see comment above https://github.com/commercetools/commercetools-sync-java/pull/60#discussion_r120322449

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
```diff
@@ -0,0 +1,177 @@
...
+                                                                                      inventoryEntryDraft) {
+        final CustomFieldsDraft custom = inventoryEntryDraft.getCustom();
+        if (custom != null) {
+            return getCustomTypeId(custom).thenApply(resolvedTypeIdOptional ->
+                resolvedTypeIdOptional.filter(StringUtils::isNotBlank)
+                                      .map(resolvedTypeId -> InventoryEntryDraftBuilder
+                                          .of(inventoryEntryDraft)
+                                          .custom(CustomFieldsDraft.ofTypeIdAndJson(resolvedTypeId, custom.getFields()))
+                                          .build())
+                                      .orElseGet(() -> InventoryEntryDraftBuilder.of(inventoryEntryDraft).build()));
```
heshamMassoud: see comment above https://github.com/commercetools/commercetools-sync-java/pull/60#discussion_r120322449

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
```diff
@@ -0,0 +1,116 @@
...
+    @Nonnull
+    private CompletionStage<CategoryDraft> fetchAndResolveParentReference(
+        @Nonnull final CategoryDraft categoryDraft,
+        @Nonnull final String parentCategoryExternalId) {
+        return categoryService.fetchCachedCategoryId(parentCategoryExternalId)
+                              .thenApply(resolvedParentIdOptional -> resolvedParentIdOptional
+                                  .filter(StringUtils::isNotBlank)
+                                  .map(resolvedParentId ->
+                                      CategoryDraftBuilder.of(categoryDraft)
+                                                          .parent(Reference.of(Category.referenceTypeId(),
```
heshamMassoud: oh, thx! Didn't know about it :) I changed everywhere: 24efb9f

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+        return CompletableFuture.completedFuture(InventoryEntryDraftBuilder.of(inventoryEntryDraft).build());
```
heshamMassoud: 858cd61

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
```diff
@@ -0,0 +1,177 @@
...
+        if (channelReference != null) {
+            try {
+                final String keyFromExpansion = getKeyFromExpansion(channelReference);
+                final String channelKey = getKeyFromExpansionOrReference(keyFromExpansion, channelReference);
+                return fetchOrCreateAndResolveReference(draft, channelKey);
+            } catch (ReferenceResolutionException exception) {
+                return CompletableFutureUtils.exceptionallyCompletedFuture(exception);
+            }
+        }
+        return CompletableFuture.completedFuture(InventoryEntryDraftBuilder.of(draft).build());
```
heshamMassoud: 858cd61

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
```diff
@@ -0,0 +1,177 @@
...
+     *      supply channel reference or an exception.
+     */
+    @Nonnull
+    private CompletionStage<InventoryEntryDraft> fetchOrCreateAndResolveReference(
+        @Nonnull final InventoryEntryDraft draft,
+        @Nonnull final String channelKey) {
+        return channelService.fetchCachedChannelIdByKeyAndRoles(channelKey,
+            Collections.singletonList(ChannelRole.INVENTORY_SUPPLY))
+                             .thenCompose(resolvedChannelIdOptional -> resolvedChannelIdOptional
+                                 .filter(StringUtils::isNotBlank)
```
heshamMassoud: yes, if the channel doesn't exist.

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
```diff
@@ -0,0 +1,177 @@
...
+     *      {@link InventoryEntryDraft} object as a result of setting the passed {@code channelId} as the id of channel
+     *      reference.
+     */
+    @Nonnull
+    private static CompletionStage<InventoryEntryDraft> setChannelReference(@Nonnull final String channelId,
+                                                                            @Nonnull final InventoryEntryDraft
+                                                                                inventoryEntryDraft) {
+        return CompletableFuture.completedFuture(InventoryEntryDraftBuilder
+            .of(inventoryEntryDraft)
+            .supplyChannel(Reference.of(Channel.referenceTypeId(), channelId))
```
heshamMassoud: 24efb9f

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    @Nonnull
+    private CompletionStage<InventoryEntryDraft> createChannelAndSetReference(@Nonnull final String channelKey,
+                                                                              @Nonnull final InventoryEntryDraft
+                                                                                  inventoryEntryDraft) {
+        if (getOptions().shouldEnsureChannels()) {
+            return channelService.createAndCacheChannel(channelKey, Collections.singleton(ChannelRole.INVENTORY_SUPPLY))
+                                 .thenApply(createdChannel -> InventoryEntryDraftBuilder
+                                     .of(inventoryEntryDraft)
+                                     .supplyChannel(Reference.of(Channel.referenceTypeId(), createdChannel.getId()))
```
heshamMassoud: 24efb9f

src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
heshamMassoud: agreed 28eb2b0

src/main/java/com/commercetools/sync/services/impl/InventoryServiceImpl.java
heshamMassoud: 4cf6f08

src/test/java/com/commercetools/sync/commons/MockUtils.java
heshamMassoud: in this specific case actually it is better to use the builder for instantiation because I was trying to mock it specifically to be able to pass null ids for tests. Which is actually impossible to do when instantiating through the normal builder. Therefore I changed it here: 	37cf6ed and removed the impossible test cases. But in some other cases it is easier to mock using mockito for example to mock a category draft without a slug..

src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java
heshamMassoud: good idea: f79e405

heshamMassoud:
@jakubchrobak I have responded to the comments. Can you please create an issue with what you mean exactly with the common code formatter? 

Regarding the refreshing caches, it's a valid use case what you are mentioning. But I would also create an issue for it. I think we need to brainstorm how to expose this mechanism through the library.

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
jakubchrobak: Ah, right. I don't know any better way then :sweat_smile:

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
jakubchrobak: So there could be a mapping of `"someChannelKey" -> null / blankString` inside a cache? I thought that cache contains only values received from API, so `id` always is not blank.
Isn't an empty `Optional` returned if the channel doesn't exists?

heshamMassoud: > So there could be a mapping of "someChannelKey" -> null / blankString inside a cache?

No this is not possible

> Isn't an empty Optional returned if the channel doesn't exists?

Yes, you're right and in that case the filter is unnecessary. Will change it ;)

heshamMassoud: bfd10ec

src/main/java/com/commercetools/sync/services/impl/InventorySupplyChannelService.java
```diff
@@ -0,0 +1,81 @@
...
+        final ChannelDraft draft = ChannelDraftBuilder.of(key)
+                                                      .roles(roles)
+                                                      .build();
+        return ctpClient.execute(ChannelCreateCommand.of(draft));
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Channel> createAndCacheChannel(@Nonnull final String key,
+                                                          @Nonnull final Set<ChannelRole> roles) {
```
heshamMassoud: d711636

src/main/java/com/commercetools/sync/services/impl/InventorySupplyChannelService.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        return CompletableFuture.completedFuture(Optional.ofNullable(keyToIdCache.get(key)));
+    }
+
+    private CompletionStage<Optional<String>> cacheAndFetch(@Nonnull final String key) {
+        final ChannelQuery query =
+            ChannelQueryBuilder.of()
+                               .plusPredicates(channelQueryModel -> channelQueryModel
+                                   .roles()
+                                   .containsAny(Collections.singletonList(ChannelRole.INVENTORY_SUPPLY)))
```
heshamMassoud: d711636

src/main/java/com/commercetools/sync/services/impl/InventorySupplyChannelService.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+
+public final class InventorySupplyChannelService implements ChannelService {
```
heshamMassoud: d711636

src/main/java/com/commercetools/sync/inventories/helpers/InventoryReferenceResolver.java
```diff
@@ -0,0 +1,172 @@
...
+    @Nonnull
+    private CompletionStage<InventoryEntryDraft> createChannelAndSetReference(@Nonnull final String channelKey,
+                                                                              @Nonnull final InventoryEntryDraft
+                                                                                  inventoryEntryDraft) {
+        if (getOptions().shouldEnsureChannels()) {
+            return channelService.createAndCacheChannel(channelKey, Collections.singleton(ChannelRole.INVENTORY_SUPPLY))
+                                 .thenApply(createdChannel -> InventoryEntryDraftBuilder
+                                     .of(inventoryEntryDraft)
+                                     .supplyChannel(Channel.referenceOfId(createdChannel.getId()))
+                                     .build());
```
heshamMassoud: e9a7959

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -140,283 +100,202 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
+            InventorySyncUtils.buildActions(entry, draft, syncOptions);
+        if (!updateActions.isEmpty()) {
+            return inventoryService.updateInventoryEntry(entry, updateActions)
+                .thenAccept(updatedEntry -> statistics.incrementUpdated())
                 .exceptionally(exception -> {
-                    statistics.incrementFailed();
-                    syncOptions.applyErrorCallback(format(CTP_INVENTORY_ENTRY_CREATE_FAILED, draft.getSku(),
-                        SkuChannelKeyTuple.of(draft).getKey()), exception);
+                    final String errorMessage = format(CTP_INVENTORY_ENTRY_UPDATE_FAILED, draft.getSku(),
+                        SkuChannelKeyTuple.of(draft).getKey());
```
jakubchrobak: `SkuKeyTuple` was designed to extract tuples from input drafts which either have an expanded references or a `key` provided within an `id` field.
At this point your `draft` should have its reference resolved (so it points to proper `id`). Because of that `SkuChannelKeyTuple.of(draft).getKey()` will return then an CTP's uuid not a `key`. (as you can see I had a `draft` and a `fixedDraft` instances before)
I think, that either a message has to be adjusted, or `key` should be fetched in other way (service cache maybe?).

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -140,283 +100,202 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
-        final Reference<Channel> supplyChannelRef = Channel.referenceOfId(supplyChannelId);
-        return InventoryEntryDraftBuilder.of(draft)
-            .supplyChannel(supplyChannelRef)
-            .build();
+    private CompletionStage<Void> attemptCreate(@Nonnull final InventoryEntryDraft draft) {
+        return inventoryService.createInventoryEntry(draft)
+                               .thenAccept(createdEntry -> statistics.incrementCreated())
+                               .exceptionally(exception -> {
+                                   final String errorMessage = format(CTP_INVENTORY_ENTRY_CREATE_FAILED, draft.getSku(),
+                                       SkuChannelKeyTuple.of(draft).getKey());
```
jakubchrobak: Same as above.

src/main/java/com/commercetools/sync/services/impl/ChannelServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.annotation.Nonnull;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+
+public final class ChannelServiceImpl implements ChannelService {
```
jakubchrobak: I see one more possible issue with the cache. As you said elsewhere:

> A very common use case of this library is streaming data from an XML or CSV file, and calling the sync(drafts) on each batch concurrently.

But running `sync(drafts)` concurrently can cause a problem here. All running processes would depend on the very same instance of a service, but an access to the cache isn't synchronized. 
This applies also to other services.

src/main/java/com/commercetools/sync/inventories/InventorySync.java
heshamMassoud: True, I have refactored the `SkuChannelKeyTuple` implementation to just `InventoryEntryIdentifier` since the only thing we need from it is to have a container for the sku and the supply channel id which both uniqley identify and inventory. Implemented here with tests: 24ee590

src/main/java/com/commercetools/sync/inventories/InventorySync.java
heshamMassoud: 24ee590

src/main/java/com/commercetools/sync/services/impl/ChannelServiceImpl.java
heshamMassoud: @jakubchrobak can you please create an issue with what you mean as it's not related to this pr?

heshamMassoud: @jakubchrobak 56a9bd2

jakubchrobak: @heshamMassoud  It's better now.
But I think that lazy initialization is still error prone.
Let's imagine 2 threads are calling `fetchCachedCategoryId`. For both the statement `externalIdToIdCache.isEmpty()` could return `true` and they both calls API. 
Another example: the first thread might start to initialize cache (but hasn't finished yet), while the second thread tries to fetch an Id from the cache. It may result in empty `Optional` because value isn't present yet (but will be as soon as the first thread finished his insertions).

heshamMassoud: > Let's imagine 2 threads are calling fetchCachedCategoryId. For both the statement externalIdToIdCache.isEmpty() could return true and they both calls API.

But this exactly shouldn't happen with `ConcurrentHashMap`. From the [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) of it:
> Bear in mind that the results of aggregate status methods including size, isEmpty, and containsValue are typically useful only when a map is not undergoing concurrent updates in other threads.

src/main/java/com/commercetools/sync/inventories/README.md
```diff
@@ -49,14 +49,17 @@ inventorySync.sync(inventoryEntryDrafts);
...
-to the `supplyChannel` expanded. This means that calling `getObj()` on the reference should not
-return `null`, but return the [Channel](https://github.com/commercetools/commercetools-jvm-sdk/blob/master/commercetools-models/src/main/java/io/sphere/sdk/channels/Channel.java)
-object, from which the sync can access its `key`. Example of sync performed that way can be found [here](https://github.com/commercetools/commercetools-sync-java/blob/master/src/integration-test/java/com/commercetools/sync/inventories/InventorySyncTest.java#L128).
-- If the reference is not expanded, then it is very important to provide the channel `key` in the `id` field of the
-reference. This means calling `getId()` on the channel reference would return it's `key`. Example of sync performed that
-way can be found [here](https://github.com/commercetools/commercetools-sync-java/blob/master/src/integration-test/java/com/commercetools/sync/inventories/InventorySyncTest.java#L181).
+
+ Every inventory entry has a reference to a supply `Channel` and a reference to the `Type` of its custom fields. These
```
jakubchrobak: I think that every inventory entry `may have` a reference, as there is no obligation for providing any references.

heshamMassoud: b1e394c

Issue #57, opened 2017-05-24
Product Sync Initial Implementation


piobra:
### Description
Provide an initial/draft implementation of Product Sync module.

### TODO
- [create a skeleton of all parts of module:] 
- [* update actions util] 
- [* product service] 
- [* adjust already existing product sync options if needed] 
- [implement first couple of actions which together form CRUD operations on products] 
- [implement Product Service to support created actions] 
- [plug some of sync options into implementation of action build util] 
- [add unit tests for created implementation] 
- [initial documentation]

PR #61, opened 2017-06-07
57 product sync initial impl

piobra:
#### Summary
WIP
Initial implementation of product sync module.
Issue #57.

#### Description
Contains fundamental implementation of Product Sync Module with unit and integration tests.

#### Hints for Review
This should be reviewed mainly from architectural point of view as most update actions are not there yet.
Attention should be paid to handling of following sync options:
* compareStaged; // to control whether to compare to the staged data or the published ones.
* updateStaged; // to control whether to update staged or current projection
* publish; // to control whether to auto-publish or not.

Javadoc quality can be neglected in this PR.

Stil todo:
* handling errors
* move integration test out of unit tests folder - it depends on it tests done for other modules
* markdown documentation
* javadoc
* null related annotations
* rest of update actions and handling related sync options


src/test/java/com/commercetools/sync/services/ProductServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+        service = ProductService.of(ctpClient);
+    }
+
+    @Test
+    public void create() {
+        when(ctpClient.execute(any())).thenReturn(completedFuture(mock(Product.class)));
+
+        Product product = service.create(mock(ProductDraft.class)).toCompletableFuture().join();
+
+        assertThat(product).isNotNull();
```
andrii-kovalenko-ct: I'm not sure this services should be covered by Unit tests - doesn't have much sense. 
But even if you decided mock it in unit test:

  1. 
```java
Product productMock = mock(Product.class);
when(ctpClient.execute(any())).thenReturn(completedFuture(productMock));
.......
assertThat(product).isSameAs(productMock);
```

  2. Since you are testing exactly `Create`:
```java
when(ctpClient.execute(any(ProductCreateCommand.class)))
......
verify(ctpClient).execute(any(ProductCreateCommand.class));
```

src/main/java/com/commercetools/sync/products/ProductSyncOptions.java
```diff
@@ -16,16 +16,17 @@
...
+    private final boolean compareStaged; // to control whether to compare to the staged data or the published ones.
+    private final boolean updateStaged; // to control whether to update staged or current projection
+    private final boolean publish; // to control whether to auto-publish or not.
+    private final boolean removeOtherVariants; // to control whether to remove other product variants or not.
 
     // defines which attributes
-    private List<String> whiteList;
-    private List<String> blackList;
+    private final List<String> whiteList;
+    private final List<String> blackList;
```
piobra: it's not currently handled but it will probably be just for convenience so that you can choose one of them. So if you wanted to only provide some attributes you can do that with whitelist instead of blacklisting all others and vice versa.

src/main/java/com/commercetools/sync/products/actions/ProductUpdateActionsBuilder.java
```diff
@@ -0,0 +1,228 @@
...
+import static java.util.Objects.nonNull;
+import static java.util.stream.Collectors.toList;
+
+public class ProductUpdateActionsBuilder {
+
+    private static final ProductUpdateActionsBuilder productUpdateActionsBuilder;
+
+    // eager singleton
+    static {
+        productUpdateActionsBuilder = new ProductUpdateActionsBuilder();
```
piobra: Right, I removed this constraint. I was also thinking it'd be good to extract interface for all action builders as providing them as utility static classes is not good. They should only provide one public method for building all actions for specific module.

src/test/java/com/commercetools/sync/it/products/SphereClientUtils.java
```diff
@@ -0,0 +1,53 @@
...
+import io.sphere.sdk.client.SphereClient;
+import io.sphere.sdk.client.SphereClientConfig;
+import io.sphere.sdk.client.SphereRequest;
+import io.sphere.sdk.queries.PagedQueryResult;
+
+import javax.annotation.Nonnull;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public class SphereClientUtils {
```
piobra: yes, I based it on @jakubchrobak work but he did not merge his work in master so please consider it as WIP for now.

src/test/java/com/commercetools/sync/services/ProductServiceTest.java
piobra: why do you think it doesn't make sense?
service's method has a certain signature which matters, provides implementation for some contract (even if not stated in javadoc yet), it has some logic, even if it's trivial. What are the cases when code should not be unit tested?

src/test/java/com/commercetools/sync/it/products/SphereClientUtils.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param client sphere client used to executing requests
+     * @param query supplier of resources that need to be processed
+     * @param ctpRequest function that takes resource and returns sphere request
+     * @param <T> type of resource
+     */
+    public static <T> void fetchAndProcess(@Nonnull final SphereClient client,
+                                           @Nonnull final Supplier<SphereRequest<PagedQueryResult<T>>> query,
+                                           @Nonnull final Function<T, SphereRequest<T>> ctpRequest) {
+
+        client.execute(query.get()).toCompletableFuture().join().getResults()
```
piobra: pls, see comment above

src/test/java/com/commercetools/sync/services/ProductServiceTest.java
piobra: 5a0bad05c2fe7095340e6ff64adfdd20a153229b

src/test/java/com/commercetools/sync/it/products/SphereClientUtils.java
andrii-kovalenko-ct: Could you put a respective comment, like: _TODO: re-factor with inventory/categoriest test sphere utils, when merged_

Thanks

src/test/java/com/commercetools/sync/it/products/SphereClientUtils.java
andrii-kovalenko-ct: this comment explains only that it is blocking, but it doesn't explain why this blocking strategy is used (why instead of blocking once of N parallel operations we block N times for every operation). Kind of: instead of parallel operation you are doing it sequentially.

As for query _query that will fetch all resources_: this is true only up to 500 items, you can't request more than 500 in a one request. Likely it is not a case for cleaning test tables, but better to explicitly mention it in the comment.

src/test/java/com/commercetools/sync/services/ProductServiceTest.java
andrii-kovalenko-ct: My point is just current Unit test implementation doesn't have much sense:
  
  1. You mock _any_ CTP client request, when it likely makes sense to mock by type (e.g., if you call create product method - mock CreateProductCommand call only.
  2. You verify only null validation of the result, when it makes sense to validate the service returns some certain data (in my example above it verifies that `service.create()` returns the value i exactly mock.

Same for `update`, `publish`, revert` and other method.

BUT, my point is: this is a simple service, and it's better to cover it by integration test:
create a draft, save it in CTP, read from CTP, update, verify the value is updated.

In this case Unit-test is redundant, because it will be re-covered by integration test.

src/test/java/com/commercetools/sync/it/products/SphereClientUtils.java
piobra: I meant the comment for your previous remark - that it is a WIP and the utils class for other modules will be used when merged

andrii-kovalenko-ct: ah, ok

src/test/java/com/commercetools/sync/services/ProductServiceTest.java
piobra: https://github.com/commercetools/commercetools-sync-java/pull/61/commits/03bb5cb759ddbe7c74bd18c918c403ce6bd85bb3

src/test/java/com/commercetools/sync/it/products/SphereClientUtils.java
piobra: https://github.com/commercetools/commercetools-sync-java/pull/61/commits/03bb5cb759ddbe7c74bd18c918c403ce6bd85bb3

src/test/java/com/commercetools/sync/services/ProductServiceTest.java
andrii-kovalenko-ct: ok

Issue #31, opened 2017-05-05
Should retry on API errors like 5xx


butenkor:
Affects all modules (should be covered by JVM SDK)

jakubchrobak:
**Proposed solution**:
- provide utility function which wraps a sphere client in [RetrySphereClientDecorator](http://commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/client/RetrySphereClientDecorator.html)
- use that function in a services' constructors (cause they are responsible for performing CRUD operations on CTP and they take a `sphereClient` as a parameter)

@butenkor I'm not sure about the strategy regarding calculation of waiting time between subsequent retry attemps. Could you provide me expected numbers so that it was reasonable?
E.g. Would a sequence `2s, 2s, 2s, 4s, 4s, 4s, ... , 512s, 512s, 512s` of waiting time between following attempts be ok? It would result in 26 attempts, lasting almost 2 hours in a case of the constant failure.

heshamMassoud:
Doesn't have to be a responsibility of the sync library. Showing an example in integration tests should be enough.

PR #62, opened 2017-06-07
31 should retry on 5xx

jakubchrobak:
#### Summary
Provide retry on server errors.

#### Description
- I provided `getRetryOn5xxClient` function in `ClientConfigurationUtils`, that wraps `SphereClient` into a `RetrySphereClientDecorator`.
- I used that function in every sync's constructor while instantiating services.
- I provided up to 27 retries with durations between each retry `2s, 2s, 2s, 4s, 4s, 4s, 8s, ... , 512s`, for each failed request. 
- I wrote relevant unit tests.

#### Relevant Issues
https://github.com/commercetools/commercetools-sync-java/issues/31

#### Hints for Review
Please express opinion about retry strategy, and if it should be more customizable within a method parameters.


src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -53,8 +54,8 @@
      */
     public CategorySync(@Nonnull final CategorySyncOptions syncOptions) {
         this(syncOptions,
-            new TypeServiceImpl(syncOptions.getCtpClient()),
-            new CategoryServiceImpl(syncOptions.getCtpClient()));
+            new TypeServiceImpl(getRetryOn5xxClient(syncOptions.getCtpClient())),
```
heshamMassoud: I wouldn't do it here on every time we need to get the client, but rather one time when it's injected and bound to the options.

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
```diff
@@ -1,22 +1,32 @@
...
 
+import static io.sphere.sdk.retry.RetryAction.ofScheduledRetry;
+import static io.sphere.sdk.retry.RetryPredicate.ofMatchingErrors;
+import static java.util.Collections.singletonList;
+
 public class ClientConfigurationUtils {
     private static BlockingSphereClient ctpClient;
     private static final long DEFAULT_TIMEOUT = 30;
     private static final TimeUnit DEFAULT_TIMEOUT_TIME_UNIT = TimeUnit.SECONDS;
+    private static final int RETRY_ON_5XX_ATTEMPTS_LIMIT = 27;
```
heshamMassoud: why 27?

heshamMassoud: Also why is it a global variable and not passed as a parameter to the method it's used in?

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
```diff
@@ -46,6 +56,36 @@ public static BlockingSphereClient createClient(@Nonnull final SphereClientConfi
...
+    @Nonnull
+    public static SphereClient getRetryOn5xxClient(@Nonnull final SphereClient sphereClient) {
+        return RetrySphereClientDecorator.of(sphereClient, singletonList(RetryRule.of(
+            ofMatchingErrors(ServerErrorException.class),
+            ofScheduledRetry(RETRY_ON_5XX_ATTEMPTS_LIMIT, ClientConfigurationUtils::retryOn5xxDurationCalculator))));
+    }
+
+    /**
+     * Given a {@link RetryContext} it returns duration for a current attempt. The duration is calculated from a
+     * following equation: <pre>duration = 2 ^ (((retryAttemptNumber - 1) / 3) + 1) [seconds]</pre>
```
heshamMassoud: could you please explain the idea behind this strategy?

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
jakubchrobak: 27 fits into proposed default strategy. More about it in a comment below.

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
jakubchrobak: About the strategy: it is only my proposition. In fact I have no clue what retry strategy is wanted here. I have expressed my concerns some time ago in a issue https://github.com/commercetools/commercetools-sync-java/issues/31#issuecomment-306434605

This particular strategy seemed reasonable to me: first it performs few attempts with a little duration and then the time is increased. There is 27 attempts so it sum to almost 2 hours of waiting (e.g. when the server is down). 

Please share your opinion about what is expected here @heshamMassoud

src/main/java/com/commercetools/sync/categories/CategorySync.java
jakubchrobak: Ah, I forgot about the options. Good point :+1:

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
heshamMassoud: retry strategy: https://github.com/sphereio/sphere-node-utils/blob/master/src/coffee/helpers/repeater.coffee#L90-L96

Provide through IT as an example, not function of sync..

Issue #73, opened 2017-07-18
Batch processing of categories


heshamMassoud:
### Description
Category sync should give the user the option to specify the size of a batch of categories to process in chunks. This is to avoid having huge queries per requests, for example, in the case when if a user inputs in the sync a list of 10000 category drafts. The library should internally batch them and split it them on multiple requests.

PR #74, opened 2017-07-18
#73: batch process categories

heshamMassoud:
#### Summary
Addresses the issue of batch processing of categories.

#### Description
This PR addresses the issue of:
1. #73: Batch processing of categories

#### Relevant Issues
#73


heshamMassoud:
@lojzatran what do you mean by "1 batch fails"? A lot of errors could occur during the sync process at many different places (E.g. at reference resolution, creation, updating, deletion, etc..) all of these cases are caught if they occur and trigger a callback for the user with the corresponding exception. If however, an unaccounted exception occurs other than all of these that should fail the sync, then `thenCompose` for the next batch won't continue for the batches, as it should since probably a fatal exception occurred which shouldn't. Isn't testing this kind of redundant?

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -105,6 +106,40 @@ public CategorySync(@Nonnull final CategorySyncOptions syncOptions) {
...
+    /**
+     * Given a list of {@code CategoryDraft} that represent a batch of category drafts, this method for the first batch
+     * only caches a list of all the categories in the CTP project in a cached map that representing each category's
+     * key to the id. It then validates the category drafts, then resolves all the references. Then it creates all
+     * categories that need to be created in parallel while keeping track of the categories that have their
+     * non-existing parents. Then it does update actions that don't require parent changes in parallel. Then in a
+     * blocking fashion issues update actions that don't involve parent changes sequentially.
+     *
+     * <p>More on the exact implementation of how the sync works here:
+     * https://sphere.atlassian.net/wiki/spaces/PS/pages/145193124/Category+Parallelisation+Technical+Concept
```
heshamMassoud: @lojzatran 903c65a

lojzatran:
Maybe this implementation has one drawback. Let's imagine you have 4 categories and you want to batch them by 2 and 2. The first batch contains cats [A, B] and the second contains [C, D]. Now you start syncing batch [A, B]. What if A is finished long before B? Then anyway you cannot start with C because B is still on going. You always need to wait for the whole batch to finish to continue with the next batch. 
I don't think it's a real problem here as parallel calls in one batch should be equally fast, but still it's a drawback.

heshamMassoud:
@lojzatran Batching is implemented for two things: 
1. To restrict the payloads of fetched data.
2. To restrict the query predicate size when matching categories, for example, we try to find the categories with keys that are **in** `[key1, key2, key3, etc..]`. If this list is huge than the request would fail because there is a size limitation for the request on CTP. (so actually batching is a must for this specific case.)

Issue #76, opened 2017-07-18
Info about categories with missing parents


heshamMassoud:
The sync statistics object should contain information about the categories that refer to missing parents.

Issue #75, opened 2017-07-18
Total processing time shouldn't be calculated by sync 


heshamMassoud:
### Description
The total processing time wouldn't be the responsibility of the sync. It should be the responsibility of the user of the sync. 

Currently the sync would support calculating the processing time of the latest batch processed.

PR #78, opened 2017-07-19
75 remove total processing time, add missing parents info in stats object, refactoring in ITs..

heshamMassoud:
#### Summary
This PR addresses the following:
1.  #75: Changes processing times counter naming, since it represents the time for the latest batch: 76914a6
2. #76: Add info about missing parents in the statistics output of the category sync:  5b6f46f, d7fb44c, 7fe9207, 3a852e5
3. Rework category deletion for integration tests. Before this PR, categories in IT were always created under a root category parent, this is to easily delete the root category of the CTP project after/before running the tests and in turn deletion is aggregated on the children and/or grandchildren. However, this is not the best idea since devs would always need to create categories under this root, otherwise categories would be dangling and therefore, they would have to manually go delete the dangling categories. In this PR: another solution is implemented for category deletion here:  c995dd7. It basically, keeps track of which categories were deleted by storing the keys and in turn never issue a delete request for a category which had an ancestor already deleted. 
4. Always set orderHint value for created categories to avoid unneeded update actions due to autogenerated orderHint value from CTP: 6f19819
5. Method naming changes in IT utils: 06548bf


src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
```diff
@@ -20,8 +27,26 @@ public CategorySyncStatistics() {
...
         reportMessage = format("Summary: %s categories were processed in total "
-                + "(%s created, %s updated and %s categories failed to sync).",
-            getProcessed(), getCreated(), getUpdated(), getFailed());
+                + "(%s created, %s updated, %s failed to sync and %s categories with a missing parent).",
+            getProcessed(), getCreated(), getUpdated(), getFailed(),
+            getNumberOfCategoriesWithMissingParents(categoryKeysWithMissingParents));
         return reportMessage;
     }
+
+    static int getNumberOfCategoriesWithMissingParents(
```
heshamMassoud: because its used in unit tests.

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
```diff
@@ -3,9 +3,16 @@
...
 
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
 import static java.lang.String.format;
 
 public class CategorySyncStatistics extends BaseSyncStatistics {
+    private Map<String, ArrayList<String>> categoryKeysWithMissingParents = new HashMap<>();
```
heshamMassoud: 5664a52

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
```diff
@@ -20,8 +27,26 @@ public CategorySyncStatistics() {
...
-            getProcessed(), getCreated(), getUpdated(), getFailed());
+                + "(%s created, %s updated, %s failed to sync and %s categories with a missing parent).",
+            getProcessed(), getCreated(), getUpdated(), getFailed(),
+            getNumberOfCategoriesWithMissingParents(categoryKeysWithMissingParents));
         return reportMessage;
     }
+
+    static int getNumberOfCategoriesWithMissingParents(
+        @Nonnull final Map<String, ArrayList<String>> categoryKeysWithMissingParents) {
+        return categoryKeysWithMissingParents.values()
```
heshamMassoud: no because a category can only have one parent.

PR #86, opened 2017-08-28
Documentation

heshamMassoud:
Add more documentation for easier library usage.


src/main/java/com/commercetools/sync/categories/README.md
```diff
@@ -94,13 +122,16 @@ entries.
 - `removeOtherProperties`
 a flag which enables the sync module to add additional object properties (e.g. custom fields, etc..) without deleting 
 existing ones, if set to `false`. If set to `true`, which is the default value of the option, it deletes the existing 
-object properties.
+object properties. -->
 - `updateActionsFilter`
 a filter function which can be applied on generated list of update actions to produce a resultant list after the filter 
 function has been applied.
 - `allowUuid`
 a flag, if set to `true`, enables the user to use keys with UUID format for references. By default, it is set to `false`.
```
lojzatran: Provide info what you need to provide if it's set to false?

heshamMassoud: a key, as defined by CTP, should be a alphanumeric string. So there is no specific format for it.

PR #88, opened 2017-08-31
Handle empty target categoy keys

heshamMassoud:
This PR handles the case if there are categories in the target project which don't have a key set. It will trigger the warningCallback.

Integration tests are added as well.

### DIFF
--- a/src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
+++ b/src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
@@ -10,6 +10,7 @@ import io.sphere.sdk.categories.commands.CategoryCreateCommand;
 import io.sphere.sdk.categories.commands.CategoryUpdateCommand;
 import io.sphere.sdk.categories.queries.CategoryQuery;
 import io.sphere.sdk.commands.UpdateAction;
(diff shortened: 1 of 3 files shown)


src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
```diff
@@ -48,7 +51,15 @@ public CategoryServiceImpl(@Nonnull final CategorySyncOptions syncOptions) {
         }
 
         final Consumer<List<Category>> categoryPageConsumer = categoriesPage ->
-            categoriesPage.forEach(category -> keyToIdCache.put(category.getKey(), category.getId()));
+            categoriesPage.forEach(category -> {
```
lojzatran: is this class/method specific only for target/source category? If yes, then you can add the name to be e.g. `targetCategory`.

heshamMassoud: The sync always takes as an input a list of drafts. It never manipulates with the source project. Therefore, the entire services of the whole project are used for manipulating the target CTP project. I think renaming it to `targetCategory` would only mean that the service might also be extended to be used for a the source project which it shouldn't.

Issue #98, opened 2017-09-15
Sync product attributes


heshamMassoud:
### Implementation details
1. Product types should be fetched and cached on-demand per batch.
2. Only some, but not the entire, information on Product types are needed to be fetched: 
 a. **Each attribute's `type`**: The types that we should be concerned with for the MVP are `BooleanType`, `TextType`, `LocalizableTextType`, `EnumType`, `LocalizableEnumType`, `NumberType` and `SetType`.
b. The `EnumType` and `LocalizableEnumType` values should not be cached since they are big and are not needed.
c. **`isRequired` field**: if the field is required but doesn't actually have a new value, the sync library would trigger the error call back instead of making a request to CTP.
d. **`attributeConstraint` field**: if it has the value `SameForAll`, then the `setAttributeInAllVariants` update action should be used for adding/removing/changing an attribute value at the same time for all the variants.

P.S. The name of every attribute is unique. 

### Hints 
`Product#masterData#staged#variants#attributes`

http://dev.commercetools.com/http-api-projects-products.html#attribute

### Update Actions required

1. [x] addVariant
2. [x] removeVariant
3. [x] changeMasterVariant
4. [x] setSku
5. [x] setAttribute
6. [x] setAttributeInAllVariants
7.  ~setProductVariantKey~

andrii-kovalenko-ct:
closed by #113

ahmetoz:
The issue looks like not managed after it's open. The changed described in the issue is already implemented.

PR #103, opened 2017-09-19
98 add remove variant

andrii-kovalenko-ct:
#### Summary
In `ProductVariantUpdateActionUtils` added:
  - `buildRemoveVariantUpdateAction()`
  - `buildAddVariantUpdateAction()`
  - `buildChangeMasterVariantUpdateAction()`

Respective unit tests added with Payment mock files.

#### Relevant Issues
  
#98

#### Todo

- [ ] more unit tests for corner cases?

#### Hints for Review

`ProductVariantUpdateActionUtils` is the main implementation

`CollectionUtils` is additional utilities class for filtering and converting collections

### DIFF
--- a/src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
@@ -19,27 +19,46 @@ public final class CommonTypeUpdateActionUtils {
      * {@link Optional}. If both the {@link Object}s have the same values, then no update action is needed and hence an
      * empty {@link Optional} is returned.
      *
-     * @param oldObject            the object which should be updated.
-     * @param newObject            the object with the new information.
-     * @param updateActionSupplier the supplier that returns the update action to return in the optional.
+     * @param oldObject            the object which should be updated
+     * @param newObject            the object with the new information
+     * @param updateActionSupplier the supplier that returns the update action to return in the optional
      * @param <T>                  the type of the {@link UpdateAction}
-     * @param <S>                  the type of the objects to compare.
-     * @return A filled optional with the update action or an empty optional if the object values are identical.
+     * @param <S>                  the type of the objects to compare
+     * @param <U>                  certain {@link UpdateAction} implementation type
+     * @return A filled optional with the update action or an empty optional if the object values are identical
      */
     @Nonnull
-    public static <T, S> Optional<UpdateAction<T>> buildUpdateAction(@Nullable final S oldObject,
-                                                                  @Nullable final S newObject,
-                                                                  @Nonnull final Supplier<UpdateAction<T>>
-                                                                    updateActionSupplier) {
+    public static <T, S, U extends UpdateAction<T>> Optional<U> buildUpdateAction(
+            @Nullable final S oldObject,
+            @Nullable final S newObject,
+            @Nonnull final Supplier<U> updateActionSupplier) {
+
         return !Objects.equals(oldObject, newObject)
-            ? Optional.ofNullable(updateActionSupplier.get()) : Optional.empty();
+                ? Optional.ofNullable(updateActionSupplier.get())
+                : Optional.empty();
     }
 
+    /**
+     * Compares two {@link Object} and returns a supplied list of {@link UpdateAction} as a result.
+     * If both the {@link Object}s have the same values, then no update action is needed
+     * and hence an empty list is returned.
+     *
+     * @param oldObject            the object which should be updated
+     * @param newObject            the object with the new information
+     * @param updateActionSupplier the supplier that returns a list of update actions if the objects are different
+     * @param <T>                  the type of the {@link UpdateAction}
+     * @param <S>                  the type of the objects to compare
+     * @param <U>                  certain {@link UpdateAction} implementation type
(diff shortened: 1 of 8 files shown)


src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
```diff
@@ -0,0 +1,116 @@
...
+import static com.commercetools.sync.commons.utils.CollectionUtils.filterCollection;
+import static com.commercetools.sync.commons.utils.CommonTypeUpdateActionUtils.buildUpdateAction;
+import static java.util.stream.Collectors.toList;
+
+public final class ProductVariantUpdateActionUtils {
+
+    /**
+     * <b>Note:</b> if you do both add/remove product variants - <b>first always remove the variants</b>,
+     * and then - add.
+     * The reason of such restriction: if you add first you could have duplication exception on the keys,
```
heshamMassoud: > if you add first, the update action could fail due to having a duplicate `key` or `sku` with variants which were expected to be removed anyways. So issuing remove update action first will fix such issue.

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
```diff
@@ -0,0 +1,116 @@
...
+     *     <li>attributes</li>
+     *     <li>prices</li>
+     *     <li>images</li>
+     * </ul>
+     *
+     * @param draft {@link ProductVariantDraft} which to add.
+     * @return new {@link AddVariant} update action with properties from {@code draft}
+     */
+    @Nonnull
+    public static AddVariant buildAddVariantUpdateActionFromDraft(@Nonnull final ProductVariantDraft draft) {
```
heshamMassoud: would be good to add the JVM SDK :)

heshamMassoud:
@andrii-kovalenko-ct Also this PR is missing using the actual implemented update action methods in the ProductSyncUtils#buildCoreActions. So that the sync process can actually use them. Please add them there and add at least one IT that tests syncing a product with different variants and check if the product was updated and synced correctly.

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
heshamMassoud: also why `public`?

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
```diff
@@ -0,0 +1,116 @@
...
+     * @param newProduct new product draft with variants <b>with resolved references prices references</b>
+     * @return list of update actions to add new variants.
+     * @see #buildRemoveVariantUpdateAction(Product, ProductDraft)
+     */
+    @Nonnull
+    public static List<UpdateAction<Product>> buildAddVariantUpdateAction(@Nonnull final Product oldProduct,
+                                                                          @Nonnull final ProductDraft newProduct) {
+        final List<ProductVariantDraft> newVariants = newProduct.getVariants();
+        final Set<String> oldVariantKeys =
+                collectionToSet(oldProduct.getMasterData().getStaged().getVariants(), ProductVariant::getKey);
```
heshamMassoud: @andrii-kovalenko-ct `getVariants()` doesn't include the master variant. `getAllVariants()` does. So It means that the master variant will never be synced this way! 😮

andrii-kovalenko-ct:
>Please move the JSON resources to the resources directory. Other than that looks good :)

but they are in the resources directory. Which JSON you mean?

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
andrii-kovalenko-ct: 03193bf

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
andrii-kovalenko-ct: c05f147
made internal to use in the tests.

heshamMassoud:
> but they are in the resources directory. Which JSON you mean?

Ah ok, I didn't notice

src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
```diff
@@ -19,27 +19,46 @@
...
+     * @param <S>                  the type of the objects to compare
+     * @param <U>                  certain {@link UpdateAction} implementation type
+     * @return A filled optional with the update action or an empty optional if the object values are identical
      */
     @Nonnull
-    public static <T, S> Optional<UpdateAction<T>> buildUpdateAction(@Nullable final S oldObject,
-                                                                  @Nullable final S newObject,
-                                                                  @Nonnull final Supplier<UpdateAction<T>>
-                                                                    updateActionSupplier) {
+    public static <T, S, U extends UpdateAction<T>> Optional<U> buildUpdateAction(
```
heshamMassoud: @andrii-kovalenko-ct Why is this better?

PR #129, opened 2017-10-02
#98: fix javadoc publishing

### LINKED ISSUE #117: JavaDoc Jar seems to not be uploaded to bintray
### Description
https://travis-ci.org/commercetools/commercetools-sync-java/builds/280844376

Seems that the JavaDoc Jar is not built still (as seen in the latest build) and hence the releases are still not making it to maven central: https://bintray.com/commercetools/maven/commercetools-sync-java#central

andrii-kovalenko-ct:
- fix #117. The main problem was in `clean` task after the build before `gitPublishPush` and `bintrayUpload`
  - rename _travis-build.sh_ to _travis-publish.sh_
  - [print actual build version during gradle build](https://github.com/commercetools/commercetools-sync-java/pull/129/files#diff-9c0be52e80e6470229a9f697ecee0ce1R6)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: true
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,7 @@ cache:
   - $HOME/.gradle/wrapper/
 after_success:
 - bash <(curl -s https://codecov.io/bash)
-- ./gradlew clean -Dbuild.version=$TRAVIS_TAG gitPublishPush
-- ./travis-build.sh
+- ./travis-publish.sh
 notifications:
   hipchat:
     template:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ cache:
   - $HOME/.gradle/wrapper/
 after_success:
 - bash <(curl -s https://codecov.io/bash)
-- ./travis-publish.sh
+- ./travis-publish.sh || travis_terminate 1
 notifications:
   hipchat:
     template:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: true
+sudo: false
 addons:
   apt:
     packages:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -19,8 +19,7 @@ cache:
   - $HOME/.gradle/wrapper/
 after_success:
 - bash <(curl -s https://codecov.io/bash)
-- ./gradlew clean -Dbuild.version=$TRAVIS_TAG gitPublishPush
-- ./travis-build.sh
+- ./travis-publish.sh || travis_terminate 1
```
heshamMassoud: @andrii-kovalenko-ct what happens here if you don't add the `travis_terminate 1`

andrii-kovalenko-ct: if `travis-publish.sh` finishes with an error (e.g. exit code ≠ 0) - the build will still be accounted as success. For this reason we should use this construction to fail the build if script is failed

andrii-kovalenko-ct:
the problem with gh publish might be in gh plugin version.

travis-publish.sh
```diff
@@ -9,7 +9,7 @@ export TAG=`if [ "$TRAVIS_PULL_REQUEST" = "false" -a -n "$TRAVIS_TAG" ] ; then e
 
 if [ "$TAG" ]; then
   echo "Build is tagged. Uploading artifact $TAG to Bintray."
-  ./gradlew --info -Dbuild.version=$TRAVIS_TAG bintrayUpload
+  ./gradlew --info -Dbuild.version="$TAG" gitPublishPush bintrayUpload
```
heshamMassoud: @andrii-kovalenko-ct Question: if the `gitPublishPush` fails due to authentication failure (or any other reason) will the `bintrayUpload` task still be executed? and also could it happen that the `gitPublishPush` succeeds but the `bintrayUpload` fails? 

We should ensure that if either of them fails, both are not excuted.

heshamMassoud:
@andrii-kovalenko-ct also could you please adjust these cmds in the README: https://github.com/commercetools/commercetools-sync-java#build-and-publish-javadoc? Thanks :)

travis-publish.sh
andrii-kovalenko-ct: 6c99781b
 -  if `gitPublishPush` fails - `bintrayUpload` won't start, the build fails
 -  if `gitPublishPush` success, but `bintrayUpload` fails - print warning message and the build fails
 - if both success - the build success

andrii-kovalenko-ct:
@heshamMassoud 
> @andrii-kovalenko-ct also could you please adjust these cmds in the README: 

6c99781b

PR #140, opened 2017-10-10
137 variant update actions

### LINKED ISSUE #137: Two potential bugs in buildVariantsUpdateActions
I just noticed 2 potential bugs in the implementation of buildVariantsUpdateActions:

https://github.com/commercetools/commercetools-sync-java/blob/7e567c3555b868c87acc0e3b475be2784cd86466/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java#L386-L406

- [x] At line [399](https://github.com/commercetools/commercetools-sync-java/blob/1.0.0-M2-beta-2/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java#L399): `masterVariant.getKey()` the masterVariant key is not checked if it is blank (empty/null) as it is the case for other variants in the (loop below in the method). This causes having a null key in the map.

- [x] At line [402](https://github.com/commercetools/commercetools-sync-java/blob/1.0.0-M2-beta-2/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java#L402): the `newProduct.getMasterVariant()` could be `null`, which means a `null` would be added in `newAllProductVariants ` and this will cause a NPE when `buildRemoveVariantUpdateActions` is called at line [485](https://github.com/commercetools/commercetools-sync-java/blob/1.0.0-M2-beta-2/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java#L485) at `newVariant.getKey()` since newVariant will be `null`.

-----
- [x] Also please make the variable `masterVariant` at line [398](https://github.com/commercetools/commercetools-sync-java/blob/1.0.0-M2-beta-2/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java#L398) `final`, since its by convention in the library that we always use won't reassign a var then we declare it as `final`, otherwise I got confused that this var might be reassigned later..

- [x] Fix SKU verification in `buildChangeMasterVariantUpdateAction()`

andrii-kovalenko-ct:
fixes #137:
  - verify master variants/keys before `buildVariantsUpdateActions()` ([ProductUpdateActionUtils](https://github.com/commercetools/commercetools-sync-java/pull/140/files#diff-540361586e271c4157d1afe06f9c8636))
  - verify master variants/keys, new variant SKU before `buildChangeMasterVariantUpdateAction()` ([ProductUpdateActionUtils](https://github.com/commercetools/commercetools-sync-java/pull/140/files#diff-540361586e271c4157d1afe06f9c8636R537))
  - add null/empty corner test cases ([ProductUpdateActionUtilsTest](https://github.com/commercetools/commercetools-sync-java/pull/140/files#diff-8c90c82014038e4c9b8161fc777b48d8R167))

### DIFF
--- a/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
@@ -57,6 +57,7 @@ import static com.commercetools.sync.products.utils.ProductVariantUpdateActionUt
 import static com.commercetools.sync.products.utils.ProductVariantUpdateActionUtils.buildProductVariantPricesUpdateActions;
 import static com.commercetools.sync.products.utils.ProductVariantUpdateActionUtils.buildProductVariantSkuUpdateActions;
 import static java.lang.String.format;
+import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonList;
 import static java.util.Objects.nonNull;
@@ -68,11 +69,10 @@ import static org.apache.commons.lang3.StringUtils.isBlank;
 
 // TODO: Update JavaDocs according to changes, tests..
 public final class ProductUpdateActionUtils {
-    private static final String FAILED_TO_BUILD_VARIANTS_ATTRIBUTES_UPDATE_ACTIONS = "Failed to build "
-            + "setAttribute/setAttributeInAllVariants update actions for the attributes of a ProductVariantDraft on the"
-            + " product with key '%s'. Reason: %s";
     private static final String BLANK_VARIANT_KEY = "The variant key is blank.";
     private static final String NULL_VARIANT = "The variant is null.";
+    private static final String BLANK_OLD_MASTER_VARIANT_KEY = "Old master variant key is blank";
+    private static final String BLANK_NEW_MASTER_VARIANT_KEY = "New master variant null or has blank key";
 
     /**
      * Compares the {@link LocalizedString} names of a {@link ProductDraft} and a {@link Product}. The name of the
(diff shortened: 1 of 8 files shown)


src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
```diff
@@ -594,4 +610,49 @@ static AddVariant buildAddVariantUpdateActionFromDraft(@Nonnull final ProductVar
...
+     * Apply error message to the {@code syncOptions}, reporting the product key and {@code reason}
+     *
+     * @param product     product which has sync error
+     * @param reason      reason to specify in the error message.
+     * @param syncOptions {@link BaseSyncOptions#applyErrorCallback(String, Throwable) applyErrorCallback} holder
+     */
+    private static void handleBuildVariantsUpdateActionsError(@Nonnull final Product product,
+                                                              @Nonnull final String reason,
+                                                              @Nonnull final ProductSyncOptions syncOptions) {
+        syncOptions.applyErrorCallback(format("Failed to build update actions on the product with key '%s'. Reason: %s",
```
heshamMassoud: other update actions for the product will be built. Only variant-related update actions will be skipped. Please adjust error message accordingly.

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
```diff
@@ -68,11 +69,11 @@
...
 // TODO: Update JavaDocs according to changes, tests..
 public final class ProductUpdateActionUtils {
-    private static final String FAILED_TO_BUILD_VARIANTS_ATTRIBUTES_UPDATE_ACTIONS = "Failed to build "
-            + "setAttribute/setAttributeInAllVariants update actions for the attributes of a ProductVariantDraft on the"
-            + " product with key '%s'. Reason: %s";
     private static final String BLANK_VARIANT_KEY = "The variant key is blank.";
     private static final String NULL_VARIANT = "The variant is null.";
+    static final String BLANK_OLD_MASTER_VARIANT_KEY = "Old master variant key is blank";
+    static final String BLANK_NEW_MASTER_VARIANT_KEY = "New master variant null or has blank key";
+    static final String BLANK_NEW_MASTER_VARIANT_SKU = "New master variant has blank SKU";
```
heshamMassoud: Please add a trailing periods here.

src/test/java/com/commercetools/sync/products/utils/ProductUpdateActionUtilsTest.java
```diff
@@ -150,6 +164,49 @@ public void buildVariantsUpdateActions_doesNotRemoveMaster() throws Exception {
...
+        final List<UpdateAction<Product>> updateActions =
+            buildVariantsUpdateActions(productOld, productDraftNew, syncOptions, emptyMap());
+        assertThat(updateActions).isEmpty();
+        assertThat(errorsCatcher).hasSize(errorMessages.length);
+
+        // verify all expected error messages
+        for (int i = 0; i < errorMessages.length; i++) {
+            assertThat(errorsCatcher.get(i))
+                .contains(format("Failed to build update actions on the product with key '%s'", productOld.getKey()))
+                .contains(errorMessages[i]);
```
heshamMassoud: why not? :
````java
assertThat(errorsCatcher.get(i)).isEqualTo(
    format("Failed to build update actions on the product with key '%s'. Reason: %",
    productOld.getKey(), errorMessages[i]));
````

File: src/test/java/com/commercetools/sync/products/utils/ProductUpdateActionUtilsTest.java [hunk 4]
```diff
@@ -182,6 +239,35 @@ public void buildChangeMasterVariantUpdateAction_changesMasterVariant() throws E
     }
 
     @Test
+    public void buildChangeMasterVariantUpdateAction_withEmptyKey() throws Exception {
```
heshamMassoud: could you please follow the naming convention used for naming tests (it makes it much easier to know everything about the test without looking on the code):

_MethodName_StateUnderTest_ExpectedBehavior_

so in this case it would be `buildVariantsUpdateActions_withEmptyKey_ShouldNotBuildActionAndTriggerCallback`

src/test/java/com/commercetools/sync/products/utils/ProductUpdateActionUtilsTest.java
andrii-kovalenko-ct: i just wanted to have some freedom for the error message, so the error message should not be exactly like this, but contain some keywords in any order.
For example, when leading/trailing/redundant spaces/periods/commas are changed in the message, the test is still valid.

I'd say usually i make it even less precise, so smth like:
```java
 assertThat(errorsCatcher.get(i))
    .containsIgnoringCase("failed")
    .contains(productOld.getKey())
    .containsIgnoringCase(errorMessages[i]);
```

In this case changes like in [the comment above](https://github.com/commercetools/commercetools-sync-java/pull/140#discussion_r143719118) do not require test changes.

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
andrii-kovalenko-ct: 67e8aacb

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
andrii-kovalenko-ct: d519defc

src/test/java/com/commercetools/sync/products/utils/ProductUpdateActionUtilsTest.java
andrii-kovalenko-ct: 2ee29177

src/test/java/com/commercetools/sync/products/utils/ProductUpdateActionUtilsTest.java
heshamMassoud: but why is it better to make it flexible for change (for us devs) against having a strict validation of the actual error message? for example, if the logic fails and another error message different than the one that should be actually is triggered, with also the word "failed" could pass the test when it shouldn't.

andrii-kovalenko-ct: like this: a601aa6a

andrii-kovalenko-ct: but i verify exact error messages `BLANK_OLD_MASTER_VARIANT_KEY`, `BLANK_NEW_MASTER_VARIANT_KEY`, `BLANK_NEW_MASTER_VARIANT_SKU` and so on

Issue #120, opened 2017-09-29
Sync Products TaxCategory


heshamMassoud:
### Description

Sync Products TaxCategory

- [x] Sync product TaxCategory (reference)
- [x] Sync product State
- [x] Reference resolve product TaxCategory
- [x] Reference resolve product State
- [x] Provide utils to prepare product draft for TaxCategory reference resolution (replace keys into id field)
- [x] Provide utils to prepare product draft for State reference resolution (replace keys into id field)

PR #146, opened 2017-10-11
#120: re-factor reference resolver to work with draft builder instead…

andrii-kovalenko-ct:
Short PR, to avoid mess in the next PRs i decided to make this change as a separate one:

  - re-factor `ProductReferenceResolver` to work with `ProductDraftBuilder` instance, instead of re-creating it every reference resolve function


src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
```diff
@@ -184,35 +183,30 @@ public ProductReferenceResolver(@Nonnull final ProductSyncOptions productSyncOpt
      * in the CTP project, the resultant draft would remain exactly the same as the passed product draft
      * (without reference resolution).
      *
-     * @param productDraft the product draft to resolve it's category references.
+     * @param draftBuilder the product draft to resolve it's category references.
```
heshamMassoud: It's a private JavaDoc, but please update it according to the change anyways, since its there

andrii-kovalenko-ct: 9dc2261e

andrii-kovalenko-ct:
> Would appreciate it also if this change is applied on all `..RefererenceResolvers`, if you think it's too much work for now, we can create an issue.

b4fc61eb

Since `CategoryDraftBuilder` and `InventoryEntryDraftBuilder` don't have generated getters - we can't easily use the same approach as in `ProductReferenceResolver`, the only one approach - to pass every time both references. 

Do you think we should use this approach now? 

Meanwhile i've started the discussion with JVM SDK developers to add respective getters to those draft builders

Issue #143, opened 2017-10-11
Remove invalidateCache in some services


heshamMassoud:
`invalidateCache` in TypeService and ChannelService is not needed and should be removed.

Issue #94, opened 2017-09-15
Refetch Concurrency modification retry in category sync


heshamMassoud:
Refetch a new version of the category on a Concurrency modification retry in category sync

PR #149, opened 2017-10-16
94 refetch category

heshamMassoud:
##### These 3 points address #94:
1. Add fetchCategory Service -> [interface](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-9aed284cb760f43ed51bae38ae986a15) | [implementation](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-38b71d2c035833260351074420bf8067) | [Integration Tests](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-bed84ab38f2508a9c2d8af75160ba1a2)

2. Add refetch before retry on concurrency modification exception: [Implementation](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-44d2fa184e625bcc0900b105f60d8677) | [Integration Tests](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-7a2564580078e482081eea430630ad3d)

3. Add release notes entry for bugfix: f9d9e1f

##### This point addresses #143:

1. Remove unneeded cache invalidation in some services:
 - Type Service:  [Interface](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-c1eac172805b03137caccf9c30e1f89c) | [Implementation](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-0e8da0bdc02e96b238728eec85a7ca87) | [Integration Tests](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-8859cb53896161a997ca23193c3ec749)
 - Channel Service:  [Interface](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-fe0bf238b4d3313bf27a01e4da9e3da6) | [Implementation](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-87e0a1f00ebdcae0ead2786d7c69a993) | [Integration Tests](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-187bbb96c2eda8e29e7037101dbe5a17)
 - ProductType service: [make cache invalidation implementation consistent with other services](https://github.com/commercetools/commercetools-sync-java/pull/149/files#diff-70764a3f46f474a3f4c5462050af43e1)

#### Relevant Issues
#94 #143


src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -607,16 +608,28 @@ static boolean requiresChangeParentUpdateAction(@Nonnull final Category category
         }
     }
 
+    private CompletionStage<Void> fetchAndUpdate(@Nonnull final Category oldCategory,
+                                                 @Nonnull final CategoryDraft newCategory) {
+        final String key = oldCategory.getKey();
```
andrii-kovalenko-ct: `getKey()` is nullable, but `fetchCategory()` category requires non-null. Either verify before supply, or make `fetchCategory()` accept nullable (and directly return empty optional in this case).
IMHO the last approach is more flexible.

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -607,16 +608,28 @@ static boolean requiresChangeParentUpdateAction(@Nonnull final Category category
...
+        return categoryService.fetchCategory(key)
+                              .thenCompose(categoryOptional -> {
+                                  if (categoryOptional.isPresent()) {
```
andrii-kovalenko-ct: `isPresent()` condition is usually used wrong. I'd recommend in this case to use more desired approach:
```java
categoryOptional
    .map(fetchedCategory -> buildUpdateActionsAndUpdate(fetchedCategory, newCategory))
    .orElseGet(() -> {
        handleError(format(UPDATE_FAILED, key, FETCH_ON_RETRY), null);
        return CompletableFuture.completedFuture(null);
    })
```

src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
```diff
@@ -92,6 +93,22 @@ public CategoryServiceImpl(@Nonnull final BaseSyncOptions syncOptions) {
 
     @Nonnull
     @Override
+    public CompletionStage<Optional<Category>> fetchCategory(@Nonnull final String key) {
+        return syncOptions.getCtpClient()
+                .execute(CategoryQuery.of().plusPredicates(categoryQueryModel -> categoryQueryModel.key().is(key)))
+                .thenApply(PagedResult::head)
+                .handle((fetchedProductOptional, sphereException) -> {
```
andrii-kovalenko-ct: why not just `exceptionally()`:
```java
.exceptionally(sphereException -> {
    syncOptions.applyErrorCallback(format(FETCH_FAILED, key, sphereException), sphereException);
    return Optional.empty();
});
```

src/main/java/com/commercetools/sync/categories/CategorySync.java
heshamMassoud: I was thinking about this actually. However, this method is a private helper method and is used in execution where key won't be `null`. So I wasn't sure:
1. Why check for `null`, when we are sure at this point that it is not `null` because its checked earlier on.. Or
2. Check anyways, just in case if the method is used for some other implementation in the future.
In general, What is the better approach in such cases?

andrii-kovalenko-ct: ok, i understand the approach with "we are sure at this point that it is not null". 

In the same it i'd _recommend_ (but not require) in general to implement the service more flexible, e.g. allow to accept **null** and return empty optional. For example, in most of our services in [`commercetools-paypal-plus-integration`](https://github.com/commercetools/commercetools-paypal-plus-integration/tree/master/src/main/java/com/commercetools/service/ctp/impl) project we accept `@Nullable`, but just return empty values in this case, so the future service consumers don't care about arguments verification.

andrii-kovalenko-ct:
> Use functional style of optional checks.
> Refactor unneeded retry flag and use functional style optional.

@heshamMassoud 
Looks like now you are enlightened by true functional approach

> Use exceptionally instead of handle.

and CompletionStage chains

👍

Issue #160, opened 2017-10-24
Support Product bundles 


heshamMassoud:
Support syncing attribute fields/ custom fields of type `ReferenceType` and `SetType` for `referenceTypeId`: `Product` only to support product bundles.

PR #167, opened 2017-10-26
160 product bundles

heshamMassoud:
#### Summary
Some products could reference other products through having one product [ReferenceType](http://dev.commercetools.com/http-api-projects-productTypes.html#referencetype)  or a set of product references [SetType](http://dev.commercetools.com/http-api-projects-productTypes.html#settype) as attribute types on their variants.

In order to sync products which have such references on their variants, reference resolution needs to be implemented in the sync library. 

This is what this PR addresses mainly. However, this is a beta feature, since it doesn't handle cases where the referenced product doesn't exist yet; i.e. It will only resolve the references if the referenced product is existing. 

#### Relevant Issues
#160 

#### Todo
 - [x] update release notes with compatibility notes, new feature and beta notice on feature.
- [x] Update Caveats sections in usage docs.

#### Hints for Review
1. Implementation of service: fetchCachedProductId in product service: 
[Interface](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-067e13f2fef8f22747c2534c0c8ed046) | [Impl](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-6dfa47f1d198543a343a810f8a0c0ab5) | [Integration test](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-4fcf28164f083f7b0c50764b1a938399)
2. Refactoring of [ProductRefernceResolver](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-0a79a67e66112ba2c9dd95da412f7048) into [VariantRefernceResolver](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-925cd7aab9dfa200affec5299f40b548).
3. Providing Reference replacement utils while also refactoring [ProductReferenceReplacementUtils](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-3efc52803f815cf85a34d7bc93cc0dbe) into [VariantReferenceReplacementUtils](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-c39e1d4e74d08c6b71a90fec82c72826).
4. Provided unit tests for all 4 classes mentioned in points 2 and 3.
5. Provided [integration test](https://github.com/commercetools/commercetools-sync-java/pull/167/files#diff-c055a497d0964e42b389ba30980944ba) for sync scenario

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
@@ -29,14 +29,13 @@ import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
-import java.util.function.BiConsumer;
-import java.util.function.Consumer;
-import java.util.function.Function;
 
 import static com.commercetools.sync.integration.commons.utils.CategoryITUtils.OLD_CATEGORY_CUSTOM_TYPE_KEY;
 import static com.commercetools.sync.integration.commons.utils.CategoryITUtils.OLD_CATEGORY_CUSTOM_TYPE_NAME;
@@ -149,20 +148,21 @@ public class ProductSyncIT {
     }
 
     private ProductSyncOptions buildSyncOptions() {
-        final BiConsumer<String, Throwable> errorCallBack = (errorMessage, exception) -> {
-            errorCallBackMessages.add(errorMessage);
-            errorCallBackExceptions.add(exception);
-        };
-        final Consumer<String> warningCallBack = warningMessage -> warningCallBackMessages.add(warningMessage);
-
         return ProductSyncOptionsBuilder.of(CTP_TARGET_CLIENT)
-                                        .setErrorCallBack(errorCallBack)
-                                        .setWarningCallBack(warningCallBack)
-                                        .setUpdateActionsFilterCallBack(builtActions -> {
-                                            updateActions.addAll(builtActions);
-                                            return builtActions;
-                                        })
+                                        .setErrorCallBack(this::errorCallback)
+                                        .setWarningCallBack(warningCallBackMessages::add)
+                                        .setUpdateActionsFilterCallBack(this::updateActionsCallback)
                                         .build();
+    }
+
+    private void errorCallback(@Nonnull final String errorMessage, @Nullable final Throwable exception) {
+        errorCallBackMessages.add(errorMessage);
+        errorCallBackExceptions.add(exception);
+    }
+
+    private List<UpdateAction<Product>> updateActionsCallback(@Nonnull final List<UpdateAction<Product>> builtActions) {
+        updateActions.addAll(builtActions);
+        return builtActions;
     }
 
     @AfterClass
@@ -277,20 +277,11 @@ public class ProductSyncIT {
     @Test
     public void sync_withProductTypeReference_ShouldUpdateProducts() {
         // Create custom options with whitelisting and action filter callback..
-        final BiConsumer<String, Throwable> errorCallBack = (errorMessage, exception) -> {
-            errorCallBackMessages.add(errorMessage);
-            errorCallBackExceptions.add(exception);
-        };
-        final Consumer<String> warningCallBack = warningMessage -> warningCallBackMessages.add(warningMessage);
-        final Function<List<UpdateAction<Product>>, List<UpdateAction<Product>>> actionCollector = builtActions -> {
-            updateActions.addAll(builtActions);
-            return builtActions;
-        };
         final ProductSyncOptions customSyncOptions =
             ProductSyncOptionsBuilder.of(CTP_TARGET_CLIENT)
-                                     .setErrorCallBack(errorCallBack)
-                                     .setWarningCallBack(warningCallBack)
-                                     .setUpdateActionsFilterCallBack(actionCollector)
+                                     .setErrorCallBack(this::errorCallback)
+                                     .setWarningCallBack(warningCallBackMessages::add)
+                                     .setUpdateActionsFilterCallBack(this::updateActionsCallback)
                                      .setSyncFilter(SyncFilter.ofWhiteList(ATTRIBUTES))
                                      .build();
         final ProductSync customSync = new ProductSync(customSyncOptions);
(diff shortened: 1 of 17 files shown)


src/main/java/com/commercetools/sync/products/helpers/VariantReferenceResolver.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private JsonNode createProductReferenceJson(@Nonnull final String productId) {
+        final ObjectNode productReferenceJsonNode = JsonNodeFactory.instance.objectNode();
+        productReferenceJsonNode.put("id", productId);
+        productReferenceJsonNode.put("typeId", "product");
+        return productReferenceJsonNode;
+    }
+}
+
```
andrii-kovalenko-ct: i found 3 duplications in this file, created a patch and sent you. Pls, review and tell me WDYT (merged most of `allOf/map/stream` futures)

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @SuppressWarnings("ConstantConditions") // NPE cannot occur due to being checked in replaceReferenceIdWithKey
+    static Optional<Set<Reference<Product>>> replaceAttributeReferenceSetIdsWithKeys(
+        @Nonnull final Attribute attribute) {
+        return getProductReferenceSet(attribute).map(productReferenceSet ->
+            productReferenceSet.stream()
+                               .map(productReference ->
+                                   replaceReferenceIdWithKey(productReference,
+                                       () -> Product.referenceOfId(productReference.getObj().getKey())))
```
andrii-kovalenko-ct: Usually it is better, instead of Supplier with closure, to have a normal function, e.g.:
`SyncUtils`:
```java

@Nullable
public static <T> Reference<T> replaceReferenceIdWithKey(
        @Nullable final Reference<T> reference,
        @Nonnull final Function<Reference<T>, Reference<T>> keyInReferenceSupplier) {
    if (reference != null && reference.getObj() != null) {
        return keyInReferenceSupplier.apply(reference);
    }
    return reference;
}
```
`VariantReferenceReplacementUtils`
```java
replaceReferenceIdWithKey(productReference,
        ref -> Product.referenceOfId(ref.getObj().getKey())))
```

Try to avoid closures as much as possible: they used to be slower and heavier (require more heap and stack to save environmen)

src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
```diff
@@ -141,6 +158,10 @@ private ProductSyncOptions buildSyncOptions() {
         return ProductSyncOptionsBuilder.of(CTP_TARGET_CLIENT)
                                         .setErrorCallBack(errorCallBack)
                                         .setWarningCallBack(warningCallBack)
+                                        .setUpdateActionsFilterCallBack(builtActions -> {
+                                            updateActions.addAll(builtActions);
+                                            return builtActions;
+                                        })
```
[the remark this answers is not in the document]
heshamMassoud: So actually what is needed to be tested here is what update actions are built by the sync process under test. And the filter function is used for that. It just adds to this mutated list the list of update actions. 
The filter function option is just used to make a faster/easier and also more strict assertion after the sync.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
```diff
@@ -67,27 +66,30 @@
...
-                                    @Nonnull final StateService stateService) {
+                                    @Nonnull final StateService stateService,
+                                    @Nonnull final ProductService productService) {
         super(productSyncOptions);
         this.productTypeService = productTypeService;
         this.categoryService = categoryService;
         this.taxCategoryService = taxCategoryService;
         this.stateService = stateService;
-        this.priceReferenceResolver = new PriceReferenceResolver(productSyncOptions, typeService, channelService);
+        this.variantReferenceResolver =
```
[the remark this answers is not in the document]
heshamMassoud: I don't really get your suggestion. Would appreciate it if you create an issue describing your proposed solution.

src/main/java/com/commercetools/sync/products/helpers/VariantReferenceResolver.java
heshamMassoud: thanks for the proposal f996ed7

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
```diff
@@ -0,0 +1,164 @@
...
+        final List<Price> variantPrices = productVariant.getPrices();
+        final List<PriceDraft> variantPriceDraftsWithKeys = new ArrayList<>();
+        variantPrices.forEach(price -> {
+            final Reference<Channel> channelReferenceWithKey = replaceChannelReferenceIdWithKey(price);
+            final PriceDraft priceDraftWithKey = PriceDraftBuilder.of(price)
+                                                                  .channel(channelReferenceWithKey).build();
+            variantPriceDraftsWithKeys.add(priceDraftWithKey);
+        });
+        return variantPriceDraftsWithKeys;
+    }
```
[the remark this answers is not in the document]
heshamMassoud: I am not sure why used foreach instead of map/collect 😄 
5c7d584

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
heshamMassoud: @andrii-kovalenko-ct ok, I didn't know about this. I would appreciate if you provide a reference to read about this. Because I don't see any other reason to not use supplier here.

heshamMassoud:
@andrii-kovalenko-ct about the test setup and naming: https://github.com/commercetools/commercetools-sync-java/issues/171

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
andrii-kovalenko-ct: well, i don't know any pure article about this, but this is just a common concept:
https://en.wikipedia.org/wiki/Closure_(computer_programming)

  - closures are created with _environment scope_, e.g. they _capture_ (copy) all the values they will be using when called, pretty much same like function calls put to stack everything they are depended on. That's why they are not absolutely _pure_: they capture and block values until they are removed. This might create heavy values in the memory, which are not called at all. In this certain case this closure captures _productReference_
  - opposite to above, pure functions don't capture environment, so they are more lightweight, more readable, more independent: when you read the function body you completely understand the logic and you don't need to understand in which environment the function is called - it depends only on input arguments.

In this certain case the difference (both from understandability and efficiency) is insignificant, this is only a recommendation for future.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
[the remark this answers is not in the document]
heshamMassoud: #174

PR #180, opened 2017-11-07
179 fix attribute reference replacement

heshamMassoud:
#### Summary
This PR addresses a bug in checking for product reference and product reference set attributes. Initially, to check if an attribute had a type product reference or a product reference set. This method `attribute.getValue(AttributeAccess.ofProductReference()) and attribute.getValue(AttributeAccess.ofProductReferenceSet())` was used from JVM SDK and checked if it throws exception, then its not any of these types.

However, these didn't throw exceptions for attributes of types localised string and actually parsed them to references with null field values. Which is wrong. Therefore, this PR the logic for checking for these attribute type is changed: [here](https://github.com/commercetools/commercetools-sync-java/pull/180/files#diff-c39e1d4e74d08c6b71a90fec82c72826)

The rest of the changes in the PR contains unit tests for all attributes types to check against.


src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
```diff
@@ -147,10 +145,25 @@
...
+    }
+
+    static boolean isProductReferenceSet(@Nonnull final Attribute attribute) {
+        final JsonNode valueAsJsonNode = attribute.getValueAsJsonNode();
+        return (valueAsJsonNode instanceof ArrayNode) && isValueAProductReference(valueAsJsonNode.elements().next());
+    }
+
+    private static boolean isValueAProductReference(@Nonnull final JsonNode valueAsJsonNode) {
+        if (valueAsJsonNode.isContainerNode()) {
+            final JsonNode typeIdNode = valueAsJsonNode.get("typeId");
```
lojzatran: Put these strings as constant fields of this class.

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
```diff
@@ -128,11 +129,8 @@
     }
 
     private static Optional<Reference<Product>> getProductReference(@Nonnull final Attribute attribute) {
-        try {
-            return Optional.of(attribute.getValue(AttributeAccess.ofProductReference()));
-        } catch (final JsonException exception) {
-            return Optional.empty();
-        }
+        return isProductReference(attribute)
+            ? Optional.of(attribute.getValue(AttributeAccess.ofProductReference())) : Optional.empty();
```
andrii-kovalenko-ct: more functional style of the implementation:
```java
Optional.of(attribute)
    .filter(VariantReferenceReplacementUtils::isProductReference)
    .map(a -> a.getValue(AttributeAccess.ofProductReference()));
```

same for [`getProductReferenceSet()`](https://github.com/commercetools/commercetools-sync-java/pull/180/files#diff-c39e1d4e74d08c6b71a90fec82c72826R148) below

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
andrii-kovalenko-ct: maybe (in another task) these kind of names should be carried out to some `CtpPlatformConstants` or something similar. Same for `"product"` references.

heshamMassoud: why? when it's only used here?

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
```diff
@@ -147,10 +145,25 @@
...
+
+    static boolean isProductReferenceSet(@Nonnull final Attribute attribute) {
+        final JsonNode valueAsJsonNode = attribute.getValueAsJsonNode();
+        return (valueAsJsonNode instanceof ArrayNode) && isValueAProductReference(valueAsJsonNode.elements().next());
+    }
+
+    private static boolean isValueAProductReference(@Nonnull final JsonNode valueAsJsonNode) {
+        if (valueAsJsonNode.isContainerNode()) {
+            final JsonNode typeIdNode = valueAsJsonNode.get("typeId");
+            return typeIdNode != null && Objects.equals(typeIdNode.asText(), "product");
```
andrii-kovalenko-ct: No need to use `Objects.equals()` in this case: `"product".equals(typeIdNode.asText())` is sufficient enough since `"product"` is definitely not null

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
heshamMassoud: In general, I don't like having separate constant "container" files which get really big by time. Especially if it's not used in different places. I might move it to such file though if I need it again.

andrii-kovalenko-ct: vote up for @lojzatran comment. i checked - "typeId" is used 3 times in 2 classes

andrii-kovalenko-ct: ok, that was just a recommendation

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
heshamMassoud: 5be11f0

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
heshamMassoud: bec3c5d

src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
heshamMassoud: @lojzatran @andrii-kovalenko-ct 1b7c07b

PR #182, opened 2017-11-13
160 provide external product bundles it

heshamMassoud:
#### Summary

1. Provide an IT for syncing product bundles from an external resource: https://github.com/commercetools/commercetools-sync-java/compare/160-provide-external-product-bundles-IT?expand=1#diff-d68ebb35acffe45efc007ae884c446b0

2. Fix typo in release notes: https://github.com/commercetools/commercetools-sync-java/compare/160-provide-external-product-bundles-IT?expand=1#diff-aa75d2e7837774f5cb565717b6744a35

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncIT.java
@@ -521,9 +521,10 @@ public class ProductSyncIT {
         final ProductSync productSync = new ProductSync(syncOptions);
         final ProductSyncStatistics syncStatistics = executeBlocking(productSync.sync(batch));
 
-        assertThat(syncStatistics.getReportMessage())
-            .isEqualTo(format("Summary: %d products were processed in total (%d created, %d updated and %d products"
-                + " failed to sync).", 3, 0, 1, 2));
+        assertThat(syncStatistics.getProcessed()).isEqualTo(3);
+        assertThat(syncStatistics.getCreated()).isEqualTo(0);
+        assertThat(syncStatistics.getUpdated()).isEqualTo(1);
+        assertThat(syncStatistics.getFailed()).isEqualTo(2);
         assertThat(errorCallBackExceptions).hasSize(2);
         assertThat(errorCallBackMessages).hasSize(2);
         assertThat(errorCallBackMessages.get(0))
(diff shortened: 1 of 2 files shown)


src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncIT.java
```diff
@@ -582,4 +589,48 @@ public void sync_withSameDraftsWithChangesInBatch_ShouldRetryUpdateBecauseOfConc
...
+                                                                                 .masterVariant(masterVariant)
+                                                                                 .build();
+
+
+        final ProductSync productSync = new ProductSync(syncOptions);
+        final ProductSyncStatistics syncStatistics =
+            executeBlocking(productSync.sync(singletonList(productDraftWithProductReference)));
+
+        assertThat(syncStatistics.getReportMessage())
+            .isEqualTo(format("Summary: %d products were processed in total (%d created, %d updated and %d products"
```
andrii-kovalenko-ct: i'm really confused that we are testing report strings so strictly. I'd say testing reports should be more flexible, e.g. testing number of exceptions/errors/warning is correct - it is a strict info, but the report message itself should be customer specific, e.g. every customer should be able to format the report message as they want, using info from exception/error/warning lists.

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncIT.java
```diff
@@ -582,4 +589,48 @@ public void sync_withSameDraftsWithChangesInBatch_ShouldRetryUpdateBecauseOfConc
...
+        final ProductDraft productDraft = createProductDraftBuilder(PRODUCT_KEY_2_RESOURCE_PATH,
+            ProductType.referenceOfId(productType.getKey()))
+            .taxCategory(null)
+            .state(null)
+            .build();
+
+        // Creating the product reference attribute value
+        final ObjectNode productReference = JsonNodeFactory.instance.objectNode();
+        productReference.put(REFERENCE_TYPE_ID_FIELD, Product.referenceTypeId());
+        productReference.put(REFERENCE_ID_FIELD, product.getKey());
```
butenkor: Why don't you use native support of SDK to create Attributes of type reference instead of using abstract ObjectNode?

```
final AttributeDraft ofReference = AttributeDraft.of("product-reference", Reference.of(Product.referenceTypeId(), product.getKey()));

final AttributeDraft ofSetOfReferences = AttributeDraft.of("product-references", asSet(Reference.of(Product.referenceTypeId(), product.getKey())));

public static <T> Set<T> asSet(final T ... params) {
  return new LinkedHashSet<>(Arrays.asList(params));
}
```

heshamMassoud: 0c60e4e

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncIT.java
heshamMassoud: This report message is one defined by the sync library. This doesn't stop the user of the library from customizing their own message with the counters when they want. But I agree that it makes it easier for testing to assert on counters rather on the entire message string which could change later: 1bde451 For the rest of the assertions I made an issue for it to be done separatley: https://github.com/commercetools/commercetools-sync-java/issues/186

Issue #191, opened 2017-11-20
Fix lib version in CTP client User-Agent header


heshamMassoud:
Fix lib version in CTP client User-Agent header, right now it's showing `unspecified`:
````
commercetools-jvm-sdk/1.25.0 (AHC/2.0) Java/1.8.0_45-b14 (Mac OS X; x86_64) commercetools-sync-java/unspecified (+https://github.com/commercetools/commercetools-sync-java; +ps-dev@commercetools.com)
````

Also, remove e-mail. As it should be enough to create issues, if there are, on this GH repo.

Issue #176, opened 2017-11-03
Ensure old/existing resource and new resource var namings


heshamMassoud:
Refer to discussion: https://github.com/commercetools/commercetools-sync-java/pull/175#discussion_r148768256

ahmetoz:
It's hard to find know what is needed in this issue, it looks like it's not up to date. I am closing the issue.

PR #175, opened 2017-11-03
172 fix doc and make uncomment release notes to be published for release

heshamMassoud:
This PR addresses the following:
1. Fix two typos in the docs: https://github.com/commercetools/commercetools-sync-java/commit/9f065e5836882b0f83584425fc17c9ae7fce39c7
2. Adjust release notes entry and make it public: https://github.com/commercetools/commercetools-sync-java/pull/175/commits/78233d036f605aee48d5c7b238fca64daa294a56

PR #192, opened 2017-11-20
191: fix lib version header

heshamMassoud:
#### Summary
1. Fixes the `unspecified ` version in the header of the client requests using the sync lib: 
1521af7

2. Removes the emergency contact and website from the header: 0fcdd69
3. Other boy-scout-behaviour issues: 
    i. Add trailing dot to javadoc publishing commit message: e0b50c6
    ii. Improve var namings: 8818f71, a5004c5
4. Update release notes: 146f16b

#### Relevant Issues
#191

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
@@ -161,7 +161,7 @@ public class ProductSyncIT {
     }
 
     private List<UpdateAction<Product>> beforeUpdateCallback(@Nonnull final List<UpdateAction<Product>> updateActions,
-                                                             @Nonnull final ProductDraft productDraft,
+                                                             @Nonnull final ProductDraft newProductDraft,
                                                              @Nonnull final Product oldProduct) {
         this.updateActions.addAll(updateActions);
         return updateActions;
--- a/src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
@@ -7,6 +7,8 @@ import javax.annotation.Nullable;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 public class SyncSolutionInfo extends SolutionInfo {
+    static final String UNSPECIFIED = "unspecified";
+
     /**
      * Extends {@link SolutionInfo} class of the JVM SDK to append to the User-Agent header with information of the
      * commercetools-sync-java library
@@ -27,7 +29,7 @@ public class SyncSolutionInfo extends SolutionInfo {
         setVersion(solutionVersion);
     }
 
-    private boolean isAttributeUnspecified(@Nullable final String attributeValue) {
-        return isBlank(attributeValue) || "unspecified".equals(attributeValue);
+    static boolean isAttributeUnspecified(@Nullable final String attributeValue) {
+        return isBlank(attributeValue) || UNSPECIFIED.equals(attributeValue);
     }
 }
--- a/src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
@@ -2,7 +2,9 @@ package com.commercetools.sync.commons.utils;
 
 import io.sphere.sdk.client.SolutionInfo;
 
-import static java.util.Optional.ofNullable;
+import javax.annotation.Nullable;
+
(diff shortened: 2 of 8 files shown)


src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
```diff
@@ -12,19 +11,15 @@
...
-     * (+https://github.com/commercetools/commercetools-sync-java; +ps-dev@commercetools.com)}</p>
+     * {implementationTitle}/{implementationVersion}}</p>
      *
      */
     public SyncSolutionInfo() {
-        final String libraryName = "commercetools-sync-java";
         final String implementationTitle = ofNullable(getClass().getPackage().getImplementationTitle())
-            .orElse(libraryName);
+            .orElse("commercetools-sync-java");
         final String implementationVersion = ofNullable(getClass().getPackage().getImplementationVersion())
```
lojzatran: As we saw before, this doesn't work, right? `getImplementationVersion` always returns `unspecifed`?

src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
```diff
@@ -12,19 +11,15 @@
...
             .orElse("DEBUG-VERSION");
         setName(implementationTitle);
         setVersion(implementationVersion);
-        setWebsite(format("https://github.com/commercetools/%s", libraryName));
```
lojzatran: In the task, you wrote `remove email`, but you want to remove website too?

heshamMassoud: because in the JVM SDK if the emergency contact is not set, the website will not be used as well: https://github.com/commercetools/commercetools-jvm-sdk/blob/93df1a10f6ed9b9a598915f3fc6e1d63c880d690/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/UserAgentUtils.java#L51-L67

src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
heshamMassoud: you mean to check if it's unspecified? 
6950522

src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
```diff
@@ -161,7 +161,7 @@ private void errorCallback(@Nonnull final String errorMessage, @Nullable final T
     }
 
     private List<UpdateAction<Product>> beforeUpdateCallback(@Nonnull final List<UpdateAction<Product>> updateActions,
-                                                             @Nonnull final ProductDraft productDraft,
+                                                             @Nonnull final ProductDraft newProductDraft,
```
andrii-kovalenko-ct: isn't `newProductDraft` redundant?

src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
andrii-kovalenko-ct: you mean doesn't work in the tests? if you pack jar properly - it should work

src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
heshamMassoud: We decided to use `oldProduct` and `newProductDraft` as convention across whole code for clarity: https://github.com/commercetools/commercetools-sync-java/pull/175#discussion_r148768256  #176

src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
heshamMassoud: > wrong commit in the PR description?

No, it is correct this actually fixes it. Because the version variable is declared in `project-info.gradle` so it should be applied before `package.gradle`

src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductSyncIT.java
andrii-kovalenko-ct: my point was that the variable is not used, but i just missed missed that it is used as a function reference. so, everything is fine here

heshamMassoud:
1. Added a comment for the manifest attributes setting:570bfa8 
2. Created an issue for better handling of version (to handle cases if someone removes the version from manifest): https://github.com/commercetools/commercetools-sync-java/issues/193
cc: @butenkor , @andrii-kovalenko-ct , @lojzatran

Issue #183, opened 2017-11-13
Add optional callback "beforeCreateCallback"


butenkor:
### Description
As a sync user i would like to be able to change/extend POST payload to CTP API. Currently it is possible to provide custom update actions thorough `updateActionsCallback` which is called before product update but it is not possible to change/extend product payload/draft on product create request. 

### Current Behaviour
Product assets are not supported and there is no way to intercept sync process in order to add assets/modify payload for product create requests.

### Expected Behaviour
It should provide optional `beforeCreateCallback` which is called just before sync sends create product request to CTP API.

### Possible Solution
The logic for calculation of assets updates can be supplied thorough `updateActionsCallback` and for create it might be achieved thorough `beforeCreateCallback`. It might be reasonable to rename `updateActionsCallback` to `beforeUpdateCallback`.

butenkor:
Parameters for callbacks:
`public List<UpdateActions<Product>> beforeUpdateCallback(List updateActions, newProductDraft, oldProduct)`

`public ProductDraft beforeCreateCallback(newProductDraft)`

heshamMassoud:
- [ ] Template packaging structure should look as follows: 
    - `templates/callbacks/beforeupdatecallback`
    - `templates/callbacks/beforecreatecallback`
- [ ] Templates should have helper methods with flexible parameters (e.g. productType) and not exactly similar to the parameters of the callbacks (e.g. the TriFunction params)..

PR #195, opened 2017-11-28
183: before create callback

heshamMassoud:
#### Description

1. Add new option beforeCreateCallback which is a takes a resource draft as input and returns a resource draft as output in [BaseSyncOptionsBuilder](https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-e97d9dee48f3d47a48f638790db0b080)
2. Implement [applyBeforeCreateCallback](https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-fcfa7e849d6ef115ad3487c591665b65R262) method, which applies the callback if not null and returns empty optional when the return is null.
3. Document null returns on the builder option setters JavaDoc: 59eebe6
4. Apply the callback in the [category sync](https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-38b71d2c035833260351074420bf8067), [product sync](https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-6dfa47f1d198543a343a810f8a0c0ab5) and [inventory sync](https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-289577bd8801a785df7dbc410d795936)
5. Add ServiceUtils as refactoring by reuse resource creation code in services: https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-f955b6cb1721fa5809e24b413f410022
5. Add unit tests https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-39d4d8c07f41f09469c0ac642867c2d9, https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-f371c5384a3f5256934b80acfdc489ba
6. Add integration tests https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-4fcf28164f083f7b0c50764b1a938399, https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-d68ebb35acffe45efc007ae884c446b0
7. release notes: https://github.com/commercetools/commercetools-sync-java/pull/195/files#diff-aa75d2e7837774f5cb565717b6744a35

#### Relevant Issues
#183 

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [x] Documentation


heshamMassoud:
> Looks goods. Tests in another PR?

Thx for early review, but it is still WIP (sync process still doesn't use the callback) and tests will be still coming in this same PR 😉

src/main/java/com/commercetools/sync/services/ServiceUtils.java
```diff
@@ -0,0 +1,42 @@
...
+     * @param resourceDraft         draft to apply callback on and then create on CTP.
+     * @param syncOptions           the sync options in which the CTP client is defined on and the callback.
+     * @param createCommandFunction the create command query to create the resource on CTP.
+     * @param responseHandler       defines how to handle the response from CTP.
+     * @param <T>                   the type of the created resource.
+     * @param <S>                   the type of the draft resource to create.
+     * @return a future containing an optional which might contain the resource if successfully creater or empty
+     *         otherwise.
+     */
+    public static <T, S> CompletionStage<Optional<T>> applyCallbackAndCreate(
```
andrii-kovalenko-ct: why this can't be done inside `syncOptions`? Why we have so many static classes now, which accept `syncOptions`, and then call it's methods wrapping/unwrapping them to Optional. This reminds me old fashion _Procedure Oriented Programming_, like it was in C ages ago.....

src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
```diff
@@ -143,22 +145,30 @@ public CategoryServiceImpl(@Nonnull final BaseSyncOptions syncOptions) {
...
+            return Optional.empty();
+        };
+
+        @SuppressWarnings("ConstantConditions") //createdCategory is only invoked if not null.
+        final Supplier<Optional<Category>> onSuccess = () -> {
+            keyToIdCache.put(createdCategory.getKey(), createdCategory.getId());
+            return Optional.of(createdCategory);
+        };
+
+        return createdCategory != null ? onSuccess.get() : onFailure.get();
```
andrii-kovalenko-ct: looks like an overhead? You created two callback, called one of them immediately, and then forgot them both...
why not just pure old `if-then-else`

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
```diff
@@ -157,18 +159,26 @@ public ProductServiceImpl(@Nonnull final ProductSyncOptions syncOptions) {
...
+            return Optional.empty();
+        };
+
+        @SuppressWarnings("ConstantConditions") //createdProduct is only invoked if not null.
+        final Supplier<Optional<Product>> onSuccess = () -> {
+            keyToIdCache.put(createdProduct.getKey(), createdProduct.getId());
+            return Optional.of(createdProduct);
+        };
+
+        return createdProduct != null ? onSuccess.get() : onFailure.get();
```
andrii-kovalenko-ct: same here

heshamMassoud: 714ce79

src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
heshamMassoud: no idea why I left with suppliers when with normal `if-then-else` would have been simpler. I was refactoring a lot around it yesterday, I didn't notice: 714ce79

Thanks for noticing it 😄

src/main/java/com/commercetools/sync/services/ServiceUtils.java
heshamMassoud: Much better: 630e995 seems that I got used to static utils.

Issue #198, opened 2017-11-30
Consider refactoring all services to extend from BaseService


heshamMassoud:
Consider refactoring all services to extend from BaseService.

Currently ProductService and CategoryService are extending it. Later all other services should also extend it and be refacored for reuse.

heshamMassoud:
`InventoryService` still needs to also be made consistent with other services.

heshamMassoud:
released in [1.8.0](https://github.com/commercetools/commercetools-sync-java/releases/tag/1.8.0)

PR #199, opened 2017-11-30
183 Move applyCallbackAndCreate to BaseService

heshamMassoud:
#### Summary
1. Move applyCallbackAndCreate to BaseService: d720205
2.  Make CategoryServiceImpl and ProductServiceImpl extend from BaseService: 2ea9f78
3. Added issue to extend BaseService by all services: #198 

#### Relevant Issues
#183 #198


src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -0,0 +1,50 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+class BaseService<U, V> {
```
andrii-kovalenko-ct: As a discussion:
Don't you want to give some more reasonable abbreviation names for the generics? like:
`BaseService<Res, ResDraft>` 
or
`BaseService<ResType, ResDraftType>`
or as a new convention
`BaseService<ResG, ResDraftG>` (like, use `G` suffix for all generic names)

   Apparently, [_Oracle Java Convention_](https://docs.oracle.com/javase/tutorial/java/generics/types.html) recommends to use single uppercase characters (as you use), but also there are some open discussions on this topic:
https://stackoverflow.com/questions/2900881/generic-type-parameter-naming-convention-for-java-with-multiple-chars
https://dzone.com/articles/naming-conventions
https://stackoverflow.com/questions/6143402/breaking-java-generics-naming-convention

And honestly, i bias for more _readable_ names. This also helps to read the class methods, which accept/return generic types, as same as make usage search between classes/inheritors and so on.

(open a new team discussion on Github, as you recommended)?

andrii-kovalenko-ct: I'd limit the lower generic types bounds:
`BaseService<U extends io.sphere.sdk.models.Resource<U>, V>`

So in the future we could access `U` values  from base class for some basic features (`id`, `version`, `toReference`).

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
```diff
@@ -157,7 +152,7 @@ public ProductServiceImpl(@Nonnull final ProductSyncOptions syncOptions) {
     @Nonnull
     @Override
     public CompletionStage<Optional<Product>> createProduct(@Nonnull final ProductDraft productDraft) {
-        return syncOptions.applyCallbackAndCreate(productDraft, ProductCreateCommand::of, this::handleProductCreation);
+        return applyCallbackAndCreate(productDraft, ProductCreateCommand::of, this::handleProductCreation);
```
andrii-kovalenko-ct: Did you notice that now `handleProductCreation()` implementation could be moved from `ProductServiceImpl` and `CategoryServiceImpl` to base `BaseService` class?
And then you even don't need to pas this function in `applyCallbackAndCreate`, because it will be available from `BaseService`, where `applyCallbackAndCreate` is actually defined too.

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -0,0 +1,50 @@
...
+    final BaseSyncOptions<U, V> syncOptions;
+    boolean isCached = false;
+    final Map<String, String> keyToIdCache = new ConcurrentHashMap<>();
```
andrii-kovalenko-ct: do we really expect to have access `keyToIdCache` in regular concurrent read/write mode, so it is implemented as `ConcurrentHashMap`?

src/main/java/com/commercetools/sync/services/impl/BaseService.java
heshamMassoud: @andrii-kovalenko-ct I have previously researched this in the beginning and have decided to follow the Oracle convention you have linked. Didn't much think about that it could be confusing. But to be honest, lately I have sometimes found it a bit confusing to know what the generic type refers to in some cases and I had to refer to a place where the generic type is used on an argument for example: @Nonnull final V resourceDraft then I would be able to know that V refers to a type that should be a draft. So I agree with you. But also I am not sure if it would also be too confusing (as Oracle states their argument) that it would be hard to know what is a generic type and what is a Java Object. I know IDEA differentiates it by colour (but is this enough to identify right away what it is) also shall we depend on the IDE that it will differentiate them for all devs of the lib?
Not sure what's the ideal why, maybe it makes sense to get majority opinion from the team, and get different opinions on the topic.

andrii-kovalenko-ct: could we move the discussion to that new github "flooding" feature?

butenkor: Java doc would be enough for me.

andrii-kovalenko-ct: here is question is not about documentation, but about design: it is usually recommended to keep the type as narrow as possible, so in future extending the functionality is less painful, e.g. we limit that the `BaseService` is intended to works only with CTP Resource Entities (Products, Categories, Inventories etc), but not with other types (Strings, Socket connections whatever)

andrii-kovalenko-ct: https://github.com/orgs/commercetools/teams/professional-services/discussions/2

heshamMassoud: https://github.com/commercetools/commercetools-sync-java/pull/199/commits/a90e5788e870b3362f920fc556b1185d977f2029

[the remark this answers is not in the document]
heshamMassoud: https://github.com/commercetools/commercetools-sync-java/pull/199/commits/a90e5788e870b3362f920fc556b1185d977f2029

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
heshamMassoud: It wasn't that easy to do because the key field is not on any class that extends Resource. 
But: b79590f
unit tests to cover after refactoring: 4888c5f

src/main/java/com/commercetools/sync/services/impl/BaseService.java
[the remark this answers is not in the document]
andrii-kovalenko-ct: it's just a specific _not very OOP_ scope, so usually looks weird and suspicious. 

Also, if you make them package-private and not protected - you restrict our library consumers to extend the class in their's own way. And then consumers are also confused: if this property is not explicitly private - it is expected to be used in another classes, but, unfortunately, not in our package. So it forces customers to extend classes in our packages namespace (which is, of course, a bad idea any way, but at least we should be as explicit as possible showing consumers how we expect to use properties and methods).

The only one place where i could accept package-private scope as more or less suitable approach - testing some specific functionality, so the test class has an access to package-private members.

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
andrii-kovalenko-ct: https://github.com/commercetools/commercetools-sync-java/commit/b79590ffb17c5cfde11acaa6ba590cb1f3237fe5#commitcomment-25982659

src/main/java/com/commercetools/sync/services/impl/BaseService.java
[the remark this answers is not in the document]
heshamMassoud: But why is it weird and suspicious? and what does "not very OOP" mean? It is a method that is meant to be used by concrete service implementations in the same package as this same class **and not any other classes in different packages**. It's not meant to be used by external library users, it's internal for the sync process. So I see it as totally suitable to have package-private access. As I said, if in the future we need (internally) use it within other packages, then we can change it to protected.

heshamMassoud:
@andrii-kovalenko-ct I thought it might make sense to leave as a function since the key could change in the future how we access it per resource, for example, product would be by sku. But it actually doesn't make that much of a difference: 
2d276fc

src/main/java/com/commercetools/sync/services/impl/BaseService.java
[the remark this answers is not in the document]
andrii-kovalenko-ct: ok

andrii-kovalenko-ct:
but the you will do just `product.getSku()` :)

My point is: don't use function in place, where value is called always, or the value calculation is very cheap. Functional approach is intent to be used (usually):
  - in postponed operations (like in `CompletionStage` and streams chains, where the result/instance is still not available)
  - in optional operations (like in `Optional` or if/else conditions, where you are not sure wether a method should be called or not)

Issue #227, opened 2018-01-08
Version is not correctly injected in the user-agent field of the JVM SDK request headers.


heshamMassoud:
### Description

If the library's jar is used by shadowJar plugin it's manifest is ignored and thus causes some applications using the job to remove the libraries manifest files.

### Possible Solution
> you could create template class with place holder for version number and modify it before release with version taken from git tag env.

### Related issues
https://github.com/commercetools/commercetools-sync-java/issues/193

Issue #193, opened 2017-11-21
Fail if version is not specified in JAR Manifest attribute "Implementation-Version"


heshamMassoud:
### Current Behaviour
Currently the version of the library appended to the JVM SDK client header is fetched from the "Implementation-Version" attribute in the JAR Manifest. (see: SyncSolutionInfo.java)

If this attribute is null/ not set the implementation falls back to a value of `DEBUG_VERSION`. This is needed for the case of debugging the project from IDEA for example because there is not JAR manifest. However, this doesn't guard against the case if someone/dev removes the attribute "Implementation-Version" from the JAR manifest.

### Expected Behaviour
It should throw an exception if the field is not set and is run from a jar.

### Possible Solution
1. Persist the library properties in a properties file (containing name and version).
2. Have a gradle task that always runs when building the project and persisting the values of the properties in this file.
3. Throw an exception in SyncSolutionInfo if the properties are not set.

### Related Issues
#191

PR #228, opened 2018-01-08
227 fix version header

heshamMassoud:
#### Summary
Addresses mainly #227.

Instead of depending on JAR manifest for fetching library version and name. Which could be removed by the users of the library, especially in the case when packing it as a dependency in a fat jar, it's quite common that the lib's manifest is replaced with the original applications' manifest.  

Therefore, the SolutionInfo version is now injected by a [bash script](https://github.com/commercetools/commercetools-sync-java/compare/227-fix-version-header?expand=1#diff-3227a58e30c2a2f634e233273d9d282d) (which potentially takes the version from the tag name, if the build was caused by a tag)

#### More details
The template string [`#{LIB_VERSION}`](https://github.com/commercetools/commercetools-sync-java/compare/227-fix-version-header?expand=1#diff-ad4bfd7c01d882b2b6238a8c1c41fabaR10) in the SyncSolutionInfo.java is replaced in the bash script [here](https://github.com/commercetools/commercetools-sync-java/compare/227-fix-version-header?expand=1#diff-3227a58e30c2a2f634e233273d9d282dR15).

This `sed` replace command makes a backup of the original file after replacement, in case a dev runs the script locally. But it's also [ignored](https://github.com/commercetools/commercetools-sync-java/compare/227-fix-version-header?expand=1#diff-a084b794bc0759e7a6b77810e01874f2) from committing on the repo.

#### Relevant Issues
#227 #193 

#### Todo

### Update (FYI: @butenkor, @andrii-kovalenko-ct):

1. Now it is written as a gradle script: [version-setter.gradle](https://github.com/commercetools/commercetools-sync-java/pull/228/files#diff-1b999cc63dfa57989b83c72d820a1e9e)
2. Which is only called manually by travis before compiling (https://github.com/commercetools/commercetools-sync-java/pull/228/files#diff-354f30a63fb0907d4ad57269548329e3)


docs/RELEASE_NOTES.md
```diff
@@ -50,6 +50,13 @@
 [Commits](https://github.com/commercetools/commercetools-sync-java/compare/v1.0.0-M8...v1.0.0-M9) |
 [Javadoc](https://commercetools.github.io/commercetools-sync-java/v/v1.0.0-M9/) | 
 [Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/v1.0.0-M9)
+
+
+**Bug Fixes** (1)
+- **Commons** - Fixed library version in User-Agent headers of JVM SDK clients using the library. Now it's fetched
+ from the JAR manifest but injected from set-solutioninfo-version.sh [#227](https://github.com/commercetools/commercetools-sync-java/issues/227)
```
butenkor: -> now it is **not** fetched

heshamMassoud: 095cd6c

heshamMassoud:
@andrii-kovalenko-ct Is there another advantage to having it as a gradle script as opposed to bash script, other than for consistently having all scripts in one place/language?

If so, then we should also better change this gradle as well (https://github.com/commercetools/commercetools-sync-java/blob/master/travis-publish.sh) right?

.travis.yml
```diff
@@ -8,6 +8,7 @@ dist: trusty
 jdk:
 - openjdk8
 install: true # skips travis' default installation step which executes gradle assemble.
+before_script: ./gradlew injectVersion
```
butenkor: Why `inject` and not just `setReleaseVersion`?

build.gradle
```diff
@@ -21,4 +21,5 @@ apply from: "$rootDir/gradle-scripts/maven-publish.gradle"
 apply from: "$rootDir/gradle-scripts/bintray-publish.gradle"
 apply from: "$rootDir/gradle-scripts/oss-publish.gradle"
 apply from: "$rootDir/gradle-scripts/javadocs-publish.gradle"
-apply from: "$rootDir/gradle-scripts/execution-order.gradle"
\ No newline at end of file
+apply from: "$rootDir/gradle-scripts/execution-order.gradle"
+apply from: "$rootDir/gradle-scripts/version-setter.gradle"
```
butenkor: Also here: `set-release-version.gradle`

gradle-scripts/version-setter.gradle
```diff
@@ -0,0 +1,32 @@
...
+        def versionFileContents = new File(versionFile).text
+        def versionPlaceholder = '#{LIB_VERSION}'
+        def libVersion = 'dev-version'
+        def isPr = System.getenv('TRAVIS_PULL_REQUEST')
+        def tagName = System.getenv('TRAVIS_TAG')
+
+        if (!versionFileContents.contains(versionPlaceholder)) {
+            throw new InvalidUserCodeException("$versionFile does not contain the placeholder: $versionPlaceholder. " +
+                    "Please make sure the file contains the placeholder, in order for the version to be injected " +
+                    "correctly.")
```
butenkor: 👍

andrii-kovalenko-ct:
@heshamMassoud 

>  Is there another advantage to having it as a gradle script as opposed to bash script, other than for consistently having all scripts in one place/language?

Isn't it enough as an advantage? Then also:
  + having it is in gradle makes it platform independent, when bash scripts are platform dependent
  + bash language is less readable than groovy
  + less code because standard plugins incapsulate some logic
  + common plugins are pretty known by other developers, when custom script require time to investigate/understand how they work (that was the first thing i noticed: i need to read bash script, instead of some standard configuration, like in ant replace plugin)
  + common plugins are used/tested by hundreds of other users - bugs/vulnerabilities are discovered better/faster

> If so, then we should also better change this gradle as well (https://github.com/commercetools/commercetools-sync-java/blob/master/travis-publish.sh) right?

Nice point! I've been dreaming to re-factor this wasp nest since my first days in the company, but never had enough time. Also, as a best approach - we could try to create our own _travis publish_ gradle plugin. A good topic for the next hackaton :)

.travis.yml
```diff
@@ -8,7 +8,7 @@ dist: trusty
 jdk:
 - openjdk8
 install: true # skips travis' default installation step which executes gradle assemble.
-before_script: ./set-solutioninfo-version.sh
+before_script: ./gradlew injectVersion
```
andrii-kovalenko-ct: why not make it as a part of build tasks graph (e.g. using `dependsOn`, `must/shouldRunAfter`, `finalizedBy` and so on).

gradle-scripts/version-setter.gradle
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        if (isPr.equals("false") && tagName) {
```
andrii-kovalenko-ct: use _groove goodness_ :) :
```groovy
isPr == "false"
```
http://mrhaki.blogspot.de/2009/09/groovy-goodness-check-for-object.html

.travis.yml
heshamMassoud: Because I only want the release version to be replaced only on travis, but not locally on every dev's machine when they run the build locally.

gradle-scripts/version-setter.gradle
```diff
@@ -0,0 +1,32 @@
...
+            libVersion = tagName
+        }
+
+        if (libVersion) {
+            println "Injecting the version: '$libVersion' in $versionFile"
+            ant.replace(file: versionFile, token: versionPlaceholder, value: libVersion)
```
andrii-kovalenko-ct: why not default gradle [`ReplaceTokens`](https://docs.gradle.org/current/userguide/userguide_single.html#sec:filtering_files) ?

.travis.yml
andrii-kovalenko-ct: then i'd recommend to create a custom gradle task (for example, `ciBuild`) and run it from travis ([here instead of `clean build`](https://github.com/commercetools/commercetools-sync-java/blob/e6b6791d27e0e42febebf57b98d093adc0039267/.travis.yml#L12), avoiding `before_script` block

heshamMassoud: It is already a custom gradle task. I made it inline now: 62bddb4, but also applies to the `after_success` then, no?

gradle-scripts/version-setter.gradle
heshamMassoud: 168c067

.travis.yml
heshamMassoud: 168c067, ac9470f, 8379ce0

build.gradle
heshamMassoud: 8379ce0

heshamMassoud:
> Isn't it enough as an advantage?

gradle-scripts/version-setter.gradle
heshamMassoud: because, as I understand, it is used within the filter function when files are being copied. Whereas in this case I just want to replace in file, rather then replace while copying.

.travis.yml
andrii-kovalenko-ct: >but also applies to the after_success

good point!

andrii-kovalenko-ct:
by default tasks order in gradle build is undefined, so it might be that `build` is executed before `setReleaseVersion `, even if you specify them in command  line in explicit order.

andrii-kovalenko-ct:
https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:ordering_tasks

heshamMassoud:
ah, I didn't know that it also doesn't ensure order when running explicitly from CLI: 

FYI, Take a look at this tho: https://discuss.gradle.org/t/ordering-tasks-on-the-command-line-versus-using-mustrunafter/12017

> Command line ordering is essentially the same as shouldRunAfter 

But still defining explicit ordering is better than relying on CLI input. dc27248 Thanks for pointing it out.

gradle-scripts/version-setter.gradle
andrii-kovalenko-ct: ok

andrii-kovalenko-ct:
hm, i was sure command line order doesn't influence the order at all....

Issue #243, opened 2018-02-02
Update category synch documentation


katmatt:
The documentation for [commercetools category sync](https://github.com/commercetools/commercetools-sync-java/blob/master/docs/usage/CATEGORY_SYNC.md) isn't up to date, the code snippet:
```
final Logger logger = LoggerFactory.getLogger(MySync.class);
final CategorySyncOptions categorySyncOptions = CategorySyncOptionsBuilder.of(sphereClient)
                                                                          .setErrorCallBack(logger::error)
                                                                          .setWarningCallBack(logger::warn)
                                                                          .build();
```
doesn't compile anymore. 

You changed the names of the methods (and their signature) - which is a great idea 👍  But it would be nice if you could update the documentation 😄 

And I guess that this inconsistency exists in the documentation for the products and inventory entries syncher doc too 😉

heshamMassoud:
@katmatt Thanks for reporting this. :) Will fix it asap.

heshamMassoud:
Fixed in #244

Issue #242, opened 2018-01-29
Concurrency issue with Statistics class


heshamMassoud:
There is an apparent concurrency issue with the statistics of the sync library which is was noticed while running the [benchmarks results](https://commercetools.github.io/commercetools-sync-java/benchmarks/) on the [travis assertions](https://travis-ci.org/commercetools/commercetools-sync-java/jobs/333700827) of the benchmark branch, which showed that sometimes, the number of resources being synced were 9999 instead of 10000.

**update 1**
Assertions were made on the actual state of the CTP projects after running the benchmark test to check if the correct number of resources were actually synced to the CTP project. These assertions passed, which means that the problem is with the statistics counting. 

**update 2**
It was noticed that the product sync concurrently updates the sync statistics on updates: https://github.com/commercetools/commercetools-sync-java/blob/b61de4b097fa87cf487f17cecf4260f8045fcc81/src/main/java/com/commercetools/sync/products/ProductSync.java#L248

and the Statistics class is not thread-safe. Therefore, this is the likely cause of the wrong statistics.

heshamMassoud:
released in v1.0.0-M10

Issue #217, opened 2017-12-17
Bump all dependencies


heshamMassoud:
Bump all dependendant libraries' versions.

heshamMassoud:
Issue not needed for that, always bump asap.

Issue #157, opened 2017-10-24
Statistics changes


heshamMassoud:
Statistics object structure should look as follows: 
````JSON
{
  "reportMessage": "Summary: 500 categories were processed in total (0 created, 485 updated, 0 failed to sync and 0 categories with a missing parent).",
  "totalUpdated": 485,
  "totalCreated": 0,
  "totalFailed": 0,
  "totalProcessed": 500,
  "totalProcessingTimeInMillis": 397,
  "totalProcessingTime": "0d, 0h, 0m, 0s, 397ms",
  "totalCategoryKeysWithMissingParents": {},
  "latestBatchStats": {
    "processingTimeInMillis": 397,
    "processingTime": "0d, 0h, 0m, 0s, 397ms"
  }
}
````
#### Which means the following changes should be made on the current statistics object:
- [ ] Fields that should be removed: `latestBatchProcessingTimeInDays`, `latestBatchProcessingTimeInHours`, `latestBatchProcessingTimeInMinutes`, `latestBatchProcessingTimeInSeconds`.
- [ ] counter field names should be prefixed with `total` to make it explicit that it infers the total number of all batches.
- [ ] human readable field should be named: `totalProcessingTime`
- [ ] New object/container for time for latest batch: `latestBatchStats` which means all other counters/processing times refer to aggregated values.
- [ ] Add feature to reset statistics, this is to support clients who will use the library as Singleton instance of the sync process which is always running and waiting for new batches. This scenario means the clients might not need the aggregated values, which could be from months, as much as they would need the latest batch's statistics.

heshamMassoud:
related: https://github.com/commercetools/commercetools-sync-java/pull/156#issuecomment-338708996

heshamMassoud:
related: https://github.com/commercetools/commercetools-sync-java/pull/244#discussion_r167013654

heshamMassoud:
related: #23

heshamMassoud:
total time for all batch sync executions can be computed by aggregating every statistic time.

heshamMassoud:
Update
````json
{
  "reportMessage": "Summary: 500 categories were processed in total (0 created, 485 updated, 0 failed to sync and 0 categories with a missing parent).",
  "totalUpdated": 485,
  "totalCreated": 0,
  "totalFailed": 0,
  "totalProcessed": 500,
  "totalCategoryKeysWithMissingParents": {}
}

````

Decision to remove time computation since it can be already done by the sync user.

PR #244, opened 2018-02-06
242 fix statistics concurrency issue

heshamMassoud:
#### Description
The counters in BaseSyncStatistics weren't thread-safe, even though there were parts in the code where they were being modified in parallel by multiple threads like here: https://github.com/commercetools/commercetools-sync-java/blob/b61de4b097fa87cf487f17cecf4260f8045fcc81/src/main/java/com/commercetools/sync/products/ProductSync.java#L248
This PR addresses this issue by making the counters atomic integers instead of ints, some maps and that were also accessed in parallel were changed to the concurrency-supporting ones. Moreover, some documentation update was made.
 

#### Summary
- [Bump dependencies of assertJ, mockito and jacoco](https://github.com/commercetools/commercetools-sync-java/pull/244/files#diff-771e2d81321cde33fec338d366924359). Bumping the rest of the dependencies cause some breaking changes, so they will be bumped in a separate PR to avoid confusion for fixing what they break.
- Make statistics' counters in [BaseSyncStatistics](https://github.com/commercetools/commercetools-sync-java/pull/244/files#diff-dfda92e7db082817e6e5c96dd809efde) all `AtomicInteger`s instead of `int` to be thread-safe.
- Make some [sets and maps to be `ConcurrentHashMap`/`ConcurrentHashMap.newKeySet`](https://github.com/commercetools/commercetools-sync-java/pull/244/files#diff-44d2fa184e625bcc0900b105f60d8677) which are used concurrently in `CategorySync` to be thread-safe.
- [Update documentation](https://github.com/commercetools/commercetools-sync-java/pull/244/commits/233d5f0d19a2404d8530ddd306e7826474357013) according to #243 

#### Relevant Issues
#242 #217 #243 

- [x] Tests
- [x] Documentation

### DIFF
--- a/src/main/java/com/commercetools/sync/categories/CategorySync.java
+++ b/src/main/java/com/commercetools/sync/categories/CategorySync.java
@@ -48,15 +48,28 @@ public class CategorySync extends BaseSync<CategoryDraft, CategorySyncStatistics
     private final CategoryReferenceResolver referenceResolver;
 
     /**
-     * Maps and sets which are thread-safe.
+     * The following set ({@code processedCategoryKeys}) is thread-safe because it is accessed/modified in a concurrent
+     * context, specifically when updating products in parallel in
+     * {@link #updateCategory(Category, CategoryDraft, List)}. It also has a global scope across the entire sync
+     * process, which means the same instance is used across all batch executions.
      */
-    private final ConcurrentHashMap<String, List<String>> categoryKeysWithMissingParents = new ConcurrentHashMap<>();
-    private final ConcurrentHashMap.KeySetView processedCategoryKeys = ConcurrentHashMap.newKeySet();
-    private ConcurrentHashMap.KeySetView categoryKeysWithResolvedParents = ConcurrentHashMap.newKeySet();
+    private final ConcurrentHashMap.KeySetView<String, Boolean> processedCategoryKeys = ConcurrentHashMap.newKeySet();
+
+    /**
+     * The following set ({@code categoryKeysWithResolvedParents}) and map ({@code categoryDraftsToUpdate}) are
+     * thread-safe because they are accessed/modified in a concurrent context, specifically when updating products in
+     * parallel in {@link #updateCategoriesInParallel(Map)}. They have a local scope within every batch execution, which
+     * means that they are re-initialized on every {@link #processBatch(List)} call.
+     */
+    private ConcurrentHashMap.KeySetView<String, Boolean> categoryKeysWithResolvedParents =
+        ConcurrentHashMap.newKeySet();
     private ConcurrentHashMap<CategoryDraft, Category> categoryDraftsToUpdate = new ConcurrentHashMap<>();
 
     /**
-     * Sets which are not thread-safe.
+     * The following sets ({@code existingCategoryDrafts}, {@code newCategoryDrafts}, {@code referencesResolvedDrafts}
+     * and {@code categoryKeysToFetch}) are not thread-safe because they are accessed/modified in a
+     * non-concurrent/sequential context. They have a local scope within every batch execution, which
+     * means that they are re-initialized on every {@link #processBatch(List)} call.
      */
     private Set<CategoryDraft> existingCategoryDrafts = new HashSet<>();
     private Set<CategoryDraft> newCategoryDrafts = new HashSet<>();
@@ -95,7 +108,6 @@ public class CategorySync extends BaseSync<CategoryDraft, CategorySyncStatistics
         super(new CategorySyncStatistics(), syncOptions);
         this.categoryService = categoryService;
         this.referenceResolver = new CategoryReferenceResolver(syncOptions, typeService, categoryService);
-        this.statistics.setCategoryKeysWithMissingParents(categoryKeysWithMissingParents);
     }
 
     /**
@@ -322,13 +334,13 @@ public class CategorySync extends BaseSync<CategoryDraft, CategorySyncStatistics
      * @param parentKey   the key of the missing parent.
      */
     private void addCategoryKeyToMissingParentsMap(@Nonnull final String categoryKey, @Nonnull final String parentKey) {
-        final List<String> childCategoryKeys = categoryKeysWithMissingParents.get(parentKey);
+        final List<String> childCategoryKeys = statistics.getMissingParentCategoryChildrenKeys(parentKey);
         if (childCategoryKeys != null) {
             childCategoryKeys.add(categoryKey);
         } else {
             final ArrayList<String> newChildCategoryKeys = new ArrayList<>();
             newChildCategoryKeys.add(categoryKey);
-            categoryKeysWithMissingParents.put(parentKey, newChildCategoryKeys);
+            statistics.putMissingParentCategoryChildrenKeys(parentKey, newChildCategoryKeys);
         }
     }
 
@@ -365,7 +377,7 @@ public class CategorySync extends BaseSync<CategoryDraft, CategorySyncStatistics
         createdCategories.forEach(createdCategory -> {
             final String createdCategoryKey = createdCategory.getKey();
             processedCategoryKeys.add(createdCategoryKey);
-            final List<String> childCategoryKeys = categoryKeysWithMissingParents.get(createdCategoryKey);
+            final List<String> childCategoryKeys = statistics.getMissingParentCategoryChildrenKeys(createdCategoryKey);
             if (childCategoryKeys != null) {
                 for (String childCategoryKey : childCategoryKeys) {
                     categoryKeysWithResolvedParents.add(childCategoryKey);
@@ -425,31 +437,16 @@ public class CategorySync extends BaseSync<CategoryDraft, CategorySyncStatistics
                 })
                                   .orElseGet(() -> CategoryDraftBuilder.of(fetchedCategory));
             if (categoryKeysWithResolvedParents.contains(fetchedCategoryKey)) {
-                final String parentKey = getMissingParentKey(fetchedCategoryKey);
-                final String parentId = keyToIdCache.get(parentKey);
-                categoryDraftBuilder.parent(Category.referenceOfId(parentId));
+                statistics.getMissingParentKey(fetchedCategoryKey)
+                          .ifPresent(missingParentKey -> {
+                              final String parentId = keyToIdCache.get(missingParentKey);
+                              categoryDraftBuilder.parent(Category.referenceOfId(parentId));
+                          });
             }
             categoryDraftsToUpdate.put(categoryDraftBuilder.build(), fetchedCategory);
         });
     }
 
-
-    /**
-     * Given a categoryKey {@code childCategoryKey} this method, checks in the {@code categoryKeysWithMissingParents}
-     * if it exists as a child to a missing parent, and returns the key of that missing parent. Otherwise, it returns
-     * null.
-     * @param childCategoryKey key of the category to look if it has a missing parent.
-     * @return the key of the parent category.
-     */
-    private String getMissingParentKey(@Nonnull final String childCategoryKey) {
-        return categoryKeysWithMissingParents.entrySet()
-                                             .stream()
-                                             .filter(missingParentEntry -> missingParentEntry.getValue().contains(
-                                                 childCategoryKey))
-                                             .findFirst()
-                                             .map(Map.Entry::getKey)
-                                             .orElse(null);
-    }
 
     /**
      * Given a {@link Set} of categories and a {@code key}. This method tries to find a category with this key in this
@@ -592,7 +589,7 @@ public class CategorySync extends BaseSync<CategoryDraft, CategorySyncStatistics
                                           processedCategoryKeys.add(categoryKey);
                                       }
                                       if (categoryKeysWithResolvedParents.contains(categoryKey)) {
-                                          removeUpdatedCategoryFromMissingParentsMap(categoryKey);
+                                          statistics.removeChildCategoryKeyFromMissingParentsMap(categoryKey);
                                       }
                                       return CompletableFuture.completedFuture(null);
                                   }
@@ -610,16 +607,6 @@ public class CategorySync extends BaseSync<CategoryDraft, CategorySyncStatistics
                                     handleError(format(UPDATE_FAILED, key, FETCH_ON_RETRY), null);
                                     return CompletableFuture.completedFuture(null);
                                 }));
-    }
-
-    /**
-     * Given a {@code categoryKey} this method removes its occurrences from the map
-     * {@code categoryKeysWithMissingParents}.
-     *
-     * @param categoryKey the category key to remove from {@code categoryKeysWithMissingParents}
-     */
-    private void removeUpdatedCategoryFromMissingParentsMap(@Nonnull final String categoryKey) {
-        categoryKeysWithMissingParents.forEach((key, value) -> value.remove(categoryKey));
     }
 
     /**
--- a/src/main/java/com/commercetools/sync/categories/CategorySync.java
+++ b/src/main/java/com/commercetools/sync/categories/CategorySync.java
@@ -17,7 +17,6 @@ import io.sphere.sdk.commands.UpdateAction;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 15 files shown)


src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
```diff
@@ -16,7 +16,7 @@
      * Map that represents categories with missing parents; the keys of the map are the keys of the missing parent
      * categories and the value of each is a list of the children category keys.
      */
-    private Map<String, List<String>> categoryKeysWithMissingParents = new HashMap<>();
+    private Map<String, List<String>> categoryKeysWithMissingParents = new ConcurrentHashMap<>();
```
lojzatran: Suggestion: I think it's better when you write comment for Map like this:
```
/**
     * Map that represents categories with missing parents
     *
     * key: keys of the missing parent categories
     * value: list of the children category keys
     */
```
The idea is to make clear what is key and what is value and not hide it in a long sentence.

lojzatran: Why do you actually need `ConcurrentHashMap` in this class? I don't see you put anything into this map.

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -186,11 +187,13 @@ private int getNumberOfProcessedCategories(@Nonnull final List<CategoryDraft> ca
...
-                                       .filter(Objects::nonNull)
-                                       .map(CategoryDraft::getKey)
-                                       .filter(categoryDraftKey -> !processedCategoryKeys.contains(categoryDraftKey))
-                                       .collect(Collectors.toList()).size();
+        final int numberOfCategoryDraftsNotProcessedBefore =
+            categoryDrafts.stream()
+                          .filter(Objects::nonNull)
+                          .map(CategoryDraft::getKey)
+                          .filter(categoryDraftKey ->
+                              categoryDraftKey == null || !processedCategoryKeys.contains(categoryDraftKey))
```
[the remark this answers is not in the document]
heshamMassoud: This null check was made to avoid hitting an NPE in `processedCategoryKeys.contains(categoryDraftKey)` if `categoryDraftKey` is `null`. Since `processedCategoryKeys` is now a `KeySetView` and this throws an NPE on it's contains method if the passed arg is null, as opposed to what it was previously (`HashSet`).

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
heshamMassoud: @lojzatran https://github.com/commercetools/commercetools-sync-java/blob/master/src/main/java/com/commercetools/sync/categories/CategorySync.java#L91 here it is set in the `CategorySync` class, and elements are put into this map concurrently in the [`CategorySync`](https://github.com/commercetools/commercetools-sync-java/blob/master/src/main/java/com/commercetools/sync/categories/CategorySync.java) class.

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -177,7 +178,7 @@ public CategorySync(@Nonnull final CategorySyncOptions syncOptions) {
     /**
      * Given a list of {@code CategoryDraft} elements. Calculates the number of drafts that need to be processed in this
      * batch, given they weren't processed before. Categories which were processed before, will have their keys stored
-     * in {@code processedCategoryKeys}. Added to the number of null categoryDrafts.
+     * in {@code processedCategoryKeys}. Added to the number of null categoryDrafts and categories with null keys.
```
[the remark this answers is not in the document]
heshamMassoud: 3d9aa0a

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
heshamMassoud: > The idea is to make clear what is key and what is value and not hide it in a long sentence.

Definitely agree:  e3f0609. Thanks for the suggestion!

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -144,7 +145,7 @@ public CategorySync(@Nonnull final CategorySyncOptions syncOptions) {
      */
     @Override
     protected CompletionStage<CategorySyncStatistics> processBatch(@Nonnull final List<CategoryDraft> categoryDrafts) {
-        final int numberOfNewDraftsToProcess = getNumberOfProcessedCategories(categoryDrafts);
+        final int numberOfNewDraftsToProcess = getNumberOfDraftsToProcess(categoryDrafts);
         referencesResolvedDrafts = new HashSet<>();
```
andrii-kovalenko-ct: 1. why all the earlier initialized as concurrent maps/sets here are reinitialized and non-concurrent? (see categoryKeysWithResolvedParents)

2. Why they are initialized twice (in the field declaration here)?

3. Looks like some fields (at least `existingCategoryDrafts`, didn't review other) should be converted to local variables, since they are used only in context of `processBatch()` function.

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -47,15 +48,15 @@
...
+    private final Set<String> processedCategoryKeys = ConcurrentHashMap.newKeySet();
 
     private Set<CategoryDraft> existingCategoryDrafts = new HashSet<>();
     private Set<CategoryDraft> newCategoryDrafts = new HashSet<>();
     private Set<CategoryDraft> referencesResolvedDrafts = new HashSet<>();
-    private Set<String> categoryKeysWithResolvedParents = new HashSet<>();
+    private Set<String> categoryKeysWithResolvedParents = ConcurrentHashMap.newKeySet();
     private Set<String> categoryKeysToFetch = new HashSet<>();
-    private Map<CategoryDraft, Category> categoryDraftsToUpdate = new HashMap<>();
+    private Map<CategoryDraft, Category> categoryDraftsToUpdate = new ConcurrentHashMap<>();
```
andrii-kovalenko-ct: guess it is better now to divide the declarations int concurrent and non-concurrent groups, otherwise hard to understand which maps/sets must be concurrent, which - not.

Maybe change the field type explicitly to _ConcurrentXXX_ so you can't re-initialize it with unsafe implementation, like in [line 152](https://github.com/commercetools/commercetools-sync-java/blob/51e530543b725d11734f70da9c412100c4b600d9/src/main/java/com/commercetools/sync/categories/CategorySync.java#L152)

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -144,7 +145,7 @@ public CategorySync(@Nonnull final CategorySyncOptions syncOptions) {
      */
     @Override
     protected CompletionStage<CategorySyncStatistics> processBatch(@Nonnull final List<CategoryDraft> categoryDrafts) {
```
andrii-kovalenko-ct: i'm confused how `processBatch` should be executed: concurrently or sequentially? If there are multiple batches, should they be processed one-by-one, or simultaneously?

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
```diff
@@ -4,19 +4,25 @@
...
+     * Map that represents categories with missing parents.
+     *
+     * <ul>
+     *     <li>key: key of the missing parent category</li>
+     *     <li>value: a list of the parent's children category keys</li>
+     * </ul>
+     *
      */
-    private Map<String, List<String>> categoryKeysWithMissingParents = new HashMap<>();
+    private Map<String, List<String>> categoryKeysWithMissingParents = new ConcurrentHashMap<>();
```
andrii-kovalenko-ct: is there any guarantee that `setCategoryKeysWithMissingParents()` below also will set it to concurrent implementation?

andrii-kovalenko-ct: can't understand why here the implementation should be concurrent? The map is stored in `CategorySyncStatistics`, but updated over getter in some other place? Looks like wrong design.

Maybe just `getCategoryKeysWithMissingParents()` should return an immutable copy/view of `categoryKeysWithMissingParents` instance?

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
andrii-kovalenko-ct: > here it is set in the CategorySync class, and elements are put into this map concurrently in the CategorySync class.

agree with @lojzatran, the implementation looks weird: the value is update in one class, but stored in another. Looks obfuscated and error prone.

src/main/java/com/commercetools/sync/categories/CategorySync.java
heshamMassoud: They should be processed one-by-one. It is not meant for batches to be processed concurrently.

src/main/java/com/commercetools/sync/categories/CategorySync.java
heshamMassoud: 1. That is actually my mistake, thanks for noticing it!
2. They are initialized twice because they need to be emptied on every batch processing.
3. They have global access, they are not only used here, but they are used by several methods.

src/main/java/com/commercetools/sync/categories/CategorySync.java
heshamMassoud: 955d5eb
 Good catch! Thanks a lot.

src/main/java/com/commercetools/sync/categories/CategorySync.java
andrii-kovalenko-ct: If so, i'm confused where in the method body/execution workflow concurrency occurs?

heshamMassoud: I added now JavaDocs on the concurrent/non-concurrent maps and sets.

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
heshamMassoud: Now it is changed to be ConcurrentHashMap as a field: https://github.com/commercetools/commercetools-sync-java/pull/244/files#diff-c0fbda50bb0857abc8f4ea35c6239db4R35

It is also now encapsulated in the Statistics and can only be modified from within the statistics.

heshamMassoud: @andrii-kovalenko-ct 
Now it is exactly changed like that: https://github.com/commercetools/commercetools-sync-java/pull/244/files#diff-c0fbda50bb0857abc8f4ea35c6239db4

It is also now a `ConcurrentMap<String, Set>` instead of `ConcurrentMap<String, List<String>>`

Also unit tests are added for the statistics class: https://github.com/commercetools/commercetools-sync-java/pull/244/files#diff-a89045e8bbd6c0998ee9220b42ca18ba

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatistics.java
```diff
@@ -21,7 +22,15 @@
     private String latestBatchHumanReadableProcessingTime;
```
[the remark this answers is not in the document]
heshamMassoud: 4e7f1b4

src/main/java/com/commercetools/sync/categories/helpers/CategorySyncStatistics.java
```diff
@@ -45,17 +61,70 @@ public String getReportMessage() {
...
-                                             .map(List::size)
-                                             .reduce(0, Integer::sum);
+                                             .mapToInt(Set::size)
+                                             .sum();
     }
 
-    public Map<String, List<String>> getCategoryKeysWithMissingParents() {
-        return categoryKeysWithMissingParents;
+    public Map<String, Set<String>> getCategoryKeysWithMissingParents() {
+        return Collections.unmodifiableMap(categoryKeysWithMissingParents);
```
andrii-kovalenko-ct: 👍

src/test/java/com/commercetools/sync/categories/helpers/CategorySyncStatisticsTest.java
```diff
@@ -76,40 +77,31 @@ public void getReportMessage_WithIncrementedStats_ShouldGetCorrectMessage() {
...
 
         categorySyncStatistics.setCategoryKeysWithMissingParents(categoryKeysWithMissingParents);
         assertThat(categorySyncStatistics.getNumberOfCategoriesWithMissingParents()).isEqualTo(0);
     }
 
     @Test
     public void getNumberOfCategoriesWithMissingParents_WithNonEmptyMap_ShouldReturnCorrectNumberOfChildren() {
-        final Map<String, List<String>> categoryKeysWithMissingParents = new HashMap<>();
-        final ArrayList<String> firstMissingParentChildrenKeys = new ArrayList<>();
+        final ConcurrentHashMap<String, Set<String>> categoryKeysWithMissingParents = new ConcurrentHashMap<>();
```
andrii-kovalenko-ct: 👍

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatistics.java
```diff
@@ -202,6 +225,8 @@ public String getLatestBatchHumanReadableProcessingTime() {
     /**
      * Gets the number of days it took to process.
      *
+     * <p>Note: This method isn't thread-safe and shouldn't be used in a concurrent context.
```
lojzatran: Why this method is not thread-safe?

andrii-kovalenko-ct: ah, @heshamMassoud Lam is right, the simple getters are thread-safe in this case:
they return primitives (or reference in case of String), so it's safe to call them.

heshamMassoud:
@andrii-kovalenko-ct
> Are this issues connected only to Categories sync? Products and Inventories don't need the concurrency? They are not hierarchical thus every item update might be independent, unless Products have references to another products? (like recommended or related products)

They also need concurrency, that's why the `AtomicInteger`s were introduced on the `BaseSyncStatistics`. The difference though with `CategorySyncStatistics` is that it had this `categoryKeysWithMissingParents` map that was accessed concurrently.

src/main/java/com/commercetools/sync/commons/helpers/BaseSyncStatistics.java
heshamMassoud: @lojzatran @andrii-kovalenko-ct ok,
1. but wouldn't the getter return not a correct value if called concurrently?
2. @andrii-kovalenko-ct Also, what do you mean "primitives -> so safe", so if the getter returns a `collection`/object it's not safe anymore?

andrii-kovalenko-ct: >but wouldn't the getter return not a correct value if called concurrently?

that's why you should decide what do you mean at "correct/not correct" values? Sequential calls of this method during batch processing could return different values, but they still all are correct in time they are called.

> Also, what do you mean "primitives -> so safe"

I was a bit wrong: `int` and `String` values are safe, because they are updated atomically, `long` (theoretically) won't be safe because it is multiword thus might require more than one tick to update entire value:
https://stackoverflow.com/a/9278798/907576

andrii-kovalenko-ct: Now i see the sync statistic became confusing with all this thread-specific staff.

Guess (might be in another task/PR) we have to use some synchronized builder and return immutable object which has some finished state, e.g. no one can call the statistic before it is finished completely, so no one gets uncompleted state of an update.

heshamMassoud: What I mean by not "correct" is that it might have been already set by another thread, and the current thread doesn't yet see this updated value. That's why I think the comment makes sense to leave it. It's not thread safe because that getter does not ensure that a thread will see the latest value, as the value may be stale. 

> Guess (might be in another task/PR) we have to use some synchronized builder and return immutable object which has some finished state, e.g. no one can call the statistic before it is finished completely, so no one gets uncompleted state of an update.

Yep, I agree. I linked it to another issue. So it can be also fixed in the statistics refactoring. #157

andrii-kovalenko-ct: :+1

PR #252, opened 2018-02-15
Safe null collections wrappers

andrii-kovalenko-ct:
Extend `CollectionUtils` with _safe wrappers_, which return empty collections (List, Set, HashMap) if the passed instance is **`null`**. 

This allows to avoid ternary operators.

  -  [new `CollectionUtil`](https://github.com/commercetools/commercetools-sync-java/pull/252/files#diff-c1854058c59ea8d4200582685a4ee2f5R97) functions
  - [Unit tests](https://github.com/commercetools/commercetools-sync-java/pull/252/files#diff-7e7c93e224538e2813ad8af3a08d4847R119)
  - re-factored sources avoiding ternary operators with collection checks.
  - re-factor some `Optionals` misuse:
    - https://github.com/commercetools/commercetools-sync-java/pull/252/files#diff-925cd7aab9dfa200affec5299f40b548L152
    - https://github.com/commercetools/commercetools-sync-java/pull/252/files#diff-a13d6c119192ae82d66e3a6c9028163bL143

@heshamMassoud 
This code https://github.com/commercetools/commercetools-sync-java/pull/252/files#diff-a13d6c119192ae82d66e3a6c9028163bL145 
was (and remains) confusing:
first it says _if old locale value is set_ and then (in the same condition branch without `oldLocaleValue` change) it says _If old locale value is not set_. Guess it should be clarified.

### DIFF
--- a/src/main/java/com/commercetools/sync/products/templates/beforeupdatecallback/SyncSingleLocale.java
+++ b/src/main/java/com/commercetools/sync/products/templates/beforeupdatecallback/SyncSingleLocale.java
@@ -142,7 +142,7 @@ final class SyncSingleLocale {
                 final String newLocaleValue = newLocalizedField.get(locale);
                 final String oldLocaleValue = oldLocalizedField.get(locale);
                 if (!Objects.equals(newLocaleValue, oldLocaleValue)) {
-                    // if old locale value is set, remove it from old localized field, otherwise use oldLocalizedField
+                    // if old locale value is set, remove it from old localized field
                     final LocalizedString withLocaleChange = ofNullable(oldLocaleValue)
                         .map(value -> LocalizedString.of(
                             oldLocalizedField.stream()
@@ -151,7 +151,8 @@ final class SyncSingleLocale {
                                     LocalizedStringEntry::getValue))))
                         .orElse(oldLocalizedField);
 
-                    // If old locale value is not set, only update if the new locale value is set.
+                    // Only if old locale value is not set and the new locale value is set,
+                    // update the old localized field with the new locale value
                     return ofNullable(newLocaleValue)
                         .map(val -> updateActionMapper.apply(withLocaleChange.plus(locale, val)));
                 }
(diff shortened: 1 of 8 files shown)


heshamMassoud:
> This code https://github.com/commercetools/commercetools-sync-java/pull/252/files#diff-a13d6c119192ae82d66e3a6c9028163bL145
was (and remains) confusing:
first it says if old locale value is set and then (in the same condition branch without oldLocaleValue change) it says If old locale value is not set. Guess it should be clarified.

You mean the comments are confusing? 

> first it says if old locale value is set 
This is the `then` part of the if condition (it unsets it inside, that's why the `if old locale value is set, remove it from old localized field.`

and that's why the comment after: 
`// If old locale value is not set, only update if the new locale value is set.`
is always valid, because at this point the old locale value is removed (whether from the `then` part of the if or from before)

src/main/java/com/commercetools/sync/products/templates/beforeupdatecallback/SyncSingleLocale.java
```diff
@@ -141,32 +142,31 @@
                 final String newLocaleValue = newLocalizedField.get(locale);
                 final String oldLocaleValue = oldLocalizedField.get(locale);
                 if (!Objects.equals(newLocaleValue, oldLocaleValue)) {
-                    LocalizedString withLocaleChange = oldLocalizedField;
-                    // if old locale value is set, remove it from old localized field.
-                    if (oldLocaleValue != null) {
-                        withLocaleChange = LocalizedString.of(
+                    // if old locale value is set, remove it from old localized field, otherwise use oldLocalizedField
```
heshamMassoud: I woudn't add the `otherwise use oldLocalizedField`, since there is another check that happens before the `oldLocalizedField` is actually used in the update action:

at the comment after:
`// If old locale value is not set, only update if the new locale value is set.`

Maybe, this bottom comment could be change to, to be more verbose:
`// Only if old locale value is not set and the new locale value is set, update the old localized field with the new locale value .`

andrii-kovalenko-ct: 752be475

PR #253, opened 2018-02-16
Support secondary resources in generic custom update action utils

heshamMassoud:
#### Summary

Custom fields don't only exist on primary resources (resources with their own endpoint e.g. Categories, Channels, Inventory Entry, etc..) but also on secondary resources (resources without an endpoint e.g. assets and prices). Therefore, the implementation for, `CustomUpdateActionUtils#buildCustomUpdateActions ` (which was a util for building up update actions for any two (primary) resources which contained custom fields), needs to change to support also secondary resources that might contain custom fields. 

Also, with custom update actions on secondary resources (e.g. assets and prices), more information is required like the variant id and the asset/price id. Therefore, all this information is needed to be passed down to the custom update action utils to be able to build the update actions correctly.

#### Description and Hints for Review
1. [Bump JVM SDK to 1.27.0 for asset support](https://github.com/commercetools/commercetools-sync-java/pull/253/files#diff-771e2d81321cde33fec338d366924359).
2. [`GenericCustomActionBuilder`](https://github.com/commercetools/commercetools-sync-java/pull/253/files#diff-5d94e7e12343c97b821792be3acc151f) is now refactored:-
   a. Generic Type T [now only extends `Resource<T>`](https://github.com/commercetools/commercetools-sync-java/pull/253/files#diff-5d94e7e12343c97b821792be3acc151fR19) but not `Custom`, since the resource getting updated might not be the resource that actually has the custom fields, but could just be a container resource, for example: A Product containing a price. In this case the price is the resource containing the custom fields and the Product is the one the update action should be of and the one containing the price.
  b. All the three abstract methods now take more optional information that could be about the secondary resource (`variantId` and `objectID`)
[...]

PR #254, opened 2018-02-16
Assets reference resolution

heshamMassoud:
#### Summary

This PR addresses reference resolution for assets. Custom Type on assets require reference resolution.

#### Description and Hints for Review
1. [Bump JVM SDK to 1.27.0 for asset support](https://github.com/commercetools/commercetools-sync-java/pull/253/files#diff-771e2d81321cde33fec338d366924359).
2. Add [`AssetReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-df287f7153b24745f621ca7ddd024828)
3. Use `AssetReferenceResolver` in [`CategoryReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-7af94039b5018756bcec77f6203a1609) and [`VariantReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-925cd7aab9dfa200affec5299f40b548).
4. Add unit tests for [`AssetReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-8524c3d1aed7b2d0eaeefbf19d9c9de1), [`CategoryReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-735b7c3be0a73955b821dde37f27d824) and [`VariantReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-9648745cd838909318a72e71d6ee15a5).

#### Boy-scout changes:
1. [Use JVM SDK `CompletabeFutureUtils#listOfFuturesToFutureOfList` instead of own impl](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-c4869415ca73705d968be9fd71086409).

2. Moved the [private utils](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-925cd7aab9dfa200affec5299f40b548L169) in `VariantReferenceResolver` to own [`CompletabeFutureUtils`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-a68bd83c3bb37d250217445d5d0117ed) and [`StreamUtils`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-2571cf584818e0b8eb82fca88f068b03).
  a. Used the utils across the project in in [`ProductReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-0a79a67e66112ba2c9dd95da412f7048), [`ProductSync`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-0d16321dedd4645c5b1ad8e28f5aa44c), [`VariantReferenceResolver`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-925cd7aab9dfa200affec5299f40b548), [`CategoryServiceImpl`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-38b71d2c035833260351074420bf8067) and [`ProductServiceImpl`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-6dfa47f1d198543a343a810f8a0c0ab5). 
  b. Add unit tests for [`CompletabeFutureUtils`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-f1b53b671e862726f8386e36c3e75cda) and [`StreamUtils`](https://github.com/commercetools/commercetools-sync-java/pull/254/files#diff-3131d2e67143198efe5c8e0ed794cdba)

3. [Add some missing annotations](https://github.com/commercetools/commercetools-sync-java/pull/254/commits/b6b591f646619dbcc1103da25a36bbaa9e1baffd). 

#### Relevant Issues
#3 

#### Todo

- Tests
    - [x] Unit tests.


src/test/java/com/commercetools/sync/commons/utils/StreamUtilsTest.java
```diff
@@ -0,0 +1,38 @@
...
+        assertThat(mappedList).isEmpty();
+    }
+
+    @Test
+    public void map_nonEmptyStream_ShouldReturnMappedStream() {
+        final Stream<Integer> mappedStream = map(Stream.of("foo", "bar"), String::length);
+        final List<Integer> mappedList = mappedStream.collect(toList());
+        assertThat(mappedList).isNotEmpty();
+        assertThat(mappedList).hasSize(2);
+        assertThat(mappedList).containsExactlyInAnyOrder(3, 3);
```
andrii-kovalenko-ct: 1. at least make input strings different length 
2. The util should not change "order" property of input stream, thus order of output is matter.
3. Same for test below

src/test/java/com/commercetools/sync/commons/utils/StreamUtilsTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Test
+    public void map_nonEmptyStreamWithNullElements_ShouldReturnMappedStream() {
+        final Stream<Integer> mappedStream = map(Stream.of("foo", "bar", null), String::length);
+        final List<Integer> mappedList = mappedStream.collect(toList());
+        assertThat(mappedList).isNotEmpty();
+        assertThat(mappedList).hasSize(2);
+        assertThat(mappedList).containsExactlyInAnyOrder(3, 3);
+    }
+}
```
andrii-kovalenko-ct: i'd also add cases with:
  - single element 
  - single **null** element 
  - only **null** elements (returns empty stream)

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
```diff
@@ -0,0 +1,27 @@
...
+     * {@code elements}.
+     *
+     * @param elements the stream of elements.
+     * @param mapper   the mapper function to apply on every element.
+     * @param <T>      the type of the elements in the stream.
+     * @param <S>      the resulting type after applying the mapper function on an element.
+     * @return a stream of the resulting mapped elements.
+     */
+    @Nonnull
+    public static <T, S> Stream<S> map(@Nonnull final Stream<T> elements, @Nonnull final Function<T, S> mapper) {
```
andrii-kovalenko-ct: i'd give the function more descriptive name, smth like `fillterNullAndMap`, because just `map` is too common name used in Streams and Optionals

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
```diff
@@ -0,0 +1,162 @@
...
+     * @param <T> the element obtained from the set of {@code CompletionStage}
+     * @return the {@code CompletableFuture} of a set of elements
+     */
+    @Nonnull
+    public static <T> CompletableFuture<Set<T>> setOfFuturesToFutureOfSet(
+        @Nonnull final Set<? extends CompletionStage<T>> set) {
+
+        final Set<CompletableFuture<T>> futureSet = set.stream()
+                                                       .map(CompletionStage::toCompletableFuture)
+                                                       .collect(toSet());
```
andrii-kovalenko-ct: minor optimization:
it doesn't have sense right here to collect to slower and more sophisticated `Set`, a `List` could be efficient enough.

And then later collect to `Set` in the return result

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    @Nonnull
+    public static <T> Set<CompletableFuture<T>> toSetOfCompletableFutures(
+        @Nonnull final Stream<CompletionStage<T>> values) {
+        return values.map(CompletionStage::toCompletableFuture).collect(toSet());
+    }
+
+    private CompletableFutureUtils() {
+    }
+}
```
andrii-kovalenko-ct: It looks like the utility could be re-factored:
  - unified methods to accept `Collections`, instead of `List` and `Set` separately, and then use `Collections#stream()`
    Maybe it should be overloaded `mapValuesToFutureOfCompletedValues(Collection<T> values, ...)`
  - `toListOfCompletableFutures` and `toSetOfCompletableFutures` could be unified using second argument of type `Collector`, and based on this collector type it will return `List` or `Set`

Then in most cases calls like `entities.entrySet().stream()` could be encapsulated passing just `entities`

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
```diff
@@ -141,18 +142,9 @@ private void cacheProductIds(@Nonnull final List<Product> products) {
...
-                          .collect(Collectors.toList());
-        return CompletableFuture.allOf(futureCreations.toArray(new CompletableFuture[futureCreations.size()]))
-                                .thenApply(result -> futureCreations.stream()
-                                                                    .map(CompletionStage::toCompletableFuture)
-                                                                    .map(CompletableFuture::join)
-                                                                    .filter(Optional::isPresent)
-                                                                    .map(Optional::get)
-                                                                    .collect(Collectors.toSet()));
+        return mapValuesSetToFutureOfCompletedValues(productsDrafts, this::createProduct)
+            .thenApply(results -> results.filter(Optional::isPresent).map(Optional::get))
```
andrii-kovalenko-ct: i found that approach `filter(Optional::isPresent).map(Optional::get)` is used very often...
Well, i voted already once to dismiss mostly Optionals in favor of Streams and flatMap, but even if we decided to stick with Optionals+Stream i'd recommend to carry out to some utils:

  - `filter(Optional::isPresent).map(Optional::get)`
  - `filter(Optional::isPresent).map(Optional::get).collect(listOrSetCollector)`

src/main/java/com/commercetools/sync/commons/helpers/AssetReferenceResolver.java
```diff
@@ -0,0 +1,57 @@
...
+        if (custom != null) {
+            return getCustomTypeId(custom,
+                format(FAILED_TO_RESOLVE_CUSTOM_TYPE, assetDraftBuilder.getKey()))
+                .thenApply(resolvedTypeIdOptional -> resolvedTypeIdOptional
+                    .map(resolvedTypeId -> assetDraftBuilder
+                        .custom(CustomFieldsDraft.ofTypeIdAndJson(resolvedTypeId, custom.getFields())))
+                    .orElse(assetDraftBuilder));
+        }
+        return completedFuture(assetDraftBuilder);
+    }
```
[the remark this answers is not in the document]
heshamMassoud: 09a9030

src/main/java/com/commercetools/sync/products/helpers/VariantReferenceResolver.java
```diff
@@ -73,32 +77,48 @@ public VariantReferenceResolver(@Nonnull final ProductSyncOptions productSyncOpt
...
+    CompletionStage<ProductVariantDraftBuilder> resolveAssetsReferences(
+        @Nonnull final ProductVariantDraftBuilder productVariantDraftBuilder) {
+        final List<AssetDraft> productVariantDraftAssets = productVariantDraftBuilder.getAssets();
+        if (productVariantDraftAssets == null) {
+            return completedFuture(productVariantDraftBuilder);
+        }
+
+        return mapValuesListToFutureOfCompletedValues(productVariantDraftAssets,
+            assetReferenceResolver::resolveReferences).thenApply(productVariantDraftBuilder::assets);
+    }
```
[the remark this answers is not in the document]
heshamMassoud: It could be achieved by having an interface that both resolvers could implement:

````java
public interface AssetContainerResolver<T>{

    @Nonnull
    CompletionStage<T> resolveAssetsReferences(@Nonnull final T draftBuilder);

    @Nonnull
    default CompletionStage<T> resolveAssetsReferences(
        @Nonnull final T draftBuilder,
        @Nonnull final Function<T, List<AssetDraft>> assetsGetter,
        @Nonnull final BiFunction<T, List<AssetDraft>, T> assetsSetter,
        @Nonnull final AssetReferenceResolver assetReferenceResolver) {

        final List<AssetDraft> draftAssets = assetsGetter.apply(draftBuilder);
        if (draftAssets == null) {
            return completedFuture(draftBuilder);
        }

        return mapValuesListToFutureOfCompletedValues(draftAssets, assetReferenceResolver::resolveReferences)
            .thenApply(resolvedAssets -> assetsSetter.apply(draftBuilder, resolvedAssets));
    }
}
````

but then the implementing method of resolvers would have to have `resolveAssetReferences` as public which is something I don't want (at least for now) since resolvers by convention now have only `resolveReferences` method exposed and no other methods. 

Since in this case, there is not much improvement by reuse. I would wait if in the future for example other resources have assets then we could start thinking about exposing more methods of the resolvers and then I could reuse a common asset resolution interface.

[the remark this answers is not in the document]
Comment (andrii-kovalenko-ct) [reply src/main/java/com/commercetools/sync/products/helpers/VariantReferenceResolver.java]: Why only by the interface? Both classes extend `BaseReferenceResolver` - the code could be implemented there.

File: src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java (hunk 5 above)
Comment (heshamMassoud) [reply src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java]: Utils are refactored: 
6d404c9

File: src/test/java/com/commercetools/sync/commons/utils/StreamUtilsTest.java (hunk 1 above)
Comment (heshamMassoud) [reply src/test/java/com/commercetools/sync/commons/utils/StreamUtilsTest.java]: d9695b1

File: src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java (hunk 3 above)
Comment (heshamMassoud) [reply src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java]: 3038fd2

File: src/test/java/com/commercetools/sync/commons/utils/StreamUtilsTest.java (hunk 2 above)
Comment (heshamMassoud) [reply src/test/java/com/commercetools/sync/commons/utils/StreamUtilsTest.java]: d9695b1

File: src/main/java/com/commercetools/sync/products/helpers/VariantReferenceResolver.java (hunk 8 above)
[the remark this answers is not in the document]
Comment (heshamMassoud) [reply src/main/java/com/commercetools/sync/products/helpers/VariantReferenceResolver.java]: Not so easy, because in the BaseReferenceResolver the generic type is a *`Draft` not a *`DraftBuilder`. So might need a bit more refactoring. So I wouldn't do it in the same PR.

File: src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java (hunk 6 above)
Comment (heshamMassoud) [reply src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java]: I definitley agree, but also too much for this PR and time constraint of asset sync feature at the moment :( 

So issue created in the backlog: https://github.com/commercetools/commercetools-sync-java/issues/255

File: src/main/java/com/commercetools/sync/products/helpers/PriceReferenceResolver.java [hunk 9]
```diff
@@ -56,18 +55,14 @@ public PriceReferenceResolver(@Nonnull final ProductSyncOptions options,
...
-                    .orElse(draftBuilder));
-        }
-        return completedFuture(draftBuilder);
+        @Nonnull final PriceDraftBuilder draftBuilder) {
+
+        return resolveCustomTypeReference(draftBuilder,
+            PriceDraftBuilder::getCustom,
+            PriceDraftBuilder::custom,
+            format(FAILED_TO_RESOLVE_CUSTOM_TYPE, draftBuilder.getCountry(), draftBuilder.getValue()));
     }
```
ahalberkamp: isn`t is enough to only pass the draftbuilder to that function, and extract the functions "getCuston" and "custom" from the given draftbuilder (anyhow)?

src/test/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolverTest.java
```diff
@@ -51,19 +58,70 @@ public void setup() {
...
+        final CategoryDraftBuilder categoryDraftBuilder =
+            getMockCategoryDraftBuilder(Locale.ENGLISH, "myDraft", "key", CACHED_CATEGORY_KEY,
+                "customTypeId", new HashMap<>())
+                .assets(null);
+
+        final CategoryDraftBuilder resolvedBuilder = referenceResolver.resolveAssetsReferences(categoryDraftBuilder)
+                                                                      .toCompletableFuture().join();
+
+        final List<AssetDraft> resolvedBuilderAssets = resolvedBuilder.getAssets();
+        assertThat(resolvedBuilderAssets).isNull();
```
ahalberkamp: Maybe the function "getAsset" should return a empty collect instead of null

src/test/java/com/commercetools/sync/commons/helpers/AssetReferenceResolverTest.java
```diff
@@ -0,0 +1,204 @@
...
+        final AssetReferenceResolver assetReferenceResolver =
+            new AssetReferenceResolver(productSyncOptions, typeService);
+
+
+        final AssetDraftBuilder resolvedDraftBuilder =
+            assetReferenceResolver.resolveCustomTypeReference(assetDraftBuilder)
+                                  .toCompletableFuture().join();
+
+        assertThat(resolvedDraftBuilder.getCustom()).isNotNull();
+        assertThat(resolvedDraftBuilder.getCustom().getType().getId()).isEqualTo("typeId");
```
ahalberkamp: Can you also check the value of custom here

src/test/java/com/commercetools/sync/commons/utils/completablefutureutils/MapStreamValuesToFutureOfCompletedValuesTest.java
```diff
@@ -0,0 +1,211 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class MapStreamValuesToFutureOfCompletedValuesTest {
+    /**
+     * Stream to List : empty values.
+     * Stream to Set : empty values.
+     **/
+
+    @Test
+    public void empty_StreamToList_ReturnsFutureOfEmptyList() {
```
ahalberkamp: I thought that the test method name should always beginn with the name of the method,which is tested

src/test/java/com/commercetools/sync/commons/utils/completablefutureutils/MapValuesToFutureOfCompletedValuesStreamTest.java
```diff
@@ -0,0 +1,219 @@
...
+        assertThat(result).isInstanceOf(Stream.class);
+    }
+
+    @Test
+    public void multiple_ListWithDiffTypeMappingNoDuplicates_ReturnsFutureOfStreamOfMappedValues() {
+        final CompletableFuture<Stream<Integer>> future = mapValuesToFutureOfCompletedValues(
+            asList("john", "smith"), element -> completedFuture(element.length()));
+        final Stream<Integer> result = future.join();
+        assertThat(result).containsExactly(4, 5);
+        assertThat(result).isInstanceOf(Stream.class);
```
ahalberkamp: amazing test :-)

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
andrii-kovalenko-ct: since you use `mapValuesToFutureOfCompletedValues()` mostly with `toList`, does it have sense to create a dedicated function <code>mapValuesToFutureOf<b>List</b>CompletedValues</code>

andrii-kovalenko-ct: smth like:
```java
@Nonnull
public static <T, S> CompletableFuture<List<S>> mapValuesToFutureOfListCompletedValues(
    @Nonnull final Collection<T> values,
    @Nonnull final Function<T, CompletionStage<S>> mapper) {
    return mapValuesToFutureOfCompletedValues(values.stream(), mapper, toList());
}

@Nonnull
public static <T, S> CompletableFuture<Set<S>> mapValuesToFutureOfSetCompletedValues(
    @Nonnull final Collection<T> values,
    @Nonnull final Function<T, CompletionStage<S>> mapper) {
    return mapValuesToFutureOfCompletedValues(values.stream(), mapper, toSet());
}
```

src/main/java/com/commercetools/sync/products/helpers/VariantReferenceResolver.java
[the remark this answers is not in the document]
andrii-kovalenko-ct: ok

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
heshamMassoud: @andrii-kovalenko-ct would be good, but there I don't have the time luxury to do that now (since I will add more tests and so on for it). There are more important things to do now. So maybe later: https://github.com/commercetools/commercetools-sync-java/issues/258

src/main/java/com/commercetools/sync/commons/helpers/AssetReferenceResolver.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    @Nonnull
+    protected CompletionStage<AssetDraftBuilder> resolveCustomTypeReference(
+        @Nonnull final AssetDraftBuilder assetDraftBuilder) {
+
+        return resolveCustomTypeReference(
+            assetDraftBuilder,
+            AssetDraftBuilder::getCustom,
+            AssetDraftBuilder::custom,
```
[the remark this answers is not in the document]
heshamMassoud: This is not my method, it's a method on the AssetDraftBuilder on the JVM SDK

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
```diff
@@ -48,11 +52,47 @@ protected CustomReferenceResolver(@Nonnull final S options, @Nonnull final TypeS
      *
      * @param draftBuilder the draft builder to resolve it's references.
      * @return a {@link CompletionStage} that contains as a result a new draft instance with resolved custom
-     *      type references or, in case an error occurs during reference resolution,
-     *      a {@link ReferenceResolutionException}.
+     *         type references or, in case an error occurs during reference resolution,
+     *         a {@link ReferenceResolutionException}.
```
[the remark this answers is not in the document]
heshamMassoud: It's the same grammitcally :)

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
```diff
@@ -48,11 +52,47 @@ protected CustomReferenceResolver(@Nonnull final S options, @Nonnull final TypeS
...
+     */
+    @Nonnull
+    protected CompletionStage<B> resolveCustomTypeReference(
+        @Nonnull final B draftBuilder,
+        @Nonnull final Function<B, CustomFieldsDraft> customGetter,
+        @Nonnull final BiFunction<B, CustomFieldsDraft, B> customSetter,
+        @Nonnull final String errorMessage) {
+
+        final CustomFieldsDraft custom = customGetter.apply(draftBuilder);
```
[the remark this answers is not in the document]
heshamMassoud: actually it's "custom" on purpose because that's how the field is named from on our API :) : 

examples: 
https://docs.commercetools.com/http-api-types.html#assetdraft
https://docs.commercetools.com/http-api-projects-channels.html#channeldraft

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
```diff
@@ -0,0 +1,163 @@
...
+     * @param values collection of values to apply a mapper function that would map each to a {@link CompletionStage}.
+     * @param mapper function to map each value to a {@link CompletionStage}
+     * @param <T>    The type of the values.
+     * @param <S>    The type of the mapped completed values.
+     * @return a future containing a stream of completed stage results of the values after the mapper function was
+     *         applied to each value in the supplied collection.
+     */
+    @Nonnull
+    public static <T, S> CompletableFuture<Stream<S>> mapValuesToFutureOfCompletedValues(
+        @Nonnull final Collection<T> values,
```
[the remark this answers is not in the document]
heshamMassoud: only to avoid always to have collection.stream() on every call from the user.

src/main/java/com/commercetools/sync/products/helpers/PriceReferenceResolver.java
heshamMassoud: no because all the builders do not implement an interface which has these `getCusom` and `custom` methods in the JVM SDK 😞 . Every builder has it's own inividual method (just with same names). That's why I have to pass the method references.

src/test/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolverTest.java
heshamMassoud: if the assets are not given to the JVM SDK AssetDraftBuilder. The assets would be `null`. Thanks for reminding me tho to add a unit test for empty case: eb4e50a

src/test/java/com/commercetools/sync/commons/helpers/AssetReferenceResolverTest.java
heshamMassoud: this is not the goal of this unit test. The goals of this "unit" test is to just check if the reference (custom type reference in this case) has been resolved.

src/test/java/com/commercetools/sync/commons/utils/completablefutureutils/MapStreamValuesToFutureOfCompletedValuesTest.java
heshamMassoud: True, but in this case, unit testing the class `CompleteableFutureUtils` in one test class would make it unreadable due to the vast permutations of inputs/outputs to the methods. So every method now has it's own unit test class that has inside all the different input/output combinations. That's why I followed the following approach:

class under test -> package name in the unit tests `/completablefutueutils/`
method under test -> unit test class name (in this case `MapStreamValuesToFutureOfCompletedValuesTest`)
method convention followed -> **`StateUnderTest1_StateUnderTest2_ExpectedBehavior`**

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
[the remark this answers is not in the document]
ahalberkamp: yes but I prefer my sentence structur

src/test/java/com/commercetools/sync/commons/utils/completablefutureutils/MapStreamValuesToFutureOfCompletedValuesTest.java
ahalberkamp: thanks for explenation

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
andrii-kovalenko-ct: ok

src/test/java/com/commercetools/sync/commons/utils/completablefutureutils/MapValuesToFutureOfCompletedValuesStreamTest.java
andrii-kovalenko-ct: ???
why do you assert instance of `Stream` variable is instance of `Stream`?

heshamMassoud: oh, copy-paste mistake. thanks: 630741b

Issue #164, opened 2017-10-25
Make util out of CompletableFuture.completedFuture(Optional.empty()) 


heshamMassoud:
### Description
`CompletableFuture.completedFuture(Optional.empty())`
we have 5 usage in the main sources + 11 usages in the tests.

Maybe it is time to make a static util method for this
Could it be a constant? (e.g. could you reused already completed/joined future multiple times? guess yes, but not sure though)

Same for `completedFuture(null)`

  - [ ] add completed `Optional` future
  - [ ] add completed `null` future

### Related
https://github.com/commercetools/commercetools-sync-java/pull/154#discussion_r146798295

andrii-kovalenko-ct:
```java
private static final CompletableFuture<Optional<?>> EMPTY = CompletableFuture.completedFuture(Optional.empty());

@Nonnull
@SuppressWarnings("unchecked")
public static <T> CompletableFuture<T> emptyOptionalCompletedFuture() {
    return (CompletableFuture<T>)EMPTY;
}
````

whitemalt:
Hi let review on #328  

integration test appear fail since before commit I did

heshamMassoud:
I reoppened it again because I have failed to notice this line: https://github.com/commercetools/commercetools-sync-java/pull/328/files#diff-a68bd83c3bb37d250217445d5d0117edR162

ahmetoz:
It looks like it resolved in a PR but we have not considered closing this issue, I wonder who will close and when 🤔 
I also think using util for this is not necessary for now, it's already util getting rid of Optional.empty does not have much value.

PR #256, opened 2018-02-22
Introduce a util method for CompletableFuture of an empty Optional

LEQADA:
#### Summary
Create a static util method for `CompletableFuture.completedFuture(Optional.empty())` usages

#### Relevant Issues
Fixes https://github.com/commercetools/commercetools-sync-java/issues/164

#### Todo
- Tests
    - [ ] Unit 
    - [ ] Integration
- [ ] Documentation

### DIFF
--- a/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
@@ -4,7 +4,7 @@ import javax.annotation.Nonnull;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
-public class CompletableFutureUtils {
+public final class CompletableFutureUtils {
 
     private static final CompletableFuture<Optional<?>> EMPTY = CompletableFuture.completedFuture(Optional.empty());
 
--- a/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
@@ -8,6 +8,8 @@ public final class CompletableFutureUtils {
 
     private static final CompletableFuture<Optional<?>> EMPTY = CompletableFuture.completedFuture(Optional.empty());
 
+    private CompletableFutureUtils() { }
+
     /**
      * Returns a {@link CompletableFuture} of an empty {@link Optional}. There is no guarantee
      * that it is a singleton.
--- a/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
@@ -1,0 +1,21 @@
+package com.commercetools.sync.commons.utils;
+
+import javax.annotation.Nonnull;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class CompletableFutureUtils {
+
+    private static final CompletableFuture<Optional<?>> EMPTY = CompletableFuture.completedFuture(Optional.empty());
+
+    /**
+     * Returns a {@link CompletableFuture} of an empty {@link Optional}. There is no guarantee
+     * that it is a singleton.
+     * @return {@link CompletableFuture} of an empty {@link Optional}
+     * */
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    public static <T> CompletableFuture<T> emptyOptionalCompletedFuture() {
+        return (CompletableFuture<T>)EMPTY;
+    }
+}
--- a/src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java
+++ b/src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java
@@ -162,7 +162,7 @@ public class InventorySyncTest {
             mock(InventoryEntry.class), mock(InventoryEntry.class));
         final ChannelService channelService = mock(ChannelService.class);
         when(channelService.fetchCachedChannelId(anyString()))
-            .thenReturn(completedFuture(Optional.empty()));
+            .thenReturn(emptyOptionalCompletedFuture());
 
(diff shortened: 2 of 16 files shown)


src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class CompletableFutureUtils {
+
+    private static final CompletableFuture<Optional<?>> EMPTY = CompletableFuture.completedFuture(Optional.empty());
+
+    /**
+     * Returns a {@link CompletableFuture} of an empty {@link Optional}. There is no guarantee
+     * that it is a singleton.
```
andrii-kovalenko-ct: > here is no guarantee that it is a singleton.

what does this mean?

src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java
```diff
@@ -469,7 +470,7 @@ public void syncDrafts_WithExceptionWhenCreatingNewSupplyChannel_ShouldTriggerEr
 
         final ChannelService channelService = getMockChannelService(getMockSupplyChannel(REF_3, KEY_3));
         when(channelService.fetchCachedChannelId(anyString()))
-            .thenReturn(CompletableFuture.completedFuture(Optional.empty()));
+            .thenReturn(emptyOptionalCompletedFuture());
```
andrii-kovalenko-ct: pls, also update:
  - https://github.com/commercetools/commercetools-sync-java/blob/90b209707541a88d577c5dae3fee17432b8fc9a6/src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java#L165

  - https://github.com/commercetools/commercetools-sync-java/blob/90b209707541a88d577c5dae3fee17432b8fc9a6/src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java#L195

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
LEQADA: From the first glance it looks like a singleton. But actually it is not. 
So I just stole a comment from the [Optional#empty()](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#empty--) JavaDoc

src/test/java/com/commercetools/sync/inventories/InventorySyncTest.java
LEQADA: https://github.com/commercetools/commercetools-sync-java/pull/256/commits/61f3cbed2ad0f416245f12fcaff2adef6f3edd51

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
andrii-kovalenko-ct: don't think we have to mention it here, but even if we decided to make such a comment - pls, format them as in that example, because it is less confusing.

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
```diff
@@ -0,0 +1,21 @@
...
+     * @return {@link CompletableFuture} of an empty {@link Optional}
+     * */
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    public static <T> CompletableFuture<T> emptyOptionalCompletedFuture() {
```
andrii-kovalenko-ct: Actually, it must return `Optional<T>`, not just `<T>`

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
```diff
@@ -0,0 +1,21 @@
+package com.commercetools.sync.commons.utils;
+
+import javax.annotation.Nonnull;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class CompletableFutureUtils {
```
andrii-kovalenko-ct: pls, add private constructor to Util classes (by convention)

andrii-kovalenko-ct:
@LEQADA @heshamMassoud 
i found we also use `completedFuture(null)` a lot -> updated updated the issue 
https://github.com/commercetools/commercetools-sync-java/issues/164

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
LEQADA: https://github.com/commercetools/commercetools-sync-java/pull/256/commits/853108b5eb23d68ca876a21358abdbc614b1f8f5

src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
andrii-kovalenko-ct: and better return every time new instance of `CompletableFuture`, since the class is mutable:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#obtrudeValue-T-

andrii-kovalenko-ct: and add a separate test case verifying mutability, smth like:
```java
@Test
public void emptyOptionalCompletedFuture_immutable() throws ExecutionException, InterruptedException {
    CompletableFuture<Optional<Integer>> completedFuture = emptyOptionalCompletedFuture();
    completedFuture.obtrudeValue(Optional.of(42));
    assertThat(completedFuture.get()).contains(42);

    CompletableFuture<Optional<Integer>> completedFuture2 = emptyOptionalCompletedFuture();
    assertThat(completedFuture2).isNotSameAs(completedFuture);
    assertThat(completedFuture2.get()).isEmpty();
}
```

Issue #238, opened 2018-01-19
Update/Create products in parallel


lojzatran:
### Description
Create and update of product is now done sequentially. Maybe it's possible to do it in parallel.

### Current Behaviour
```java
@Nonnull
    private CompletionStage<Void> createOrUpdateProducts() {
        return productService.createProducts(draftsToCreate)
                             .thenAccept(createdProducts ->
                                 processCreatedProducts(createdProducts, draftsToCreate.size()))
                             .thenCompose(result -> syncProducts(productsToSync));
    }
```

### Expected Behaviour
`productService.createProducts(draftsToCreate)` and `syncProducts(productsToSync)` can be done with `CompletableFuture.allOf()`

heshamMassoud:
released in v1.0.0-M14

PR #269, opened 2018-03-08
#238: Create and update products in parallel in product sync

heshamMassoud:
#### Description
Create and update of product is used to be done sequentially as follows:

```java
@Nonnull
    private CompletionStage<Void> createOrUpdateProducts() {
        return productService.createProducts(draftsToCreate)
                             .thenAccept(createdProducts ->
                                 processCreatedProducts(createdProducts, draftsToCreate.size()))
                             .thenCompose(result -> syncProducts(productsToSync));
    }
```

This PR addresses the issue and does them [in parallel using `CompletableFuture.allOf()`](https://github.com/commercetools/commercetools-sync-java/pull/269/commits/993863b6df19045f68e7098ba322f61453a91d1a)

#### Boy-Scout changes

Bump `test-logger`, `mockito` and `Junit5` versions.

#### Relevant Issues
#238

#### Todo

- Tests
    - [x] ~~Unit~~ - not needed
    - [x] ~~Integration~~  - not needed
- [x] ~~Documentation~~  - not needed


src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -181,11 +181,16 @@ private void processFetchedProducts(@Nonnull final Set<Product> fetchedProducts)
     }
 
     @Nonnull
-    private CompletionStage<List<Optional<Product>>> createOrUpdateProducts() {
-        return productService.createProducts(draftsToCreate)
-                             .thenAccept(createdProducts ->
-                                 updateStatistics(createdProducts, draftsToCreate.size()))
-                             .thenCompose(ignoredResult -> syncProducts(productsToSync));
+    private CompletionStage<Void> createOrUpdateProducts() {
```
andrii-kovalenko-ct: I'd say it is <code>create<b>And</b>UpdateProducts</code>, since there is no dependency between these actions.
Actually, because of this name i was confused during reviewing: first i thought the stages should be chained.

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -181,11 +181,16 @@ private void processFetchedProducts(@Nonnull final Set<Product> fetchedProducts)
...
+        final CompletableFuture<Void> futureCreates =
+            productService.createProducts(draftsToCreate)
+                          .thenAccept(createdProducts -> updateStatistics(createdProducts, draftsToCreate.size()))
+                          .toCompletableFuture();
+
+        final CompletableFuture<List<Optional<Product>>> futureUpdates =
+            syncProducts(productsToSync).toCompletableFuture();
```
andrii-kovalenko-ct: `ProductService#createProducts` and `ProductSync#syncProducts` could return more specific `CompletableFuture<T>` instead of `CompletionStage<T>`, then `toCompletableFuture()` becomes redundant!

Actually, i think we did design mistake from the beginning: it is generally recommended to return as narrow (low) class as possible, and for most of our functions we have to return `CompletableFuture`!

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -181,11 +181,16 @@ private void processFetchedProducts(@Nonnull final Set<Product> fetchedProducts)
...
+
+        return CompletableFuture.allOf(futureCreates, futureUpdates);
```
andrii-kovalenko-ct: For the whole approach in this function i'd say `CompletableFuture#thenCombine()` looks simpler and more sufficient, than `CompletableFuture#allOf()`:
```java
return productService.createProducts(draftsToCreate)
    .thenAccept(createdProducts -> updateStatistics(createdProducts, draftsToCreate.size()))
    .thenCombine(syncProducts(productsToSync), (void1, void2) -> null);
```

src/main/java/com/commercetools/sync/products/ProductSync.java
heshamMassoud: Actually, the reason why it's `OR` and not `AND` in the name, is because it could be that there are no creates and updates only, and it also could be that there are creates only and no updates. Hmm, but I see your point of confusion. Hmm, I will change to `AND` since it's at least less confusing.

heshamMassoud: d963190

src/main/java/com/commercetools/sync/products/ProductSync.java
heshamMassoud: 2e15fe9

src/main/java/com/commercetools/sync/products/ProductSync.java
heshamMassoud: Indeed, I agree :) for now at least: a60dfcf. For the rest issue will be created ;)

heshamMassoud: https://github.com/commercetools/commercetools-sync-java/issues/271

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -186,11 +187,16 @@ private void processFetchedProducts(@Nonnull final Set<Product> fetchedProducts)
...
+    private CompletableFuture<Void> createAndUpdateProducts() {
+        final CompletableFuture<Void> createRequestsStage =
+            productService.createProducts(draftsToCreate)
+                          .thenAccept(createdProducts -> updateStatistics(createdProducts, draftsToCreate.size()))
+                          .toCompletableFuture();
+
+        final CompletableFuture<List<Optional<Product>>> updateRequestsStage =
+            syncProducts(productsToSync).toCompletableFuture();
+
+        return allOf(createRequestsStage, updateRequestsStage);
```
butenkor: Did you notice any performance difference between previous and current approach as i doubt there will be any?

heshamMassoud: Actually I haven't noticed yet. Maybe I will notice on the benchmarks. But why do you doubt there will be any difference? Because of the number of parallel requests set on the client?

PR #273, opened 2018-03-16
255 create util for option resolution

ahalberkamp:
#### Summary
Time to make a util out of filter(Optional::isPresent).map(Optional::get)

For more context check the comment: #254 (comment)


ahalberkamp:
deprecated

Issue #255, opened 2018-02-22
Make Util out of filter(Optional::isPresent).map(Optional::get)


heshamMassoud:
### Description

Time to make a util out of `filter(Optional::isPresent).map(Optional::get)`

For more context check the comment:
https://github.com/commercetools/commercetools-sync-java/pull/254#discussion_r169344842

heshamMassoud:
Released in [1.0.0](https://github.com/commercetools/commercetools-sync-java/releases/tag/1.0.0)

PR #275, opened 2018-03-22
#255: stream filtering over flatMap

andrii-kovalenko-ct:
@heshamMassoud @ahalberkamp 

My proposition for #255 is a bit different from Andi's solution. IMHO it is more "functional" and more in common design of Java Streams API.

My idea:
  - we don't wrap collections/list/streams to util methods (e.g. like Andi's `asList()`/`asSet()`)
  - instead we call `Stream#flatMap(Function)` using our [`StreamUtils::filterEmptyOptionals`](https://github.com/commercetools/commercetools-sync-java/pull/275/files#diff-2571cf584818e0b8eb82fca88f068b03R40) function reference which filters empty optionals. Note, similar approach is introduced in Java 9 [Optional#stream()](https://docs.oracle.com/javase/9/docs/api/java/util/Optional.html#stream--) method
  - this method is a bit more verbose since requires from users to explicitly terminate the stream with a collector, but this is more in genera Java Stream API design.
  - for better readability static import of `getFilterEmptyOptionals()` could be used instead of `StreamUtils::filterEmptyOptionals` reference. Compare [this](https://github.com/commercetools/commercetools-sync-java/pull/275/files#diff-3131d2e67143198efe5c8e0ed794cdbaR96) and [this](https://github.com/commercetools/commercetools-sync-java/pull/275/files#diff-3131d2e67143198efe5c8e0ed794cdbaR104) with other util calls in the same test class which use `StreamUtils::filterEmptyOptionals`

So what do you think: which approach should remain?


src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
```diff
@@ -32,38 +27,29 @@
...
-     * @param elements the stream of elements.
-     * @return a {@link List} of resolved optionals.
+     * @param optional Optional to convert
+     * @param <T>      type of optional value
+     * @return Stream of one element from the supplied {@code optional} if not empty, otherwise - an empty stream.
      */
     @Nonnull
-    public static <T extends List<?>> T asList(@Nonnull final Stream<? extends Optional> elements) {
-        return filterStream(elements, Collectors.toList());
+    public static <T> Stream<T> filterEmptyOptionals(final @Nonnull Optional<T> optional) {
```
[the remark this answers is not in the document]
heshamMassoud: I would also put it in a new utils `OptionalUtils` not `StreamUtils` as its operating it's a util for `Optional` types

src/main/java/com/commercetools/sync/categories/utils/CategoryAssetUpdateActionUtils.java
```diff
@@ -65,7 +65,9 @@
     @Nonnull
     private static List<UpdateAction<Category>> buildUpdateActionsFromOptionals(
         @Nonnull final List<Optional<UpdateAction<Category>>> optionalUpdateActions) {
-        return StreamUtils.asList(optionalUpdateActions.stream());
+        return optionalUpdateActions.stream()
```
heshamMassoud: I would remove this method at all and all it's similar ones and have a new util that can be re-used, for example:
in `OptionalUtils`.:

````java
    public static <T> Stream<T> toStream(@Nonnull final Optional<T>... optionals) {
        return Stream.of(optionals).flatMap(OptionalUtils::toStream);
    }
````
(also refer to this related comment: https://github.com/commercetools/commercetools-sync-java/pull/275/files#diff-2571cf584818e0b8eb82fca88f068b03R39).

Ideally also, if toList is used more often we could have one `toList`:
````java
public static <T> List<T> toList(@Nonnull final Optional<T>... optionals) {
        return Stream.of(optionals).flatMap(OptionalUtils::toStream).collect(Collectors.toList());
    }
````

File: src/main/java/com/commercetools/sync/categories/utils/CategorySyncUtils.java [hunk 3]
```diff
@@ -89,7 +89,9 @@
     @Nonnull
     private static List<UpdateAction<Category>> buildUpdateActionsFromOptionals(
         @Nonnull final List<Optional<UpdateAction<Category>>> optionalUpdateActions) {
-        return asList(optionalUpdateActions.stream());
+        return optionalUpdateActions.stream()
```
heshamMassoud: Refer to this comment: https://github.com/commercetools/commercetools-sync-java/pull/275/files#diff-93cb8feb4d531930ae6babf9eb6b4c4fR68

src/main/java/com/commercetools/sync/commons/utils/AssetsUpdateActionUtils.java
```diff
@@ -220,14 +219,15 @@
...
-
-        return asList(IntStream.range(0, newAssetDrafts.size())
+        return IntStream.range(0, newAssetDrafts.size())
             .mapToObj(assetDraftIndex ->
                 ofNullable(newAssetDrafts.get(assetDraftIndex))
                     .filter(assetDraft -> !oldAssetsKeyMap.containsKey(assetDraft.getKey()))
                     .map(assetDraft -> assetActionFactory.buildAddAssetAction(assetDraft, assetDraftIndex))
             )
-        );
+            .flatMap(StreamUtils::filterEmptyOptionals)
```
heshamMassoud: Please fix the indentation so all the chains' . have the same indentation

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
```diff
@@ -32,38 +27,29 @@
...
-     * @return a {@link Collection<T>} of resolved optionals.
+     * @param <T> type of optional value
+     * @return function which returns a stream of one element from the supplied {@code optional} if not empty,
+     *     otherwise - an empty stream.
      */
     @Nonnull
-    private static <T extends Collection> T filterStream(@Nonnull final Stream<? extends Optional> elements, @Nonnull
-        Collector<Object, ?, ?> collector) {
-        return (T) elements.filter(Optional::isPresent).map(Optional::get).collect(collector);
+    public static <T> Function<Optional<? extends T>, Stream<? extends T>> getFilterEmptyOptionals() {
```
heshamMassoud: I think it's ok to explicitly write the type of the generic with method reference. So I vote not to keep this util

heshamMassoud:
I like this design more for addressing the issue as it adheres to the same design of the stream API of JDK. SO I would vote for this one. @andrii-kovalenko-ct can you please base this PR off of master rather than Andi's PR to make it easier for review?

src/main/java/com/commercetools/sync/categories/utils/CategoryAssetUpdateActionUtils.java
andrii-kovalenko-ct: i didn't get where (besides tests) we use Optional varargs to collect them to list?

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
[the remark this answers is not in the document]
andrii-kovalenko-ct: > I would name the method toStream instead of filterEmptyOptionals as it only takes one optional and simply converts to a stream if it is present.

I tried: it doesn't seem so, for example:
``` 
 .flatMap(Utils::filterEmptyOptionals) // here i see explicitly from the name what's going on 
or
 .flatMap(Utils::toStream) // just some stream flatted to another stream
```

src/main/java/com/commercetools/sync/categories/utils/CategoryAssetUpdateActionUtils.java
andrii-kovalenko-ct: Maybe you mean smth like:
```java
public static <T> List<T> toList(@Nonnull final List<T> list) {
    return list.stream().flatMap(OptionalUtils::toStream).collect(Collectors.toList());
}
```

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
andrii-kovalenko-ct: Not sure what do you mean, what's wrong with generics in this method?
the only one thing i noticed - redundant `? extends`: b56194e3

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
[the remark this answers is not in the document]
andrii-kovalenko-ct: > I would also put it in a new utils OptionalUtils not StreamUtils as its operating it's a util for Optional types

60d0a9b4

andrii-kovalenko-ct:
>can you please base this PR off of master rather than Andi's PR to make it easier for review?
not sure it will be easier because i applied my changes on top of Andi's branch, so re-basing it on master could be more overwhelming...

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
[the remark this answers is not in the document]
heshamMassoud: @andrii-kovalenko-ct 
> // just some stream flatted to another stream

That's what exactly `OptionalUtils` would show explicitly when using `OptionalUtils::toStream`, you see exactly that an optional is being converted to a stream.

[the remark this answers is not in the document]
andrii-kovalenko-ct: but `OptionalUtils::toStream` doesn't looks verbose enough.
At least for me, despite that fact i implemented the method and know what it does. Guess other developers would be even more confused:
  - maybe this method converts stream to optional? 
  - or merges multiple optionals to common stream?
  - or converts nullable instances to stream of optionals?

src/main/java/com/commercetools/sync/categories/utils/CategoryAssetUpdateActionUtils.java
heshamMassoud: > i didn't get where (besides tests) we use Optional varargs to collect them to list?

check the usage/caller of the method `buildUpdateActionsFromOptionals`:

currently it is:
````java
final List<UpdateAction<Category>> updateActions = buildUpdateActionsFromOptionals(asList(
            buildChangeAssetNameUpdateAction(oldAsset, newAsset),
            buildSetAssetDescriptionUpdateAction(oldAsset, newAsset),
            buildSetAssetTagsUpdateAction(oldAsset, newAsset),
            buildSetAssetSourcesUpdateAction(oldAsset, newAsset)
        ));
````

instead it would be:
````java
OptionalUtils.toStream(buildChangeAssetNameUpdateAction(oldAsset, newAsset),
            buildSetAssetDescriptionUpdateAction(oldAsset, newAsset),
            buildSetAssetTagsUpdateAction(oldAsset, newAsset),
            buildSetAssetSourcesUpdateAction(oldAsset, newAsset))
                     .collect(toList());
````

heshamMassoud: and with the `toList` sugar it could even be:

````java
final List<UpdateAction<Category>> updateActions = 
OptionalUtils.toList(buildChangeAssetNameUpdateAction(oldAsset, newAsset),
            buildSetAssetDescriptionUpdateAction(oldAsset, newAsset),
            buildSetAssetTagsUpdateAction(oldAsset, newAsset),
            buildSetAssetSourcesUpdateAction(oldAsset, newAsset));
````

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
heshamMassoud: No, I didn't mean the generics in this method signature. I meant the need for this method at all. As you mentioned already:

> for better readability static import of `getFilterEmptyOptionals()` could be used instead of `StreamUtils::filterEmptyOptionals` reference. Compare this and this with other util calls in the same test class which use `StreamUtils::filterEmptyOptionals`

I prefered to have the consistent style of using method references with stream api and lambda's. i.e. using `StreamUtils::filterEmptyOptionals` instead of `getFilterEmptyOptionals()` even if we have to explicitly mention the generic types of the stream elements, like here: [`Stream.<Optional<String>>of`](https://github.com/commercetools/commercetools-sync-java/pull/275/files#diff-75070d8e3eb2924f08d9b46edc1ab484R92) I think it's totally ok, readability-wise.

src/main/java/com/commercetools/sync/categories/utils/CategoryAssetUpdateActionUtils.java
andrii-kovalenko-ct: but the it should be smth like:
`toMutableListFilteringEmpty`
:D

andrii-kovalenko-ct: also, about mutability and common collectors:
it thought it will be a separate PR. do you want to put it to this PR?

src/main/java/com/commercetools/sync/commons/utils/StreamUtils.java
andrii-kovalenko-ct: e93ecdc0

src/main/java/com/commercetools/sync/categories/utils/CategoryAssetUpdateActionUtils.java
andrii-kovalenko-ct: https://github.com/commercetools/commercetools-sync-java/issues/276

Issue #101, opened 2017-09-19
Price Sync


heshamMassoud:
Currently the sync always updates prices with `setPrices` update action regardless if needed or not. It should, however, calculate update actions, and only issue and update request if needed. 

### Price update actions
- [x] addPrice
- [x] setPrices: (should only be used in the case of no existing prices at all: https://github.com/sphereio/sphere-product-import/issues/117, `ChangePrice` should be used instead)
- [x] changePrice
- [x] removePrice
- [x] setProductPriceCustomType
- [ ] setProductPriceCustomField
- [ ] setDiscountedPrice

**Note:** Price order should not matter.

### Prices are uniquely identified by a combination of
1. `country`
2. `customerGroup`
3. `channel`
4. `validFrom`
5. `validUntil`

----------
1. The price matcher should first find the most precise match which has a match having the same exact 5 field values.
2. If not it, the next precise match would be one which has the same exact first 3 field values (`country`, `customerGroup` and `channel`).
3. If not, then there is no match and the prices should be created ( @butenkor right?)
------------ 
**Update**
Price matching would be based on above logic and not on key matching.

butenkor:
Correct. Additionally there should be flag `allowPriceRemoval`, which by default is set to `true` and would remove all prices in existing variant which are not defined in a new variant draft.

heshamMassoud:
> Additionally there should be flag `allowPriceRemoval`, which by default is set to true and would remove all prices in existing variant which are not defined in a new variant draft.

We have discussed earlier general options as boolean flags in the sync lib like: 
1. removeOtherLocales
2. removeOtherSetEntries
3. removeOtherCollectionEntries
4. removeOtherProperties

So prices would fall in the "removeOtherCollectionEntries" case. However, it would be better if we can specify exactly which collectionEntries to remove extra stuff from.  For example, only remove extra prices, but don't remove extra images. This I guess would mean something as described in this issue: https://github.com/commercetools/commercetools-sync-java/issues/51 . So changing all these boolean flags to blacklist of which resources to remove extra locales/set entries/collection entries/ properties. 

What do you think?

heshamMassoud:
@butenkor wouldn't have a price key save us all the hassle of checking the preciseness of price matching described in the issue description?

butenkor:
I also think that issue #51 is more suitable for blacklisting instead of separate boolean flags.
Keys for prices would be simpler indeed.

heshamMassoud:
[...]

PR #290, opened 2018-07-23
WIP: 101 price sync support

heshamMassoud:
#### Summary
This PR addresses adding price sync support to the product sync.

#### Description
TODO

#### Relevant Issues
#101 

#### Todo

- Tests
    - [x] Unit 
    - [ ] Integration
- [ ] Documentation


src/main/java/com/commercetools/sync/internals/utils/UnorderedCollectionSyncUtils.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param <T>                      type of the resulting update actions.
+     * @param <S>                      type of the new resource key.
+     * @param <U>                      type of the old resource.
+     * @param <V>                      type of the new resource.
+     * @return a list of all the remove update actions. If there are no missing matching drafts, an empty list is
+     *         returned.
+     */
+    @Nonnull
+    public static <T, S, U, V> List<UpdateAction<T>> buildRemoveUpdateActions(
+        @Nonnull final Map<S, U> oldResourcesMap,
```
jortizsao: This is particular opinion but I think would be better to pass the List of oldResources rather than a Map. I would create the map inside this function based on the list. why?

- The map is a "processed" parameter (the list of old resources is the original input) so it creates a dependency with the function that transforms the list to map.
- Easy to test and maintain because the code is more decoupled and we don't need to create this map in our unit tests
- The performance is not an issue even if we generate the map inside each buildUpdateAction (remove, add, etc...) since the cost is O(n).

src/main/java/com/commercetools/sync/products/utils/ProductVariantPriceUpdateActionUtils.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param newPrice  the price draft where we get the new name.
+     * @return A filled optional with the update action or an empty optional if the names are identical.
+     */
+    @Nonnull
+    public static Optional<ChangePrice> buildChangePriceUpdateAction(
+        @Nonnull final Price oldPrice,
+        @Nonnull final PriceDraft newPrice) {
+
+        final Optional<ChangePrice> actionAfterValuesDiff = buildUpdateAction(oldPrice.getValue(), newPrice.getValue(),
+            () -> ChangePrice.of(oldPrice, newPrice, true));
```
jortizsao: since `value` is a required field within Price (https://docs.commercetools.com/http-api-projects-products.html#pricedraft), shouldn't we validate that the `newPrice.getValue()` is not null? if so, should't we call to the warning callback of `ProductSyncOptions`?

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
```diff
@@ -117,24 +129,81 @@
...
+        final List<PriceDraft> newPrices = newProductVariant.getPrices();
+
+        final List<UpdateAction<Product>> updateActions = buildRemoveUpdateActions(oldPricesMap, newPrices,
+            PriceCompositeId::of, price -> RemovePrice.of(price, true));
+
+        final Integer variantId = oldProductVariant.getId();
+
+        emptyIfNull(newPrices).forEach(newPrice -> {
+            if (newPrice == null) {
+                syncOptions.applyErrorCallback(format("Failed to build prices update actions for one price on the "
```
jortizsao: How does the user know the product that the variant belongs to? would be this message enough for the user to identify the product and variant?

src/test/java/com/commercetools/sync/commons/MockUtils.java
```diff
@@ -44,6 +44,33 @@ public static CustomFieldsDraft getMockCustomFieldsDraft() {
...
+     * returned instance will return {@link Map} populated with given {@code fieldName} and {@code fieldValue}
+     *
+     * @param typeId custom type id
+     * @param fieldName custom field name
+     * @param fieldValue custom field value
+     * @return mock instance of {@link CustomFields}
+     */
+    public static CustomFields getMockCustomFields(final String typeId, final String fieldName,
+                                                   final JsonNode fieldValue) {
+        final CustomFields customFields = mock(CustomFields.class);
```
jortizsao: Why mocking rather than returning the actual `CustomFields` instance? are you going to test the behavior of the mock? 

I only see asserts for facts but not for behaviors in your tests.

src/test/java/com/commercetools/sync/products/utils/ProductSyncUtilsTest.java
```diff
@@ -148,16 +145,79 @@ public void buildActions_FromDraftsWithMultipleDifferentValues_ShouldBuildUpdate
             AddExternalImage.ofVariantId(1,
                 Image.of("https://53346cfbf3c7e017ed3d-6de74c3efa80f1c837c6a988b57abe66.ssl.cf3.rackcdn.com/"
                     + "new-image-2.png", ImageDimensions.of(0, 0), null), true),
-            SetPrices.of(1,
-                asList(PriceDraft.of(BigDecimal.valueOf(1.18), DefaultCurrencyUnits.EUR)
-                                 .withChannel(Channel.referenceOfId("channel-key_1")),
-                    PriceDraft.of(BigDecimal.valueOf(1), MoneyImpl.createCurrencyByCode("EGP"))
-                              .withChannel(Channel.referenceOfId("channel-key_1")))),
+            RemovePrice.of(Price.of(MoneyImpl.ofCents(108, EUR))
```
jortizsao: wow! this test looks like an all-in-one 😄. Too many asserts and setup code. I'm not sure how easy will be to maintain this test in the future

src/test/java/com/commercetools/sync/products/utils/ProductSyncUtilsTest.java
```diff
@@ -148,16 +145,79 @@ public void buildActions_FromDraftsWithMultipleDifferentValues_ShouldBuildUpdate
...
+            AddPrice.ofVariantId(1,
+                PriceDraft.of(MoneyImpl.ofCents(100, createCurrencyByCode("EGP")))
+                          .withChannel(Channel.referenceOfId("channel-key_1")), true),
             AddAsset.ofVariantId(1, assetDraft).withPosition(0).withStaged(true),
             SetSku.of(1, "3065831", true)
         );
     }
 
+    @Test
+    public void buildActions_FromDraftsWithMultipleDifferentValues_ShouldBuildUpdateActionsInCorrectOrder() {
```
jortizsao: wouldn't be better to just test that you are calling the right functions with the proper parameters (basically test the behavior)? I mean, product update actions are massive, so trying to test in the same unit test different kind of update actions makes the test huge, the setup gets complicated, and finally difficult to follow and maintain (it's brittle to small changes).

Since you have unit tests for the different types of update actions within the product ("general" update actions such as "Change Name", "Change Slug", etc..., variant update actions, variant images update actions, etc...), wouldn't be better to test for the ProductSyncTestUtils just the following:

"when I call to `ProductSyncUtils.buildActions` with this new product draft and this product, it should call `my_build_product_update_function` with the same product draft and the same product"

i.e when I call to `ProductSyncUtils.buildActions(oldProduct, newProductDraft, productSyncOptions, new HashMap<>())`, it should call `buildSetDescriptionUpdateAction` with `oldProduct`, `newProductDraft`. 

You already have a test for the "buildSetDescriptionUpdateAction", so testing that you are calling it with the proper parameters is enough.

And the same for the rest of build update actions functions... `buildVariantsUpdateActions`, 

For me this way helps:

- the developer to understand what's going on within the ProductTypeSyncUtils --> tests as documentation. Reading the test as it is, it's almost impossible to know why there has to be a RemoveImage.ofVariantId(1), AddExternalImage.ofVariantId(1), etc.. update actions.
- to easily identify the cause if the test fails because the scope is smaller. In this huge test, if an assert fails it's difficult to identify why. maybe it's been a bad setup with wrong values? maybe an actual error inside one of the multiple functions is calling?

Sorry for the long comment 🙈. But I think that for someone that has to maintain the code, this kind of tests are tough to fix and understand the cause if it fails.

It's just my opinion 🙂. WDYT?

src/main/java/com/commercetools/sync/internals/utils/UnorderedCollectionSyncUtils.java
heshamMassoud: agreed!: 5bb3cc4

src/main/java/com/commercetools/sync/products/utils/ProductVariantPriceUpdateActionUtils.java
heshamMassoud: good catch! I missed that. 2b80d56

Thanks :)

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
heshamMassoud: yes :) and the product variant key is unique across a project. (unfortunately, I didn't find this in the API documentation, but by trying it)

src/test/java/com/commercetools/sync/commons/MockUtils.java
heshamMassoud: Because the `CustomFields` cannot be instantiated through our JVM SDK, as it's treated as a read-only model, unlike `CustomFieldsDraft` which is the read-write model that can actually be instantiated.

src/test/java/com/commercetools/sync/products/utils/ProductSyncUtilsTest.java
heshamMassoud: It is indeed because it's testing the all-in-one util `ProductSyncUtils#buildActions` which does all of these things.

src/test/java/com/commercetools/sync/products/utils/ProductSyncUtilsTest.java
heshamMassoud: I think this is a great idea! :)

heshamMassoud: It only makes sense to check all the actual built update actions in the integration tests.

heshamMassoud: @jortizsao but I would rather create a separate PR for it, not to clutter the main goal of this one. I created an issue for it already: https://github.com/commercetools/commercetools-sync-java/issues/292

jortizsao:
LGTM

src/test/java/com/commercetools/sync/internals/utils/UnorderedCollectionSyncUtilsTest.java
```diff
@@ -0,0 +1,94 @@
...
+import static com.commercetools.sync.internals.utils.UnorderedCollectionSyncUtils.buildRemoveUpdateActions;
+import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class UnorderedCollectionSyncUtilsTest {
+
+    @Test
+    public void buildRemoveUpdateActions_withNullNewDraftsAndEmptyOldCollection_ShouldNotBuildActions() {
```
lojzatran: I just wonder - can you do parameterized test here as well?

src/test/java/com/commercetools/sync/products/utils/productvariantupdateactionutils/prices/PriceFixtures.java
```diff
@@ -0,0 +1,211 @@
...
+        getMockCustomFields("customType1", "foo", JsonNodeFactory.instance.textNode("X")));
+
+    public static final Price DE_22_USD = getPrice(BigDecimal.valueOf(22), USD, DE,
+        null, null, null, null, null);
+
+
+    @Nonnull
+    private static Price getPrice(@Nonnull final BigDecimal value,
+                                  @Nonnull final CurrencyUnit currencyUnits,
+
```
lojzatran: Why blank line here?

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
```diff
@@ -117,24 +128,78 @@
...
+     * @param updateActions list of update actions to sort.
+     * @return a new sorted list of update actions (remove, change, add).
+     */
+    @Nonnull
+    private static List<UpdateAction<Product>> sortPriceActions(
+        @Nonnull final List<UpdateAction<Product>> updateActions) {
+
+        final List<UpdateAction<Product>> actionsCopy = new ArrayList<>(updateActions);
+        actionsCopy.sort((action1, action2) -> {
+            if (action1 instanceof RemovePrice) {
```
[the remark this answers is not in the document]
heshamMassoud: @lojzatran actually you don't have to compare one-to-one as long as you cover all the cases. But thanks a lot for pointing it out, I discovered I wasn't covering all possible cases. 

I have now changed it into it's own internal utils: 3d0c5d2
I also added unit tests to cover all possible cases: 6246f76

src/test/java/com/commercetools/sync/internals/helpers/PriceCompositeIdTest.java
```diff
@@ -0,0 +1,230 @@
...
+    private static final String CASE_26 = "with currency, countryCode, validFrom and validUntil";
+    private static final String CASE_27 = "with currency, channel, validFrom and validUntil";
+    private static final String CASE_28 = "with currency, countryCode, channel, validFrom and validUntil";
+    private static final String CASE_29 = "with currency, customerGroup, validFrom and validUntil";
+    private static final String CASE_30 = "with currency, countryCode, customerGroup, validFrom and validUntil";
+    private static final String CASE_31 = "with currency, channel, customerGroup, validFrom and validUntil";
+    private static final String CASE_32 =
+        "with currency, countryCode, channel, customerGroup, validFrom and validUntil";
+
+    @ParameterizedTest(name = "[#ofPriceDraft]: {0}")
```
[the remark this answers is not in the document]
heshamMassoud: Yes, I'm loving Junit5 till now. Still want to investigate Spock in the future. Both have their advantages and disadvantages I guess. For me the biggest disadvantage with spock is writing in groovy -> so implementation code will be different syntax than test code. 


Also, regarding the hard to read stuff, I think the tests in the link you sent are very trivial compared to actual tests in bigger projects like this lib, so I don't think in spock it will be That much easier ;)

heshamMassoud:
@lojzatran about your question:
> What about max. 700 prices? Are you going to check it somehow? 

No, we decided not to do it. We rely on the platform error in validation cases. 

The library doesn't mirror all the validations of the API. It will make the logic of the library very complex and also much hard to maintain in the future with every changing constraint of validation.

src/main/java/com/commercetools/sync/internals/helpers/PriceCompositeId.java
```diff
@@ -0,0 +1,133 @@
...
+    public ZonedDateTime getValidFrom() {
+        return validFrom;
+    }
+
+    public ZonedDateTime getValidUntil() {
+        return validUntil;
+    }
+
+    @Override
+    public boolean equals(final Object otherObject) {
```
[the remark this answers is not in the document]
heshamMassoud: yes It's actually generated by Intellij.

This condition covers the null check you mentioned:

````java
if (!(otherObject instanceof PriceCompositeId)) {
            return false;
        }
````

File: src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java [hunk 12]
```diff
@@ -386,45 +388,42 @@
...
+                    final List<UpdateAction<Product>> updateOrAddVariant = ofNullable(matchingOldVariant)
+                        .map(oldVariant -> collectAllVariantUpdateActions(oldProduct, oldVariant, newProductVariant,
+                            attributesMetaData, syncOptions))
+                        .orElseGet(() -> singletonList(buildAddVariantUpdateActionFromDraft(newProductVariant)));
+                    updateActions.addAll(updateOrAddVariant);
+                }
             }
+        });
 
-            // 2.1 if both old/new variants lists have an item with the same key - create update actions for the variant
```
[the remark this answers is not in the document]
heshamMassoud: because it's not needed. It's cluttering the code and the code itself below it actually self explains it 😉

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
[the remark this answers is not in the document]
heshamMassoud: @lojzatran after our discussion, to follow the provisions in the effective java book. In order to take care of cases like not changing order of two actions of the same time (Which we actually don't care about atm) but still It's considered as a side-effect of the sort. 

Now it's changed and added tests to test for all provisions possible: 
36a9951

src/test/java/com/commercetools/sync/internals/utils/UnorderedCollectionSyncUtilsTest.java
heshamMassoud: yes you definitely can, but I though it's more clear this way have a separate method for each case since there are not a lot of different cases.

src/test/java/com/commercetools/sync/products/utils/productvariantupdateactionutils/prices/PriceFixtures.java
heshamMassoud: that was weird. Thanks for noticing: 82af316

src/main/java/com/commercetools/sync/products/utils/ProductVariantUpdateActionUtils.java
```diff
@@ -117,24 +128,78 @@
...
-     * TODO: calculated GITHUB ISSUE#101.
-     *
-     * <p>TODO: UNIT TESTS
-     *
      * @param oldProductVariant the {@link ProductVariant} which should be updated.
      * @param newProductVariant the {@link ProductVariantDraft} where we get the new list of prices.
+     * @param syncOptions the sync options wrapper which contains options related to the sync process supplied by
+     *                    the user. For example, custom callbacks to call in case of warnings or errors occurring
+     *                    on the build update action process. And other options (See {@link ProductSyncOptions}
+     *                    for more info.
```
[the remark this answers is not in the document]
heshamMassoud: thanks: 9769460

src/main/java/com/commercetools/sync/internals/helpers/PriceCompositeId.java
```diff
@@ -0,0 +1,133 @@
...
+import javax.annotation.Nullable;
+import java.time.ZonedDateTime;
+import java.util.Objects;
+
+import static java.util.Optional.ofNullable;
+
+/**
+ * This class is only meant for the internal use of the commercetools-sync-java library.
+ */
+public final class PriceCompositeId {
```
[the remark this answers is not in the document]
heshamMassoud: I added it now: 
dc7aa33

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncWithAssetsIT.java
```diff
@@ -288,7 +285,7 @@ public void sync_withMatchingProductWithDuplicateAssets_shouldFailToUpdateProduc
             executeBlocking(productSync.sync(singletonList(productDraft)));
 
         // Assert results of sync
-        assertThat(syncStatistics).hasValues(1, 0, 1, 0);
+        assertThat(syncStatistics).hasValues(1, 0, 0, 0);
```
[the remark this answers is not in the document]
heshamMassoud: Before, there was always an update happening, since we were always issuing a setPrice action to copy the prices. But now after there is a price sync. We don't do that anymore, we only update if needed based on built actions from comparison.

docs/RELEASE_NOTES.md
```diff
@@ -49,18 +50,42 @@
...
+-->
+
+### v1.0.0-M13 -  Jul 26, 2018
 [Commits](https://github.com/commercetools/commercetools-sync-java/compare/v1.0.0-M12...v1.0.0-M13) |
 [Javadoc](https://commercetools.github.io/commercetools-sync-java/v/v1.0.0-M13/) | 
 [Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/v1.0.0-M13)
--->
+
+**New Features** (4) 
+- **Product Sync** - Support for syncing product prices. [#101](https://github.com/commercetools/commercetools-sync-java/issues/101)
```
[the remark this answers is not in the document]
heshamMassoud: check the comment here: https://github.com/commercetools/commercetools-sync-java/pull/290#discussion_r205417894

PR #291, opened 2018-07-25
286 product type attributes actions one to one

jortizsao:
#### Summary
This PR adds the product type attribute definition update actions for those attribute definitions that match by name.

#### Description
- `AttributeDefinitionUpdateActionUtils` 
   - Change Attribute Definition Label update action
   - Set Attribute Definition Input Tip update action
   - Change Attribute Definition isSearchable update action
   - Change Attribute Definition inputHint update action
   - Change Attribute Definition attribute contraint update action
   - Remove unnecessary null validations for required fields
   - Add DuplicateNameException unit test

#### Todo

- Tests
    - [X] Unit

### DIFF
--- a/src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
@@ -164,4 +164,6 @@ public final class AttributeDefinitionUpdateActionUtils {
             )
         );
     }
+
+    private AttributeDefinitionUpdateActionUtils() { }
 }
--- a/src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
@@ -27,26 +27,25 @@ public final class AttributeDefinitionUpdateActionUtils {
      * and the {@link AttributeDefinitionDraft} have identical fields, then no update action is needed and hence an
      * empty {@link List} is returned.
      *
-     * @param oldAttributeDefinition        the attribute definition which should be updated.
-     * @param newAttributeDefinitionDraft   the attribute definition draft where we get the new fields.
+     * @param oldAttributeDefinition      the attribute definition which should be updated.
+     * @param newAttributeDefinitionDraft the attribute definition draft where we get the new fields.
      * @return A list with the update actions or an empty list if the attribute definition fields are identical.
-     *
      */
     @Nonnull
     public static List<UpdateAction<ProductType>> buildActions(
-        @Nonnull final AttributeDefinition oldAttributeDefinition,
-        @Nonnull final AttributeDefinitionDraft newAttributeDefinitionDraft) {
+            @Nonnull final AttributeDefinition oldAttributeDefinition,
+            @Nonnull final AttributeDefinitionDraft newAttributeDefinitionDraft) {
 
         return Stream.of(
-            buildChangeLabelUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
-            buildSetInputTipUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
-            buildChangeIsSearchableUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
-            buildChangeInputHintUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
-            buildChangeAttributeConstraintUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft)
+                buildChangeLabelUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+                buildSetInputTipUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+                buildChangeIsSearchableUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+                buildChangeInputHintUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+                buildChangeAttributeConstraintUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft)
         )
-        .filter(Optional::isPresent)
-        .map(Optional::get)
-        .collect(toList());
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(toList());
     }
 
     /**
@@ -61,14 +60,12 @@ public final class AttributeDefinitionUpdateActionUtils {
      */
     @Nonnull
     public static Optional<UpdateAction<ProductType>> buildChangeLabelUpdateAction(
-        @Nonnull final AttributeDefinition oldAttributeDefinition,
-        @Nonnull final AttributeDefinitionDraft newAttributeDefinition) {
+            @Nonnull final AttributeDefinition oldAttributeDefinition,
+            @Nonnull final AttributeDefinitionDraft newAttributeDefinition) {
 
         return buildUpdateAction(oldAttributeDefinition.getLabel(), newAttributeDefinition.getLabel(),
-            () -> ChangeAttributeDefinitionLabel.of(
-                oldAttributeDefinition.getName(),
-                newAttributeDefinition.getLabel()
-            )
+            () -> ChangeAttributeDefinitionLabel.of(oldAttributeDefinition.getName(),
+                    newAttributeDefinition.getLabel())
         );
     }
 
@@ -84,14 +81,11 @@ public final class AttributeDefinitionUpdateActionUtils {
      */
     @Nonnull
     public static Optional<UpdateAction<ProductType>> buildSetInputTipUpdateAction(
-        @Nonnull final AttributeDefinition oldAttributeDefinition,
-        @Nonnull final AttributeDefinitionDraft newAttributeDefinition) {
+            @Nonnull final AttributeDefinition oldAttributeDefinition,
+            @Nonnull final AttributeDefinitionDraft newAttributeDefinition) {
 
         return buildUpdateAction(oldAttributeDefinition.getInputTip(), newAttributeDefinition.getInputTip(),
-            () -> SetInputTip.of(
-                oldAttributeDefinition.getName(),
-                newAttributeDefinition.getInputTip()
-            )
+            () -> SetInputTip.of(oldAttributeDefinition.getName(), newAttributeDefinition.getInputTip())
         );
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
```diff
@@ -20,20 +29,139 @@
...
 
-        return Collections.emptyList();
+        return buildUpdateAction(oldAttributeDefinition.getAttributeConstraint(),
+            newAttributeDefinition.getAttributeConstraint(),
+            () -> ChangeAttributeConstraint.of(
+                oldAttributeDefinition.getName(),
+                newAttributeDefinition.getAttributeConstraint()
+            )
+        );
     }
```
heshamMassoud: don't forget the private empty constructor ;)

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
```diff
@@ -180,8 +170,7 @@
                     .map(attributeDefinitionDraft ->
                         buildActions(
                             oldAttributeDefinition,
-                            attributeDefinitionDraft,
-                            syncOptions
+                            attributeDefinitionDraft
```
heshamMassoud: why not all on the same line :) ?
````java
.map(attributeDefinitionDraft -> buildActions(oldAttributeDefinition, attributeDefinitionDraft))
````

File: src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java [hunk 3]
```diff
@@ -20,20 +29,139 @@
...
+    @Nonnull
+    public static Optional<UpdateAction<ProductType>> buildChangeInputHintUpdateAction(
+        @Nonnull final AttributeDefinition oldAttributeDefinition,
+        @Nonnull final AttributeDefinitionDraft newAttributeDefinition) {
+
+        return buildUpdateAction(oldAttributeDefinition.getInputHint(), newAttributeDefinition.getInputHint(),
+            () -> ChangeInputHint.of(
+                oldAttributeDefinition.getName(),
+                newAttributeDefinition.getInputHint()
+            )
```
heshamMassoud: This fits on one line too.

src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
```diff
@@ -20,20 +29,139 @@
...
-        @Nonnull final AttributeDefinitionDraft newAttributeDefinitionDraft,
-        @Nonnull final ProductTypeSyncOptions syncOptions) {
+        @Nonnull final AttributeDefinitionDraft newAttributeDefinitionDraft) {
+
+        return Stream.of(
+            buildChangeLabelUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+            buildSetInputTipUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+            buildChangeIsSearchableUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+            buildChangeInputHintUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft),
+            buildChangeAttributeConstraintUpdateAction(oldAttributeDefinition, newAttributeDefinitionDraft)
```
heshamMassoud: are enums actions going to be added here too?

src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,214 @@
...
+            .inputHint(null)
+            .isSearchable(true)
+            .build();
+    }
+
+    @Test
+    public void buildChangeLabelAction_WithDifferentValues_ShouldReturnAction() {
+        final Optional<UpdateAction<ProductType>> result = buildChangeLabelUpdateAction(old, newDifferent);
+
+        assertThat(result).containsInstanceOf(ChangeAttributeDefinitionLabel.class);
```
heshamMassoud: not sure if you need this assert, isn't the following one enough?

src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
jortizsao: If I put in one line I get an error from checkstyle that line length is greater than 120 characters

`[ant:checkstyle] [WARN] /Users/javierortiz/Projects/commercetools/platform/commercetools-sync-java/src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java:161: Line is longer than 120 characters (found 129). [LineLength]`

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
jortizsao: this was due to there was an third parameter that mades the line being greater than 120, so after removing this third parameter I forgot to format it in one line

src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
jortizsao: No, they have its own class. Thus, it's easier to test only the enums separated from the attribute definitions unit tests. 

Furthermore, if in the future there is more update actions for enums, they already have its own class

src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java
jortizsao: I saw it in another tests. I agree. I remove it

src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
jortizsao: 3f1adcd1f054f72e37c1ab02e1bb844b8af45b7c

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
jortizsao: 1560ab147719ff45e48e57b7ff470a836f225d9d

src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java
jortizsao: e06fbb36efabf1d88963f9c4f175f76a51c73e36

src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
heshamMassoud: It shouldn't be longer if you have it like that: 
````java
return buildUpdateAction(oldAttributeDefinition.getInputHint(), newAttributeDefinition.getInputHint(), 
            () -> ChangeInputHint.of(oldAttributeDefinition.getName(), newAttributeDefinition.getInputHint()));
````

**Comment by jortizsao** [comment]
> We have it as convention in the library to indent chaining on the same chain dot. This way its with the style of the library

Ok, I try to follow the convention as much as possible in order to give consistency to the whole style of the application. However, I find several drawbacks to this style:

1. If you chain several functions, in those cases where the function has a body you are shifting to the right, and finally you don't have space for new function calls because of the line length limitation. 

See an example from `ProductSync#processBatch` (another function call with body would be impossible)

https://github.com/commercetools/commercetools-sync-java/blob/master/src/main/java/com/commercetools/sync/products/ProductSync.java#L121

it's true that you can avoid this by taking body function out so the chain is more lineal, but in cases where the function transformation is just one line ( i.e ```.map(value -> value + 1)```) doesn't make too much sense to take it to an external function.

2. This style doesn't match with the tabs of the editors, because the dot `.` falls in any position (matching the tab or not), so the following lines that are according to the first `.` are likely to not follow the tab indentation. Thus, when you copy and paste from one editor to another the code is broken in some cases (i.e if I copy and paste the code here in a github comment, the code is not aligned to the `.` and looks broken)

3. would't be good to use some kind of automatic formatter in order to avoid the subjectivity in the formatting? The same way as Prettier (https://github.com/prettier/prettier) in javascript. If you are already using one, please let me know and I install it.




> Some times you break the lines of method calls (as I put some comments) so it's on different lines. I can see that maybe it's clearer for you this way. But on the other hand, I also think it's clear on one line. But this is a relative thing for everyone. But what's important is that we keep a consistent style across the codebase so it's easier for a maintainer/dev to read without distractions.

The convention that I'm following is"

1. if it fits in one line, I put all parameters in one line
2 otherwise, I break down the parameters one per line, why?

- it's easy to remove a parameter by deleting the whole line (usually using an editor shortcut)
- the diffs are more clear when you add/delete a parameter. Similar to what encourages the comma-dangle eslint rule (https://eslint.org/docs/rules/comma-dangle). it's for javascript but the reason also applies here.

My opinion is that there are not good/bad rules and we should follow the same formatting, so in my opinion we should have any kind of automatic formatting in order to avoid these issues and what is more important, spending time in formatting according to it.

**Comment by heshamMassoud** [comment]
@jortizsao thanks for your constructive feedback. I agree with you about the autoformatting to avoid wasting time on such style issues, I created an issue for it here: https://github.com/commercetools/commercetools-sync-java/issues/293

For now, as we discussed, let's stick to the current style to avoid mixing up styles. We can change the whole formatting of the whole project in a separate PR, where we can discuss the best style approaches, or when we find setup the autoformatter tool we can agree on the configurations 😉

File: src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java [hunk 6]
```diff
@@ -18,22 +27,131 @@
...
+     * @param newAttributeDefinitionDraft the attribute definition draft where we get the new fields.
      * @return A list with the update actions or an empty list if the attribute definition fields are identical.
-     *
      */
     @Nonnull
     public static List<UpdateAction<ProductType>> buildActions(
         @Nonnull final AttributeDefinition oldAttributeDefinition,
-        @Nonnull final AttributeDefinitionDraft newAttributeDefinitionDraft,
-        @Nonnull final ProductTypeSyncOptions syncOptions) {
+        @Nonnull final AttributeDefinitionDraft newAttributeDefinitionDraft) {
```
heshamMassoud: no unit tests for this method?

jortizsao: I haven't added a unit test for this method. I added a test in the integration tests, since this is a method that calls to other methods that already have its own unit test.

heshamMassoud:
Hi @jortizsao can you please don't forget to update the release notes with newly exposed stuff in this PR? thanks.

PR #294, opened 2018-07-30
286 product type attributes enum actions

jortizsao:
#### Summary
This PR adds the attribute definition update actions for plain enum values and localized enum values

#### Description
- `AttributeDefinitionUpdateActionUtils` 
   - Add PlainEnumValue to attribute definition update action
   - Change order of PlainEnumValue update action
   - Remove PlainEnumValues from attribute definition update action
   - Change plain enum value label update action
   - Add LocalizedEnumValue to attribute definition update action
   - Change order of LocalizedEnumValue update action
   - Remove LocalizedEnumValues from attribute definition update action
   - Change localized enum value label update action

#### Todo

- Tests
    - [X] Unit

### DIFF
--- a/src/main/java/com/commercetools/sync/commons/exceptions/DifferentTypeException.java
+++ b/src/main/java/com/commercetools/sync/commons/exceptions/DifferentTypeException.java
@@ -1,0 +1,17 @@
+package com.commercetools.sync.commons.exceptions;
+
+import javax.annotation.Nonnull;
+
+public class DifferentTypeException extends RuntimeException {
+    public DifferentTypeException(@Nonnull final String message) {
+        super(message);
+    }
+
+    public DifferentTypeException(@Nonnull final Throwable cause) {
+        super(cause);
+    }
+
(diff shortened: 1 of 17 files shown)


jortizsao:
The `ProductTypeUpdateLocalizedEnumActionUtils` and `ProductTypeUpdatePlainEnumActionUtils` classes are similar in the `buildUpdateAction` method and someone could argue that I could also reuse the code in the `ProductTypeUpdateEnumActionsUtils` class, the same as it's done with other methods.

I did it that way first but the code ended up more tight and difficult to understand since all three particular methods for adding, removing and change order of the enums have to be passed as parameters in the common method placed at ProductTypeUpdateEnumActionsUtils, making this method to have 6 parameters where some of them are BiFunction, TriFunction...  (long and "complex" signature)

So I finally decided to repeat some code in `ProductTypeUpdateLocalizedEnumActionUtils` and `ProductTypeUpdateLocalizedEnumActionUtils` in order to clarify and decouple the `buildUpdateAction` method of both classes.

PR #296, opened 2018-08-01
286 Add product type sync

jortizsao:
#### Summary
This PR adds the ProductTypeSync class which is responsible for creating the batches from the product type drafts, and sync them accordingly:
- if it is a new product type draft, creates the product type in commercetools
- if it is an existing product type, updates the product type with the data coming from the product type draft.

#### Description
- `ProductTypeSync` 
   - Add process method
   - Add process batch method
   - Add update existing product type method
   - Add create product type method
   - Update product type service
   - Add integration tests
   - Add documentation

#### Todo

- Tests
    - [X] Unit 
    - [X] Integration


src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeSyncUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Compares all the fields (including the attributes see
+     * {@link ProductTypeUpdateActionUtils#buildAttributesUpdateActions}) of a {@link ProductType} and a
+     * {@link ProductTypeDraft}. It returns a {@link List} of {@link UpdateAction}&lt;{@link ProductType}&gt; as a
+     * result. If no update action is needed, for example in case where both the {@link ProductType} and the
+     * {@link ProductTypeDraft} have the same description, an empty {@link List} is returned.
+     *
+     * @param oldProductType the product which should be updated.
+     * @param newProductType the product draft where we get the new data.
```
heshamMassoud: `the product draft where we get the new data.` -> `the {@link ProductTypeDraft} which should be updated.`

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeSyncUtils.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param syncOptions    the sync options wrapper which contains options related to the sync process supplied by
+     *                       the user. For example, custom callbacks to call in case of warnings or errors occurring
+     *                       on the build update action process. And other options (See {@link ProductTypeSyncOptions}
+     *                       for more info.
+     * @return A list of product type-specific update actions.
```
heshamMassoud: productType-specific

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
```diff
@@ -125,7 +126,7 @@
 
             return updateActions;
 
-        } catch (final DuplicateNameException | DuplicateKeyException | DifferentTypeException exception) {
+        } catch (final DuplicateNameException | DuplicateKeyException exception) {
             throw new BuildUpdateActionException(exception);
```
heshamMassoud: Are there unit tests for the case where the `BuildUpdateActionException` is thrown here. I tried to find unit tests for `ProductTypeUpdateAttributeDefinitionActionUtils#buildAttributeDefinitionsUpdateActions` but didn't find any :(

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
```diff
@@ -158,15 +159,44 @@
...
+                                return buildActions(oldAttributeDefinition, attributeDefinitionDraft);
+                            } else {
+                                return Arrays.asList(
+                                    RemoveAttributeDefinition.of(oldAttributeDefinitionName),
+                                    AddAttributeDefinition
+                                        .of(AttributeDefinitionCustomBuilder.of(attributeDefinitionDraft))
+                                );
+                            }
+                        } else {
+                            return new ArrayList<UpdateAction<ProductType>>();
```
heshamMassoud: but I don't see you doing as the comment actually says, so if the attributeType on the draft is null, no actions will be built and also the user won't get any warning/error?

Why not actually warn the user about the null type?

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
```diff
@@ -158,15 +159,44 @@
...
-                    )
+                    .map(attributeDefinitionDraft -> {
+                        // attribute type is required so if null we let commercetools to throw exception
+                        if (attributeDefinitionDraft.getAttributeType() != null) {
+                            if (haveSameAttributeType(oldAttributeDefinition, attributeDefinitionDraft)) {
+                                return buildActions(oldAttributeDefinition, attributeDefinitionDraft);
+                            } else {
+                                return Arrays.asList(
+                                    RemoveAttributeDefinition.of(oldAttributeDefinitionName),
+                                    AddAttributeDefinition
```
heshamMassoud: ah ok, I see you doing it here :)

heshamMassoud: I think it would be important to put in the Javadoc of the method and/or the documentation of the productType sync.

src/main/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
```diff
@@ -36,92 +33,63 @@
...
-                ));
-            }
-        } else {
-            throw new DifferentTypeException(format("The attribute type of the attribute definitions are different. "
-                + "Attribute name: '%s'. Old attribute type: '%s', new attribute type: '%s'. "
-                + "Attribute type has to remain the same for the same attribute name.",
-            oldAttributeDefinition.getName(),
-            oldAttributeDefinition.getAttributeType().getClass().getName(),
-            newAttributeDefinitionDraft.getAttributeType().getClass().getName()));
+        updateActions = Stream
```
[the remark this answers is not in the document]
heshamMassoud: ok I saw it down there. Please disregard this comment.

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
```diff
@@ -158,15 +159,44 @@
...
+                                        .of(AttributeDefinitionCustomBuilder.of(attributeDefinitionDraft))
+                                );
+                            }
+                        } else {
+                            return new ArrayList<UpdateAction<ProductType>>();
+                        }
+
+                    })
```
heshamMassoud: general questions about this method: Why does it take already a map `newAttributeDefinitionDraftsNameMap ` as a parameter and not a list? can't the map be already created inside of it   as it's only needed by it?

src/test/java/com/commercetools/sync/producttypes/utils/producttuypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java
```diff
@@ -1,14 +1,14 @@
...
 import com.commercetools.sync.commons.exceptions.BuildUpdateActionException;
-import com.commercetools.sync.commons.exceptions.DifferentTypeException;
 import com.commercetools.sync.commons.exceptions.DuplicateNameException;
 import com.commercetools.sync.producttypes.ProductTypeSyncOptions;
 import com.commercetools.sync.producttypes.ProductTypeSyncOptionsBuilder;
 import io.sphere.sdk.client.SphereClient;
 import io.sphere.sdk.commands.UpdateAction;
 import io.sphere.sdk.products.attributes.AttributeDefinition;
 import io.sphere.sdk.products.attributes.AttributeDefinitionBuilder;
+import io.sphere.sdk.products.attributes.LocalizedStringAttributeType;
```
heshamMassoud: There is a type in the name of the package of this test class:
`producttuypeactionutils` -> `producttypeactionutils` ;)

src/test/java/com/commercetools/sync/commons/asserts/statistics/ProductTypeSyncStatisticsAssert.java
```diff
@@ -0,0 +1,13 @@
+package com.commercetools.sync.commons.asserts.statistics;
+
+import com.commercetools.sync.producttypes.helpers.ProductTypeSyncStatistics;
+
+import javax.annotation.Nullable;
+
+public class ProductTypeSyncStatisticsAssert extends
```
heshamMassoud: Great that you are creating your own statistics assertion 🎉 

but maybe `final` 😊 ?

src/main/java/com/commercetools/sync/services/impl/ProductTypeServiceImpl.java
```diff
@@ -42,54 +49,79 @@ public ProductTypeServiceImpl(@Nonnull final ProductSyncOptions syncOptions) {
...
+                    } else {
+                        syncOptions.applyWarningCallback(format("ProductType with id: '%s' has no key set. Keys are"
+                                + " required for productType matching.", id));
+                    }
+                });
 
         return CtpQueryUtils.queryAll(syncOptions.getCtpClient(), ProductTypeQuery.of(), productTypePageConsumer)
-                            .thenAccept(result -> isCached = true)
-                            .thenApply(result -> Optional.ofNullable(keyToIdCache.get(key)));
+                .thenAccept(result -> isCached = true)
```
heshamMassoud: :( indentation

src/main/java/com/commercetools/sync/services/impl/ProductTypeServiceImpl.java
```diff
@@ -42,54 +49,79 @@ public ProductTypeServiceImpl(@Nonnull final ProductSyncOptions syncOptions) {
...
     }
 
     @Nonnull
     private static Map<String, AttributeMetaData> getAttributeMetaDataMap(@Nonnull final ProductType productType) {
         return productType.getAttributes().stream()
-                          .map(AttributeMetaData::of)
-                          .collect(
-                              Collectors.toMap(AttributeMetaData::getName, attributeMetaData -> attributeMetaData));
+                .map(AttributeMetaData::of)
+                .collect(
```
heshamMassoud: :( indentation

src/main/java/com/commercetools/sync/producttypes/ProductTypeSyncOptions.java
```diff
@@ -14,7 +14,7 @@
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-public final class ProductTypeSyncOptions extends BaseSyncOptions<ProductType, ProductTypeDraft> {
+public class ProductTypeSyncOptions extends BaseSyncOptions<ProductType, ProductTypeDraft> {
```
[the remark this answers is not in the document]
jortizsao: honestly, I don't know... it's been a mistake

src/main/java/com/commercetools/sync/producttypes/helpers/ProductTypeSyncStatistics.java
```diff
@@ -0,0 +1,23 @@
+package com.commercetools.sync.producttypes.helpers;
+
+import com.commercetools.sync.commons.helpers.BaseSyncStatistics;
+
+import static java.lang.String.format;
+
+public class ProductTypeSyncStatistics extends BaseSyncStatistics {
```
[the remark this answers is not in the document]
jortizsao: I created an integration test for that: https://github.com/commercetools/commercetools-sync-java/pull/296/files#diff-07965ad8d45b1a77e4bf0fc63ff109f8R393

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
jortizsao: Yes, there is https://github.com/commercetools/commercetools-sync-java/blob/master/src/test/java/com/commercetools/sync/producttypes/utils/producttuypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java#L159

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
jortizsao: Attribute type is required by commercetools, so if I'm not mistaken, in one conversation we had (and previously discussed with Roman), we decided to not add the validations in the library that are documented in commercetools  (like required values), and let commercetools to throw the error.

Did I get it wrong when we discussed it?

src/main/java/com/commercetools/sync/producttypes/helpers/ProductTypeSyncStatistics.java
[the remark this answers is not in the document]
jortizsao: > We should have private constructor for this class, since we only expect it to be instantiated through the builder. Also we should make it final. I know this is not the case for the other *Statistics classes, but maybe we should change that.

Is there a builder for the *Statistics classes? or you mean that I have to develop one?

src/main/java/com/commercetools/sync/producttypes/ProductTypeSyncOptions.java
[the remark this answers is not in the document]
jortizsao: 34ffaea3b1fd6be13db18a4d4e2a52e1998e79a6

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeSyncUtils.java
```diff
@@ -0,0 +1,56 @@
...
+public final class ProductTypeSyncUtils {
+
+    /**
+     * Compares all the fields (including the attributes see
+     * {@link ProductTypeUpdateActionUtils#buildAttributesUpdateActions}) of a {@link ProductType} and a
+     * {@link ProductTypeDraft}. It returns a {@link List} of {@link UpdateAction}&lt;{@link ProductType}&gt; as a
+     * result. If no update action is needed, for example in case where both the {@link ProductType} and the
+     * {@link ProductTypeDraft} have the same description, an empty {@link List} is returned.
+     *
+     * @param oldProductType the product which should be updated.
```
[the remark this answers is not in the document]
jortizsao: 7a196a6c573acbde9603e1046cb35d06b450626c

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeSyncUtils.java
jortizsao: 7a196a6c573acbde9603e1046cb35d06b450626c

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
jortizsao: 7a196a6c573acbde9603e1046cb35d06b450626c

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
jortizsao: 6f72b6200fd20fa1017916e7be71e9880eb3fc8b

src/test/java/com/commercetools/sync/producttypes/utils/producttuypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java
jortizsao: 83a625c638edb93103fca429e43f00bced3cdde0

src/test/java/com/commercetools/sync/commons/asserts/statistics/ProductTypeSyncStatisticsAssert.java
jortizsao: 23f9597c9a5e0d2d072c0622d84677bd8e034569

src/main/java/com/commercetools/sync/services/impl/ProductTypeServiceImpl.java
jortizsao: 2b9a06a9309658a9869d5b71852f4eccc02efd94

src/main/java/com/commercetools/sync/services/impl/ProductTypeServiceImpl.java
jortizsao: 2b9a06a9309658a9869d5b71852f4eccc02efd94

src/main/java/com/commercetools/sync/producttypes/helpers/ProductTypeSyncStatistics.java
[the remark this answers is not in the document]
heshamMassoud: @jortizsao 

> I created an integration test for that

I would say:
1. We should cover as much of the code (as possible) with unit tests.
2. a unit test is sufficient for the statistics since the integration tests have an overhead of the setting up/tearing down requests. So they are quite costly in the sense that they increase the time of the build significantly, etc..

see: https://martinfowler.com/bliki/TestPyramid.html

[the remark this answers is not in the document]
heshamMassoud: > Is there a builder for the *Statistics classes? or you mean that I have to develop one?

Oh, my bad. I thought there is already one :( disregard the comment then. Just consider using `final`..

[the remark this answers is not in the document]
jortizsao: > I would say:
> 
> We should cover as much of the code (as possible) with unit tests.
> a unit test is sufficient for the statistics since the integration tests have an overhead of the setting up/tearing down requests. So they are quite costly in the sense that they increase the time of the build significantly, etc..

Yes, I'm familiar with the test pyramid and I love unit tests (I usually develop using TDD when possible) and I prefer them over integration tests.

For the ProductTypeSync class and related classes and methods (basically the ones I've added in this PR and that don't create update actions), I found complex to add unit tests to these classes due to coupling/architecture, so I decided to follow the same existing approach (cover those test cases with integration tests ).

In this case is true that is possible to add a unit test for the `getReportMessage` method but in order to avoid a single test class for this unit test, I thought that the integration test was enough. 

Anyways, I add the unit test :)

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
heshamMassoud: - Hmm, but the thing is I don't see how CT will throw an exception this case. Because this method will return an empty list so no actual update action will be issued to CT. Or do you mean only in case if there is an update action from the calling method (`ProductTypeUpdateAttributeDefinitionActionUtils#buildUpdateActions`) then CT will throw an exception?

- Also, Please remove the `throws DuplicateKeyException` declaration on the method signature ;)

jortizsao: > do you mean only in case if there is an update action from the calling method (ProductTypeUpdateAttributeDefinitionActionUtils#buildUpdateActions) then CT will throw an exception?

Yes, I meant that. If there is any update action CT will throw an error no matter the update actions

jortizsao: > Also, Please remove the throws DuplicateKeyException declaration on the method signature ;)

9f0255dafd8b7fd1a10703e21913f6182ce3cb34

src/main/java/com/commercetools/sync/producttypes/helpers/ProductTypeSyncStatistics.java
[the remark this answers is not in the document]
jortizsao: f2e2841c61d90f1f0cd0430767b6ee687b7a574d

src/test/java/com/commercetools/sync/producttypes/helpers/ProductTypeSyncStatisticsTest.java
```diff
@@ -0,0 +1,27 @@
...
+    @Before
+    public void setup() {
+        productTypeSyncStatistics = new ProductTypeSyncStatistics();
+    }
+
+    @Test
+    public void getReportMessage_WithIncrementedStats_ShouldGetCorrectMessage() {
+        productTypeSyncStatistics.incrementCreated(1);
+        productTypeSyncStatistics.incrementFailed(1);
+        productTypeSyncStatistics.incrementUpdated(1);
```
lojzatran: I would give created=1, failed=2, updated=3 so you know that you didn't mix the variables.

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,298 @@
...
+     *
+     * @param productTypeDrafts {@link List} of {@link ProductTypeDraft}'s that would be synced into CTP project.
+     * @return {@link CompletionStage} with {@link ProductTypeSyncStatistics} holding statistics of all sync
+     *         processes performed by this sync instance.
+     */
+    @Override
+    protected CompletionStage<ProductTypeSyncStatistics> process(
+            @Nonnull final List<ProductTypeDraft> productTypeDrafts) {
+
+        final List<List<ProductTypeDraft>> batches = batchElements(productTypeDrafts, syncOptions.getBatchSize());
```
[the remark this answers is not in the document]
jortizsao: The invalid drafts count as a processed items in the statistics so they have to be "processed" (they are not actually processed, but the process method will consider them as a processed items)

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,298 @@
...
+    protected CompletionStage<ProductTypeSyncStatistics> processBatch(@Nonnull final List<ProductTypeDraft> batch) {
+        final List<ProductTypeDraft> validProductTypeDrafts = batch.stream()
+                .filter(this::validateDraft)
+                .collect(toList());
+
+        if (validProductTypeDrafts.isEmpty()) {
+            statistics.incrementProcessed(batch.size());
+            return completedFuture(statistics);
+        } else {
+            final Map<String, ProductTypeDraft> keysProductTypeDraftMap = getKeysProductTypeMap(validProductTypeDrafts);
```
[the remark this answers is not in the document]
jortizsao: I wanted to reuse the existing method `getKeysProductTypeMap`, but I change it.

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateAttributeDefinitionActionUtils.java
```diff
@@ -158,15 +159,44 @@
                 final AttributeDefinitionDraft matchingNewAttributeDefinitionDraft =
                     newAttributeDefinitionDraftsNameMap.get(oldAttributeDefinitionName);
                 return ofNullable(matchingNewAttributeDefinitionDraft)
-                    .map(attributeDefinitionDraft ->
-                        buildActions(oldAttributeDefinition, attributeDefinitionDraft)
-                    )
+                    .map(attributeDefinitionDraft -> {
+                        // attribute type is required so if null we let commercetools to throw exception
```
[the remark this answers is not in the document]
jortizsao: This comment means that even knowing that the product type is invalid (because the attribute type is required for an attribute) we don't throw any exception. 

We let commercetools to throw the exception. This is a decision we made because the commercetools documentation states that the attribute type is required, and we decided not throw library validations for those cases that are documented in commercetools.

In this case the validation is just for checking if there is attribute type to avoid a null pointer exception in the `haveSameAttributeType`.

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
```diff
@@ -0,0 +1,827 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class ProductTypeSyncIT {
+
+    /**
+     * Deletes inventories and supply channels from source and target CTP projects.
```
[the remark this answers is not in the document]
jortizsao: 4cbc765204ca1b1608efcacff6579d4e79a39548

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,298 @@
...
+
+    private final ProductTypeService productTypeService;
+
+    /**
+     * Takes a {@link ProductTypeSyncOptions} instance to instantiate a new {@link ProductTypeSync} instance that
+     * could be used to sync product type drafts with the given product types in the CTP project specified in the
+     * injected {@link ProductTypeSyncOptions} instance.
+     *
+     * @param productTypeSyncOptions the container of all the options of the sync process including the CTP project
+     *                               client and/ormconfiguration and other sync-specific options.
```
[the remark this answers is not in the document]
jortizsao: I have removed the javadoc from the constructors and add it to the class, so this typo doesn't exist anymore.

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,298 @@
...
+    private static final String CTP_PRODUCT_TYPE_CREATE_FAILED = "Failed to create product type of key '%s'.";
+    private static final String PRODUCT_TYPE_DRAFT_HAS_NO_KEY = "Failed to process product type draft without key.";
+    private static final String PRODUCT_TYPE_DRAFT_IS_NULL = "Failed to process null product type draft.";
+
+    private final ProductTypeService productTypeService;
+
+    /**
+     * Takes a {@link ProductTypeSyncOptions} instance to instantiate a new {@link ProductTypeSync} instance that
+     * could be used to sync product type drafts with the given product types in the CTP project specified in the
+     * injected {@link ProductTypeSyncOptions} instance.
```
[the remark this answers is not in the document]
jortizsao: 24722ff2291e9732efaf750eef70669095766dcd

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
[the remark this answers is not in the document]
jortizsao: f04583df9ba11b02246dd38bc3664740e07d6b0d

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,298 @@
...
+        return false;
+    }
+
+    /**
+     * Given a set of product type keys, fetches the corresponding product types from CTP if they exist.
+     *
+     * @param keys the keys of the product types that are wanted to be fetched.
+     * @return a future which contains the list of product types corresponding to the keys.
+     */
+    private CompletionStage<Optional<List<ProductType>>> fetchExistingProductTypes(@Nonnull final Set<String> keys) {
```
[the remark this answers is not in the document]
jortizsao: 350a08f37404f2ec09337a61a75ef5124c6d8383

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,298 @@
...
+            return completedFuture(statistics);
+        } else {
+            final Map<String, ProductTypeDraft> keysProductTypeDraftMap = getKeysProductTypeMap(validProductTypeDrafts);
+
+
+            return fetchExistingProductTypes(keysProductTypeDraftMap.keySet())
+                .thenCompose(oldProductTypeOptional -> oldProductTypeOptional
+                    .map(oldProductTypes -> syncBatch(oldProductTypes, validProductTypeDrafts))
+                    .orElseGet(() -> completedFuture(statistics)))
+                .thenApply(v -> {
```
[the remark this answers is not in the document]
jortizsao: f894dbde8feae3bc6c1adc6b1ba936ae6bdf885c

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,298 @@
...
+    /**
+     * Given a {@link String} {@code errorMessage} and a {@link Throwable} {@code exception}, this method calls the
+     * optional error callback specified in the {@code syncOptions} and updates the {@code statistics} instance by
+     * incrementing the total number of failed product types to sync.
+     *
+     * @param errorMessage The error message describing the reason(s) of failure.
+     * @param exception    The exception that called caused the failure, if any.
+     */
+    private void handleError(@Nonnull final String errorMessage, @Nullable final Throwable exception,
+                             final int failedTimes) {
```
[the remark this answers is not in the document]
jortizsao: a406d30e0bf818804aa092ad8a5af4689b0f801b

src/main/java/com/commercetools/sync/producttypes/utils/ProductTypeSyncUtils.java
```diff
@@ -0,0 +1,56 @@
...
+import static java.util.stream.Collectors.toList;
+
+public final class ProductTypeSyncUtils {
+
+    /**
+     * Compares all the fields (including the attributes see
+     * {@link ProductTypeUpdateActionUtils#buildAttributesUpdateActions}) of a {@link ProductType} and a
+     * {@link ProductTypeDraft}. It returns a {@link List} of {@link UpdateAction}&lt;{@link ProductType}&gt; as a
+     * result. If no update action is needed, for example in case where both the {@link ProductType} and the
+     * {@link ProductTypeDraft} have the same description, an empty {@link List} is returned.
```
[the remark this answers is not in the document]
jortizsao: 7897d8478b5a1d76f8fc8674e8fc12e78e773f65

src/test/java/com/commercetools/sync/producttypes/helpers/ProductTypeSyncStatisticsTest.java
jortizsao: e93cbf8102c6a2375f8ce8f775afbf9c591f388b

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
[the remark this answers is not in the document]
lojzatran: What I meant is this:
```
final List<ProductTypeDraft> validProductTypeDrafts = productTypeDrafts.stream()
                                                                   .filter(this::validateDraft)
                                                                   .collect(toList());

        final List<List<ProductTypeDraft>> batches = batchElements(validProductTypeDrafts, syncOptions.getBatchSize());
```

and remove `validateDraft` part from `processBatch()` method.

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,279 @@
...
+                        return createProductType(newProductType);
+                    } else {
+                        return updateProductType(oldProductType, newProductType);
+                    }
+                })
+                .map(CompletionStage::toCompletableFuture)
+                .collect(toList());
+
+        return CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]))
+                .thenApply(result -> statistics);
```
heshamMassoud: what do you think if we can refactor this to return in one step: 

````java
    private CompletionStage<ProductTypeSyncStatistics> syncBatch(
            @Nonnull final List<ProductType> oldProductTypes,
            @Nonnull final List<ProductTypeDraft> newProductTypes) {

        final Map<String, ProductType> oldProductTypeMap = getKeysProductTypeMap(oldProductTypes);

        return CompletableFuture.allOf(newProductTypes
            .stream()
            .map(newProductType -> {
                final ProductType oldProductType = oldProductTypeMap.get(newProductType.getKey());

                return ofNullable(oldProductType)
                    .map(productType -> updateProductType(oldProductType, newProductType))
                    .orElseGet(() -> createProductType(newProductType));
            })
            .map(CompletionStage::toCompletableFuture)
            .toArray(CompletableFuture[]::new)).thenApply(result -> statistics);
    }
````

File: src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java [hunk 27]
```diff
@@ -0,0 +1,279 @@
...
+     * {@link ProductTypeSync#syncBatch(List, List)} function. When fetching existing product types results in
+     * an empty optional then {@code batch} isn't processed.
+     *
+     * @param batch batch of drafts that need to be synced
+     * @return {@link CompletionStage} of {@link Void} that indicates method progress.
+     */
+    @Override
+    protected CompletionStage<ProductTypeSyncStatistics> processBatch(@Nonnull final List<ProductTypeDraft> batch) {
+        final List<ProductTypeDraft> validProductTypeDrafts = batch.stream()
+                .filter(this::validateDraft)
```
heshamMassoud: please always indent the . chains on the same line, so it's easier to read:

````java
        final List<ProductTypeDraft> validProductTypeDrafts = batch.stream()
                                                                   .filter(this::validateDraft)
                                                                   .collect(toList());
````

File: src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java [hunk 28]
```diff
@@ -0,0 +1,279 @@
...
+
+        if (validProductTypeDrafts.isEmpty()) {
+            statistics.incrementProcessed(batch.size());
+            return completedFuture(statistics);
+        } else {
+            final Set<String> keys = validProductTypeDrafts.stream().map(ProductTypeDraft::getKey).collect(toSet());
+
+            return fetchExistingProductTypes(keys)
+                .thenCompose(oldProductTypes -> syncBatch(oldProductTypes, validProductTypeDrafts))
+                .thenApply((ignored) -> {
```
heshamMassoud: the parantheses around the `ignored` are not needed ;)

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
```diff
@@ -0,0 +1,827 @@
...
+        AssertionsForStatistics.assertThat(productTypeSyncStatistics).hasValues(1, 0, 1, 0);
+
+
+        final Optional<ProductType> oldProductTypeAfter = getProductTypeByKey(CTP_TARGET_CLIENT, PRODUCT_TYPE_KEY_1);
+
+        assertThat(oldProductTypeAfter).isNotEmpty();
+        assertThat(oldProductTypeAfter.get().getName()).isEqualTo(PRODUCT_TYPE_NAME_2);
+        assertThat(oldProductTypeAfter.get().getDescription()).isEqualTo(PRODUCT_TYPE_DESCRIPTION_2);
+        assertAttributesAreEqual(oldProductTypeAfter.get().getAttributes(),
+            singletonList(ATTRIBUTE_DEFINITION_DRAFT_1));
```
heshamMassoud: instead of those 4 asserts and checking that it is not empty explicitly and doing `.get()` after. You can do it like this:

````java
assertThat(oldProductTypeAfter).hasValueSatisfying(productType -> {
            assertThat(productType.getName()).isEqualTo(PRODUCT_TYPE_NAME_2);
            assertThat(productType.getDescription()).isEqualTo(PRODUCT_TYPE_DESCRIPTION_2);
            assertAttributesAreEqual(productType.getAttributes(), singletonList(ATTRIBUTE_DEFINITION_DRAFT_1));
        });
````

Comment (heshamMassoud) [reply src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java]: you can use it across all asserts of this class actually ;)

File: src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java [hunk 30]
```diff
@@ -0,0 +1,827 @@
...
+
+        final ProductTypeSync productTypeSync = new ProductTypeSync(spyProductTypeSyncOptions);
+
+        productTypeSync.sync(singletonList(newProductTypeDraft)).toCompletableFuture().join();
+
+        verify(spyProductTypeSyncOptions, never()).applyBeforeUpdateCallBack(any(), any(), any());
+    }
+
+    @Test
+    public void sync_beforeUpdate_ShouldCallBeforeUpdateCallback() {
```
heshamMassoud: I am wondering why are these tests significant? Namely: 

- public void sync_beforeCreate_ShouldCallBeforeCreateCallback
- public void sync_beforeCreate_ShouldNotCallBeforeUpdateCallback
- public void sync_beforeUpdate_ShouldCallBeforeUpdateCallback
- public void sync_beforeUpdate_ShouldNotCallBeforeCreateCallback

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
```diff
@@ -0,0 +1,827 @@
...
+    @Test
+    public void sync_WithSeveralBatches_ShouldReturnProperStatistics() {
+        // Default batch size is 50 (check ProductTypeSyncOptionsBuilder) so we have 2 batches of 50
+        final List<ProductTypeDraft> productTypeDrafts = IntStream
+            .range(0, 100)
+            .mapToObj(i -> ProductTypeDraft.ofAttributeDefinitionDrafts(
+                "product_type_key_" + Integer.toString(i),
+                "product_type_name_" + Integer.toString(i),
+                "product_type_description_" + Integer.toString(i),
+                Arrays.asList(ATTRIBUTE_DEFINITION_DRAFT_1)
```
heshamMassoud: you can use here instead: `singletonList(ATTRIBUTE_DEFINITION_DRAFT_1)`

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
```diff
@@ -0,0 +1,827 @@
...
+        AssertionsForStatistics.assertThat(productTypeSyncStatistics).hasValues(1, 0, 0, 1);
+    }
+
+    @Test
+    public void sync_WithErrorUpdatingTheProductTypeInCT_ShouldExecuteCallbackOnErrorAndIncreaseFailedCounter() {
+        final ProductTypeDraft newProductTypeDraft = ProductTypeDraft.ofAttributeDefinitionDrafts(
+            PRODUCT_TYPE_KEY_1,
+            PRODUCT_TYPE_NAME_1,
+            PRODUCT_TYPE_DESCRIPTION_1,
+            Arrays.asList(ATTRIBUTE_DEFINITION_DRAFT_1)
```
heshamMassoud: singletonList(ATTRIBUTE_DEFINITION_DRAFT_1)

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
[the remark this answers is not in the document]
jortizsao: mmm... but if I validate there, then the counter for "processed" items is not increased in `processBatch` because the invalid drafts are removed before and `processBatch` is responsible for increasing the counter of processed items in the statistics. 

In the other synchronization of entities, the invalid drafts are considered as processed. Furthermore, in the `processBatch` the invalid drafts are filtered before processing, so it seems that there is no performance advantage of moving the filtering to the process.

Maybe I'm wrong, what do you think @heshamMassoud ?

docs/usage/PRODUCT_TYPE_SYNC.md
```diff
@@ -0,0 +1,118 @@
...
+ the update action. One example is the `buildChangeNameUpdateAction` which compares names:
+````java
+Optional<UpdateAction<ProductType>> updateAction = ProductTypeUpdateActionUtils.buildChangeNameAction(oldProductType, productTypeDraft);
+````
+More examples of those utils for different fields can be found [here](/src/test/java/com/commercetools/sync/producttypes/utils/ProductTypeUpdateActionUtilsTest.java).
+
+
+## Caveats
+
+1. Product types are either created or updated. Currently the tool does not support product type deletion.
```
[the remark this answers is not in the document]
jortizsao: you are right. I add it. thanks!

README.md
```diff
@@ -43,6 +43,7 @@ Currently this library supports synchronising
  - [Categories](/docs/usage/CATEGORY_SYNC.md)
  - [Products](/docs/usage/PRODUCT_SYNC.md) (_Beta_: Not recommended for production use yet.)
  - [InventoryEntries](/docs/usage/INVENTORY_SYNC.md) (_Beta_: Not recommended for production use yet.)
+ - [ProductTypes](/docs/usage/PRODUCT_TYPE_SYNC.md) (_Beta_: Not recommended for production use yet.)
```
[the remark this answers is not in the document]
jortizsao: we can keep it in beta or remove it (the ball is in your court xD). The `Nested`type is in Beta in commercetools, so keep the product type sync in Beta just for that reason maybe is not enough.

I put it in Beta because even after already tested it and so on... I thought it was wise to set it as Beta just for some time until we have confidence that it actually works as the customers expect.

WDYT?

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
jortizsao: :) I change it but my understanding when we discussed about was that the second chained function and successive could be indented in the beginning on the next line, because otherwise the code is moving to the right each time we have a chaining function with body, until you finally reach the right gutter where you don't have enough space (i.e https://github.com/commercetools/commercetools-sync-java/blob/master/src/main/java/com/commercetools/sync/categories/CategorySync.java#L175 , the arrow functions don't fit in one line making the code tough to read because the chain of function is almost in the right end)

And I completely agree that the `.` should be aligned but my understanding was from the second chained function onwards :)

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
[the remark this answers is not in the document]
heshamMassoud: No, you are arguments are absolutely correct @jortizsao.

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
jortizsao: Basically is to verify that the right callbacks are called in the correct stages. 

i.e How would you verify that `applyBeforeUpdateCallBack` function is called when a product type is created?

README.md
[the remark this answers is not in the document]
heshamMassoud: > quote 

I put it in Beta because even after already tested it and so on... I thought it was wise to set it as Beta just for some time until we have confidence that it actually works as the customers expect.

src/integration-test/java/com/commercetools/sync/integration/producttypes/utils/ProductTypeITUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Optional;
+
+import static com.commercetools.sync.integration.commons.utils.ITUtils.queryAndExecute;
+import static com.commercetools.sync.integration.commons.utils.SphereClientUtils.CTP_SOURCE_CLIENT;
+import static com.commercetools.sync.integration.commons.utils.SphereClientUtils.CTP_TARGET_CLIENT;
+
+public class ProductTypeITUtils {
```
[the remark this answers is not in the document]
jortizsao: eb319912d29145bfcd6e44a284dacd03d818f678

docs/usage/PRODUCT_TYPE_SYNC.md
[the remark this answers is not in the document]
jortizsao: 02d0eff84d20b2cde1268e254765278922b35377

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * Given a list of {@link ProductType} or {@link ProductTypeDraft}, returns a map of keys to the
+     * {@link ProductType}/{@link ProductTypeDraft} instances.
+     *
+     * @param productTypes list of {@link ProductType}/{@link ProductTypeDraft}
+     * @param <T>          a type that extends of {@link WithKey}.
+     * @return the map of keys to {@link ProductType}/{@link ProductTypeDraft} instances.
+     */
+    protected <T extends WithKey> Map<String, T> getKeysProductTypeMap(@Nonnull final List<T> productTypes) {
```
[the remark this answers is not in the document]
jortizsao: 1602afffabb64340e94fb91389c41a44b3137f1e

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
jortizsao: a3dc4dd7cd01e75ed24e2ae0a02cd125c084b967

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
jortizsao: 959790a37dcf9d23270261d97dc20f663d44f01d

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
jortizsao: cbc9b9699ecf191fe8a806e9df7c75ff8cc0afa8

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
jortizsao: 601ffb1e1af54363e7d5a52b771054f5d7d1a900

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
jortizsao: 601ffb1e1af54363e7d5a52b771054f5d7d1a900

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
jortizsao: ed60b406dbe2396d2956624e7636e3d7e26a22ec

src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java
heshamMassoud: but do you think we really need an integration test for that, isn't a unit test sufficient to test that behaviour?

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
heshamMassoud: no, then we had a misunderstanding 😁 What I meant is if it's too hard to read like the example you sent, then start the first chain from the beginning of the next line, like:


````java

return categoryService
.cacheKeysToIds()
.thenCompose(keyToIdCache -> {
  prepareDraftsForProcessing(categoryDrafts, keyToIdCache);
  categoryKeysToFetch = existingCategoryDrafts.stream().map(CategoryDraft::getKey)
  ...
````

and in this case:

````java
final List<ProductTypeDraft> validProductTypeDrafts = batch
.stream()
.filter(this::validateDraft)
.collect(toList());
````

but honestly, in that case I think it's still possible to have the chains on the same indentation while keeping the first chain on the same line (like I mentioned in the comment before):

````java
final List<ProductTypeDraft> validProductTypeDrafts = batch.stream()
                                                           .filter(this::validateDraft)
                                                           .collect(toList());
````

WDYT?

**Comment by heshamMassoud** [comment]
@jortizsao We are also missing benchmarks for the product Type Sync. I'll create it as an issue for now.

**Comment by heshamMassoud** [comment]
release notes entries are also missing. I will add them.

File: src/integration-test/java/com/commercetools/sync/integration/producttypes/ProductTypeSyncIT.java [hunk 37]
```diff
@@ -0,0 +1,828 @@
...
+            PRODUCT_TYPE_NAME_2,
+            PRODUCT_TYPE_DESCRIPTION_2,
+            singletonList(ATTRIBUTE_DEFINITION_DRAFT_1)
+        );
+
+        final ProductTypeSyncOptions productTypeSyncOptions = ProductTypeSyncOptionsBuilder
+            .of(CTP_TARGET_CLIENT)
+            .build();
+
+        ProductTypeSyncOptions spyProductTypeSyncOptions = spy(productTypeSyncOptions);
```
heshamMassoud: `final`?

PR #297, opened 2018-08-03
101 customer group reference resolution

heshamMassoud:
#### Summary
Add reference resolution for customer group reference on prices.

#### Relevant Issues
#101 #297 #290 

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [x] Documentation

#### Hints for Review

1. Remove `BaseReferenceResolver#getKeyFromExpansionOrReference` as we won't do reference resolution this way anymore. But rather only from the `id` field for `reference`s or `key` field on `resourceIdentifier`.

2. Use `getKeyFromResourceIdentifier` to fetch the key channel reference resolution in `InventoryReferenceResolver`.

3. Add `InventoryReferenceReplacementUtils#replaceChannelReferenceIdWithKey` for replacing channel reference keys with ids.

4. Support `CustomerGroup` reference resolution on variant prices:
     - Add `CustomerGroup` service and implementation.
     - Inject `CustomerGroupService` in `ProductSync`, `ProductReferenceResolver`, `VariantReferenceResolver ` and `PriceReferenceResolver `.
     - Add unit tests and integration tests


src/test/java/com/commercetools/sync/inventories/utils/InventoryReferenceReplacementUtilsTest.java
```diff
@@ -55,24 +65,74 @@
         replaceInventoriesReferenceIdsWithKeys_WithNonExpandedReferences_ShouldReturnReferencesWithoutReplacedKeys() {
         final String customTypeId = UUID.randomUUID().toString();
         final List<InventoryEntry> mockInventoryEntries = new ArrayList<>();
+
```
lojzatran: preparation/test/assertion phases missing.

src/test/java/com/commercetools/sync/products/helpers/PriceReferenceResolverTest.java
```diff
@@ -199,14 +208,90 @@ public void resolveChannelReference_WithChannelKeyAsUuidSetAndAllowed_ShouldReso
...
+        final PriceReferenceResolver priceReferenceResolver =
+            new PriceReferenceResolver(syncOptions, typeService, channelService, customerGroupService);
+
+
+        priceReferenceResolver.resolveChannelReference(priceBuilder)
+                              .exceptionally(exception -> {
+                                  assertThat(exception).isExactlyInstanceOf(CompletionException.class);
+                                  assertThat(exception.getCause())
+                                      .isExactlyInstanceOf(ReferenceResolutionException.class);
+                                  assertThat(exception.getCause().getCause())
```
lojzatran: Why is it wrapped twice?

src/test/java/com/commercetools/sync/inventories/utils/InventoryReferenceReplacementUtilsTest.java
heshamMassoud: yep, it's actually something new that I am adding to the newly added tests because I think it makes reading the tests much more easier. But I cannot added to all the tests that were already existing.

src/main/java/com/commercetools/sync/inventories/utils/InventoryReferenceReplacementUtils.java
```diff
@@ -1,42 +1,71 @@
...
      */
     @Nonnull
-    public static List<InventoryEntryDraft> replaceInventoriesReferenceIdsWithKeys(@Nonnull final List<InventoryEntry>
-                                                                                       inventoryEntries) {
+    public static List<InventoryEntryDraft> replaceInventoriesReferenceIdsWithKeys(
+        @Nonnull final List<InventoryEntry> inventoryEntries) {
         return inventoryEntries
             .stream()
             .map(inventoryEntry -> {
                 final CustomFieldsDraft customTypeWithKeysInReference = replaceCustomTypeIdWithKeys(inventoryEntry);
```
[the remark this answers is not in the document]
heshamMassoud: Why would we need a separate class for just the method? I think it doesn't deserve it's own class yet. The SyncUtils class has these utils which are commonly needed by different places.

src/test/java/com/commercetools/sync/products/helpers/PriceReferenceResolverTest.java
heshamMassoud: @lojzatran because there are two causes for the exception:
1. Reference resolution exception "Failed to resolve the `channel` reference on the PriceDraft..."
2. Channel with key 'channelKey' doesn't exits.

src/integration-test/java/com/commercetools/sync/integration/services/CustomerGroupServiceIT.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static com.commercetools.sync.integration.commons.utils.CustomerGroupITUtils.createCustomerGroup;
+import static com.commercetools.sync.integration.commons.utils.CustomerGroupITUtils.deleteCustomerGroups;
+import static com.commercetools.sync.integration.commons.utils.SphereClientUtils.CTP_TARGET_CLIENT;
+import static com.commercetools.sync.services.impl.CustomerGroupServiceImpl.CUSTOMER_GROUP_KEY_NOT_SET;
+import static com.commercetools.tests.utils.CompletionStageUtil.executeBlocking;
+import static java.lang.String.format;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CustomerGroupServiceIT {
```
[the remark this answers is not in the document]
lojzatran: True. But what I tried to say is you're actually testing `CustomerGroupServiceImpl` class, not the interface itself, so it should be named accordingly `CustomerGroupServiceImplIT` + better to put it in the same package. However, it's not a big deal.

src/main/java/com/commercetools/sync/inventories/utils/InventoryReferenceReplacementUtils.java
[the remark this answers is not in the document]
lojzatran: But your utils are for only one method, e.g. InventorySyncUtils.java. And it's also used at one place :) I thought you have the utils classes, so the users of this lib can use more granular methods to construct their own behaviour if needed?

src/main/java/com/commercetools/sync/services/impl/CustomerGroupServiceImpl.java
```diff
@@ -0,0 +1,64 @@
...
+            return CompletableFuture.completedFuture(Optional.empty());
+        }
+        if (keyToIdCache.isEmpty()) {
+            return cacheAndFetch(key);
+        }
+        return CompletableFuture.completedFuture(ofNullable(keyToIdCache.get(key)));
+    }
+
+    @Nonnull
+    private CompletionStage<Optional<String>> cacheAndFetch(@Nonnull final String key) {
```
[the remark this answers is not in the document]
heshamMassoud: 5eaf142

src/test/java/com/commercetools/sync/inventories/utils/InventoryReferenceReplacementUtilsTest.java
heshamMassoud: @lojzatran 5802457

src/integration-test/java/com/commercetools/sync/integration/services/CustomerGroupServiceIT.java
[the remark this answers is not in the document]
heshamMassoud: agree with naming change. Will do it.

src/main/java/com/commercetools/sync/inventories/utils/InventoryReferenceReplacementUtils.java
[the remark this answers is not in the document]
heshamMassoud: True. I agree.

src/integration-test/java/com/commercetools/sync/integration/services/CustomerGroupServiceIT.java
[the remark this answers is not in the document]
heshamMassoud: @lojzatran 
1822eac

src/main/java/com/commercetools/sync/inventories/utils/InventoryReferenceReplacementUtils.java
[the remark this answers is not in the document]
heshamMassoud: @lojzatran 
d73463d

src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.annotation.Nonnull;
+import java.util.List;
+
+import static com.commercetools.sync.commons.utils.CustomTypeReferenceReplacementUtils.replaceCustomTypeIdWithKeys;
+import static java.util.stream.Collectors.toList;
+
+public final class AssetReferenceReplacementUtils {
+
+    /**
+     * Takes a an asset list that is supposed to have all its assets' custom references expanded in order to be able to
```
LEQADA: `Takes a an asset` -> `Takes an asset `

src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
```diff
@@ -0,0 +1,36 @@
...
+     * fetch the keys and replace the reference ids with the corresponding keys for the custom references. This method
+     * returns as a result a {@link List} of {@link AssetDraft} that has all custom references with keys replacing the
+     * ids.
+     *
+     * <p>Any custom reference that is not expanded will have it's id in place and not replaced by the key.
```
LEQADA: `will have it's id` -> `will have its id `

src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @param assets the list of assets to replace their custom ids with keys.
+     * @return a {@link List} of {@link AssetDraft} that has all channel references with keys replacing the ids.
+     */
+    @Nonnull
+    public static List<AssetDraft> replaceAssetsReferencesIdsWithKeys(@Nonnull final List<Asset> assets) {
```
LEQADA: `replaceAssetsReferencesIdsWithKeys` -> `replaceAssetReferenceIdsWithKeys`? Hm? WDYT?

src/main/java/com/commercetools/sync/commons/utils/CustomTypeReferenceReplacementUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public final class CustomTypeReferenceReplacementUtils {
+    /**
+     * Given a resource of type {@code T} that extends {@link Custom} (i.e. it has {@link CustomFields}, this method
+     * checks if the custom fields are existing (not null) and they are reference expanded. If they are then
+     * it returns a {@link CustomFieldsDraft} instance with the custom type key in place of the id of the reference.
+     * Otherwise, if its not reference expanded it returns a {@link CustomFieldsDraft} with the id not replaced. If the
```
LEQADA: `if its not` -> `if it's not `

src/main/java/com/commercetools/sync/commons/utils/CustomTypeReferenceReplacementUtils.java
```diff
@@ -0,0 +1,41 @@
...
+     * resource has no or null {@link Custom}, then it returns {@code null}.
```
LEQADA: If a resource doesn't have a `Custom` it means the Custom is `null`. So maybe:
`If the resource has no or null {@link Custom}` -> `If {@link Custom} of the resource is null`?

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
```diff
@@ -68,40 +65,12 @@ protected BaseReferenceResolver(@Nonnull final S options) {
     protected static String getKeyFromResourceIdentifier(@Nonnull final ResourceIdentifier resourceIdentifier,
                                                          final boolean allowUuidKeys)
         throws ReferenceResolutionException {
+
```
LEQADA: We get ID from the `resourceIdentifier` and name it as a `key`. However there is a `key` field on the `resourceIdentifier` too which makes everything confusing. Could you add a comment explaining why it's used this way?

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
```diff
@@ -68,40 +65,12 @@ protected BaseReferenceResolver(@Nonnull final S options) {
...
         final String key = resourceIdentifier.getId();
```
LEQADA: We get ID from the resourceIdentifier and name it as a key. However there is a key field on the resourceIdentifier too which makes everything confusing. Could you add a comment explaining why it's used this way?
Sorry for out of scope comment 🐳

src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
heshamMassoud: I know it sounds more correct, however, in that case reference*s* is correct because the method is supposed to replace all the ids of  "references" in an asset not just a reference.

src/main/java/com/commercetools/sync/services/CustomerGroupService.java
```diff
@@ -0,0 +1,30 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+public interface CustomerGroupService {
+    /**
+     * Given a {@code key}, this method first checks if a cached map of customer group keys -&gt; ids is not empty.
+     * If not, it returns a completed future that contains an optional that contains what this key maps to in
```
LEQADA: {@link Optional}

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
heshamMassoud: a099c2d

src/main/java/com/commercetools/sync/services/CustomerGroupService.java
heshamMassoud: https://github.com/commercetools/commercetools-sync-java/pull/297/commits/16290ad05156b3d5e0b45ad47ad00d7737e451c2

src/main/java/com/commercetools/sync/commons/utils/CustomTypeReferenceReplacementUtils.java
heshamMassoud: https://github.com/commercetools/commercetools-sync-java/pull/297/commits/7dcb9c6314c51346c2a59bb73feb70f44043e8b4

src/main/java/com/commercetools/sync/services/CustomerGroupService.java
```diff
@@ -0,0 +1,30 @@
...
+     * key.
+     *
+     * @param key the key by which a {@link CustomerGroup} id should be fetched from the CTP project.
+     * @return {@link CompletionStage}&lt;{@link Optional}&lt;{@link String}&gt;&gt; in which the result of its
+     *         completion could contain an {@link Optional} with the id inside of it or an empty {@link Optional} if no
+     *         {@link CustomerGroup} was found in the CTP project with this key or a blank (null/empty string) key was
+     *         supplied.
+     */
+    @Nonnull
+    CompletionStage<Optional<String>> fetchCachedCustomerGroupId(@Nullable final String key);
```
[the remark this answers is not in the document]
heshamMassoud: It's only used there. But this is a service and maybe in the future some one would call it like (e.g. when we have a CustomerGroup sync). The service should be flexible and free the user from having to null check the passed key. So the service can accept null and it will handle it insider ;)

Issue #171, opened 2017-11-02
Use consistent naming in Tests and ITs setup and tearing down.


heshamMassoud:
Use consistent naming in Tests and ITs setup and tearing down: 

@Before usually is called setup()/setUp()
@After - teadDown()
@BeforeClass - setupClass()/beforeClass()
@AfterClass - tearDownClass()/afterClass()

heshamMassoud:
I moved it to M15 since you started working on it.

ahmetoz:
That issue is not looking important to me, imo it's ok-ish to have different names, it is enough that test hook names are easy to understand. We don't need a standard here.

PR #320, opened 2018-10-11
WIP #171 - refactor the test method names

ahmetoz:
#### Summary
Refactor test method names based on issue #171


src/benchmark/java/com/commercetools/sync/benchmark/ProductSyncBenchmark.java
```diff
@@ -61,15 +61,15 @@
...
+    public static void setupClass() {
         deleteProductSyncTestData(CTP_TARGET_CLIENT);
         createCategoriesCustomType(OLD_CATEGORY_CUSTOM_TYPE_KEY, ENGLISH,
             OLD_CATEGORY_CUSTOM_TYPE_NAME, CTP_TARGET_CLIENT);
         productType = createProductType(PRODUCT_TYPE_RESOURCE_PATH, CTP_TARGET_CLIENT);
     }
 
     @Before
-    public void setupTest() {
+    public void setup() {
```
heshamMassoud: what do you think about the name `setupEach`? do you think it's maybe more self expressive that means its before each test?

src/test/java/com/commercetools/sync/products/utils/productvariantupdateactionutils/prices/BuildProductVariantPricesUpdateActionsTest.java
```diff
@@ -21,8 +21,53 @@
 import java.util.List;
 
 import static com.commercetools.sync.products.utils.ProductVariantUpdateActionUtils.buildProductVariantPricesUpdateActions;
-import static com.commercetools.sync.products.utils.productvariantupdateactionutils.prices.PriceDraftFixtures.*;
-import static com.commercetools.sync.products.utils.productvariantupdateactionutils.prices.PriceFixtures.*;
+import static com.commercetools.sync.products.utils.productvariantupdateactionutils.prices.PriceDraftFixtures.DRAFT_DE_100_EUR_01_02_CHANNEL1_CUSTOMTYPE1_CUSTOMFIELDX;
```
heshamMassoud: Can we please leave these imports to be imported with a wild card? It was done on purpose here due to the huge amount of constants from the same class.

heshamMassoud:
There is a missed refactoring in:`BuildProductVariantAttributesUpdateActionsTest#setupMethod`

src/test/java/com/commercetools/sync/products/utils/productvariantupdateactionutils/prices/BuildProductVariantPricesUpdateActionsTest.java
ahmetoz: 😅  IntelliJ Idea changed it probably, Ok, I will revert.

src/benchmark/java/com/commercetools/sync/benchmark/ProductSyncBenchmark.java
heshamMassoud: I like this idea! Maybe it should be more self explaining naming what is it actually doing. And we leave the explanation of setup each or class to be explained by the annotation alone.

ahmetoz: Yes, The method names are for developers to understand without looking at the body of the method. The annotations are for that to make it meaningfully about the context and design. So, in this case, we are just naming it twice (btw it is nice to have the same convention if you don't think about naming these things). But probably it will take time to rename all the methods like everyone heard sth like this quote 😄 

> There are only two hard things in Computer Science: cache invalidation and naming things.
-- Phil Karlton

heshamMassoud: Take more time yes. But definitely better :)

src/test/java/com/commercetools/sync/products/utils/productvariantupdateactionutils/prices/BuildProductVariantPricesUpdateActionsTest.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/320/commits/395547b2d36451f6ce8a2cbfee3070f440bfba5c

PR #322, opened 2018-10-14
Bump dependencies

heshamMassoud:
#### Summary
- Bump JVM SDK dependency to 1.35.0 (apply changes after dep bump)
- Bump mockito dependency to 2.23.0

#### Relevant Issues
https://github.com/commercetools/commercetools-jvm-sdk/issues/1786

### DIFF
--- a/src/test/java/com/commercetools/sync/producttypes/utils/producttypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java
+++ b/src/test/java/com/commercetools/sync/producttypes/utils/producttypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java
@@ -8,6 +8,8 @@ import io.sphere.sdk.client.SphereClient;
 import io.sphere.sdk.commands.UpdateAction;
 import io.sphere.sdk.products.attributes.AttributeDefinition;
 import io.sphere.sdk.products.attributes.AttributeDefinitionBuilder;
+import io.sphere.sdk.products.attributes.AttributeDefinitionDraft;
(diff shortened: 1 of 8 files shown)


src/test/java/com/commercetools/sync/producttypes/utils/producttypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java
```diff
@@ -231,7 +232,11 @@ public void buildAttributesUpdateActions_WithOneExtraAttribute_ShouldBuildAddAtt
         );
 
         assertThat(updateActions).containsExactly(
-            AddAttributeDefinition.of(ATTRIBUTE_DEFINITION_D)
+            AddAttributeDefinition.of(AttributeDefinitionDraftBuilder
```
ahmetoz: Is there a difference or you just checking the factory method is working?

heshamMassoud: because of 2 things:

1. The previously used `AddAttributeDefinition#of(AttributeDefinition)` is now deprecated.
2. The created `ATTRIBUTE_DEFINITION_D` gets by default set values for `attributeConstraint`, `required` and `searchable` fields. Which is not the case for when creating an `attributeDefinitionDraft`. And the action itself is built from a draft now after a change as opposed to from a AttributeDefinition like it was before.

PR #328, opened 2018-10-30
apply empty completableUtil

whitemalt:
#### Summary
apply empty completable util

#### Description
change all CompletableFuture.completedFuture(Optional.empty()) to use util function instead

#### Relevant Issues
[issues 164](https://github.com/commercetools/commercetools-sync-java/issues/164)

#### Todo

- Tests
    - [ ] Unit 
    - [/ ] Integration
- [ ] Documentation

#### Hints for Review
CompletableFuture.completedFuture(Optional.empty()) replace with emptyOptionalCompletedFuture() function

### DIFF
--- a/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java
@@ -4,6 +4,7 @@ import javax.annotation.Nonnull;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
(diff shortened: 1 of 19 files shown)


whitemalt:
Yes, I'm agree with you.

It may useful for easier tracking when want to change an empty CompletableFuture to something later, For example return a default value. 

Single point change to change all ,that is what util function propose for me.

heshamMassoud:
@whitemalt but then it will be a totally different util in that case (with even a different name, not: `emptyOptionalCompletedFuture`). I guess it would be also obfuscating what is the default value in that case, since every caller would have to track down the util to check what is it actually returning.

So I guess for now it's not needed very much. Hence, I would close the issue and the PR. 

Thanks again for your PR! Looking forward for more contributions.

heshamMassoud:
Having it as constant variable could indeed lead to **micro**improvements in regards to memory and time performance. Even though it's a negligible improvement as compared to the IO latencies of the application of the library. So It's a question of whether it is readable enough or not. IMO, it's not very unreadable/obfuscated with the new util.

With that said, it would be great if you could cover the new util with unit tests covering the different cases :)

heshamMassoud:
Closing this PR due to being stale for a long time.

PR #350, opened 2018-12-07
Prepare 1.0.0

heshamMassoud:
#### Summary
- Update to new versions in md files.
- Add release notes summary. https://github.com/commercetools/commercetools-sync-java/blob/Prepare-1.0.0/docs/RELEASE_NOTES.md#v100----dec-10-2018

### DIFF
--- a/docs/RELEASE_NOTES.md
+++ b/docs/RELEASE_NOTES.md
@@ -41,7 +41,18 @@
 [Javadoc](https://commercetools.github.io/commercetools-sync-java/v/v1.0.0/) | 
 [Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/v1.0.0)
 
+##### The Beta is Over 🎉
 
+We're happy to announce that the commercetools-sync-java is finally out of beta! Big thanks to all the bold users 
+who were using it when it was still in beta. Your feedback was definitely valuable for us to reach the current state of 
+the library. `1.0.0` is here for you to use with all new features, enhancements and bug fixes including:
+
+- The library now supports importing/syncing [`types`](https://docs.commercetools.com/http-api-projects-types.html) into a CTP project from an external feed or another CTP project. [Read more](https://commercetools.github.io/commercetools-sync-java/doc/usage/TYPE_SYNC/).
+- The library now handles concurrency modification exceptions for the `productType` sync.
+- All new documentation pages including a [quick start guide](https://commercetools.github.io/commercetools-sync-java/doc/usage/QUICK_START/).
+- Many more improvements and bug fixes. 
+
+##### Full Release Notes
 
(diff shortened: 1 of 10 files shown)


docs/RELEASE_NOTES.md
```diff
@@ -9,9 +9,9 @@
 3. link to Javadoc of release.
 4. link to Jar of release.
 
-4. Add a summary of the release that is not too detailed or technical.
+5. Add a summary of the release that is not too detailed or technical.
 
-5. Depending on the contents of the release use the subtitles below to 
+6. Depending on the contents of the release use the subtitles below to
```
butenkor: subtitel -> subitem

docs/RELEASE_NOTES.md
```diff
@@ -77,19 +90,21 @@
...
-    Its  behaviour is not guaranteed if used externally. [#302](https://github.com/commercetools/commercetools-sync-java/issues/302)
-    - **ProductType Sync** - Renamed `ProductTypeUpdateLocalizedEnumActionUtils` to `LocalizedEnumsUpdateActionUtils.` [#302](https://github.com/commercetools/commercetools-sync-java/issues/302)
-    - **ProductType Sync** - Renamed `ProductTypeUpdatePlainEnumActionUtils` to `PlainEnumsUpdateActionUtils`. [#302](https://github.com/commercetools/commercetools-sync-java/issues/302)
-    - **ProductType Sync** - Renamed `LocalizedEnumUpdateActionsUtils` to `LocalizedEnumUpdateActionUtils`. [#302](https://github.com/commercetools/commercetools-sync-java/issues/302)
-    - **ProductType Sync** - Renamed `PlainEnumUpdateActionsUtils` to `PlainEnumUpdateActionUtils`. [#302](https://github.com/commercetools/commercetools-sync-java/issues/302)
+    - **ProductType Sync** - `EnumsUpdateActionUtils` is now `EnumValuesUpdateActionUtils` and is meant to be **only used internally** by the library. 
+    Its  behaviour is not guaranteed if used externally. [#300](https://github.com/commercetools/commercetools-sync-java/issues/300)
```
butenkor: If it should be use internally only why not change class scope instead?

butenkor: Is the linked issue correct?

heshamMassoud: @butenkor because it's used by two different packages `types` and `productTypes`, so it has to be `public` .

heshamMassoud: yes, because this refactoring happened in the type pr. So the commit should be linked in this issue.

docs/RELEASE_NOTES.md
heshamMassoud: 1b40b57

Issue #368, opened 2019-01-15
A set attribute with an empty value causes a NoSuchElementException


heshamMassoud:
#### Description

Reference replacement of ids with keys in the `VariantReferenceReplacementUtils` of a set attribute with an empty set as value causes `NoSuchElementException`

#### Cause
https://github.com/commercetools/commercetools-sync-java/blob/f81def7c624717dd3cb147c6a932e0ce59391cff/src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java#L177

The problem is the method expects the list of values to never be empty. However, CTP allows that case.

heshamMassoud:
closed in 1.1.1

PR #369, opened 2019-01-15
368 handle set attributes with empty values

heshamMassoud:
Addresses #368 

- Tests
    - [x] Unit 
    - [x] Integration
- [ ] ~Documentation~ not relevant.

### DIFF
--- a/src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
+++ b/src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
@@ -19,6 +19,7 @@ import io.sphere.sdk.types.CustomFieldsDraft;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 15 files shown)


src/main/java/com/commercetools/sync/products/utils/VariantReferenceReplacementUtils.java
```diff
@@ -174,7 +175,10 @@ static boolean isProductReference(@Nonnull final Attribute attribute) {
 
     static boolean isProductReferenceSet(@Nonnull final Attribute attribute) {
         final JsonNode valueAsJsonNode = attribute.getValueAsJsonNode();
-        return (valueAsJsonNode instanceof ArrayNode) && isValueAProductReference(valueAsJsonNode.elements().next());
+        final Iterator<JsonNode> setIterator = valueAsJsonNode.elements();
+
+        return (valueAsJsonNode instanceof ArrayNode) && setIterator.hasNext()
```
butenkor: I would put instanceof check before line 178.

heshamMassoud: af73bd2

Issue #379, opened 2019-04-23
Add support for sync of cart discounts


butenkor:
Allow to import/sync cart discounts. As there is no cart discount key available from API yet we should use cart discount key defined as custom fields. 
Note: Check current state of API development (CTP-1913) to check if native keys will be available soon.

heshamMassoud:
Shipped with release [`1.3.0`](https://github.com/commercetools/commercetools-sync-java/releases/tag/1.3.0)

PR #381, opened 2019-05-06
#379 add cart discount sync options and builder with unit tests

ahmetoz:
#### Description
This PR simply adds the `CartDiscountSyncOptions` for cart-discount sync.

#### Hints for Review
- I choose **50** as batch size, for now.. payload size is dependent on the predicate size, I have checked some production cases which has some long predicate + custom fields.

- I used name field instead of the key (because CTP is not supporting the key yet) for the unique field, I will update this to a custom field probably in next PR's.

### DIFF
--- a/src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptions.java
+++ b/src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptions.java
@@ -1,0 +1,30 @@
+package com.commercetools.sync.cartdiscounts;
+
+import com.commercetools.sync.commons.BaseSyncOptions;
+import com.commercetools.sync.commons.utils.TriFunction;
+import io.sphere.sdk.cartdiscounts.CartDiscount;
+import io.sphere.sdk.cartdiscounts.CartDiscountDraft;
+import io.sphere.sdk.client.SphereClient;
+import io.sphere.sdk.commands.UpdateAction;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+public class CartDiscountSyncOptions extends BaseSyncOptions<CartDiscount, CartDiscountDraft> {
+
+    CartDiscountSyncOptions(
+        @Nonnull final SphereClient ctpClient,
+        @Nullable final BiConsumer<String, Throwable> errorCallBack,
+        @Nullable final Consumer<String> warningCallBack,
+        final int batchSize,
+        @Nullable final TriFunction<List<UpdateAction<CartDiscount>>, CartDiscountDraft,
+            CartDiscount, List<UpdateAction<CartDiscount>>> beforeUpdateCallback,
+        @Nullable final Function<CartDiscountDraft, CartDiscountDraft> beforeCreateCallback) {
+
+        super(ctpClient, errorCallBack, warningCallBack, batchSize, beforeUpdateCallback, beforeCreateCallback);
+    }
+}
--- a/src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptionsBuilder.java
+++ b/src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptionsBuilder.java
@@ -1,0 +1,58 @@
+package com.commercetools.sync.cartdiscounts;
+
+import com.commercetools.sync.commons.BaseSyncOptionsBuilder;
+import io.sphere.sdk.cartdiscounts.CartDiscount;
+import io.sphere.sdk.cartdiscounts.CartDiscountDraft;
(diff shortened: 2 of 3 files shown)


src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptions.java
```diff
@@ -0,0 +1,30 @@
...
+import io.sphere.sdk.commands.UpdateAction;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+public class CartDiscountSyncOptions extends BaseSyncOptions<CartDiscount, CartDiscountDraft> {
```
heshamMassoud: Should be `final`

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptionsBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param ctpClient instance of the {@link SphereClient} responsible for interaction with the target CTP project.
+     * @return new instance of {@link CartDiscountSyncOptionsBuilder}
+     */
+    public static CartDiscountSyncOptionsBuilder of(@Nonnull final SphereClient ctpClient) {
+        return new CartDiscountSyncOptionsBuilder(ctpClient).batchSize(BATCH_SIZE_DEFAULT);
+    }
+
+    /**
+     * Creates new instance of {@link CartDiscountSyncOptions} enriched with all attributes provided to {@code this}
```
heshamMassoud: `Creates new instance` -> `Creates a new instance`

src/test/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptionsBuilderTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+public class CartDiscountSyncOptionsBuilderTest {
+
+    private static final SphereClient CTP_CLIENT = mock(SphereClient.class);
+    private CartDiscountSyncOptionsBuilder cartDiscountSyncOptionsBuilder =
+        CartDiscountSyncOptionsBuilder.of(CTP_CLIENT);
+
+    @Test
+    public void of_WithClient_ShouldCreateCartDiscountSyncOptionsBuilder() {
+        final CartDiscountSyncOptionsBuilder builder = CartDiscountSyncOptionsBuilder.of(CTP_CLIENT);
```
heshamMassoud: You can use here the global instance attribute `cartDiscountSyncOptionsBuilder` at line 33 instead of defining a new one.

ahmetoz:
yes, a nice point about the duplicate tests @heshamMassoud I agree about the duplicate test cases would take some time, but it's normal to duplicate some test cases at least we are sure we are not duplicating the actual implementation 😉

src/test/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptionsBuilderTest.java
```diff
@@ -0,0 +1,273 @@
...
+        final List<UpdateAction<CartDiscount>> filteredList =
+            cartDiscountSyncOptions
+                .applyBeforeUpdateCallBack(updateActions, mock(CartDiscountDraft.class), mock(CartDiscount.class));
+        assertThat(filteredList).isNotEqualTo(updateActions);
+        assertThat(filteredList).isEmpty();
+    }
+
+    @Test
+    public void applyBeforeCreateCallBack_WithCallback_ShouldReturnFilteredDraft() {
+        //todo: there is no key, instead of key, use name as an option
```
heshamMassoud: But actually why do we need this TODO? We don't need to use the key here at all? It's just a test for the callback, so the key/name of the draft doesn't matter which is used.

heshamMassoud:
Not sure what do you mean by that?

> but it's normal to duplicate some test cases at least we are sure we are not duplicating the actual implementation

ahmetoz:
> Not sure what do you mean by that?
> 
> > but it's normal to duplicate some test cases at least we are sure we are not duplicating the actual implementation

I meant, we are extending the CartDiscountSyncOptions from BaseSyncOptions so the actual implementations are not duplicated, just tests are duplicated. I'd noted this comment of yours about the duplication maybe I can take this refactoring into my priority also I will open an enhancement issue for it to not to forget, now I will focus more important topics and for now, I will skip this comment about the duplication of tests.

src/test/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptionsBuilderTest.java
ahmetoz: you are goddamn right 😄

PR #387, opened 2019-05-21
#379 add cart discount sync options and builder with unit tests

ahmetoz:
#### Description
This PR simply adds the `CartDiscountSyncOptions` for cart-discount sync.

#### Hints for Review
- I choose **50** as batch size, for now.. payload size is dependent on the predicate size, I have checked some production cases which has some long predicate + custom fields.

- I used name field instead of the key (because CTP is not supporting the key yet) for the unique field, I will update this to a custom field probably in next PR's.

### DIFF
--- a/src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptions.java
+++ b/src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSyncOptions.java
@@ -1,0 +1,30 @@
+package com.commercetools.sync.cartdiscounts;
+
+import com.commercetools.sync.commons.BaseSyncOptions;
+import com.commercetools.sync.commons.utils.TriFunction;
+import io.sphere.sdk.cartdiscounts.CartDiscount;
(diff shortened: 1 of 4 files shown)


ahmetoz:
@JudeNiroshan @heshamMassoud please try to review this small boilerplate PR about cart discount sync options object in 2 days 🙏 
otherwise, it would be hard to manage many PRs for cart discounts.
Thanks 👍

ahmetoz:
after discussion with @heshamMassoud I opened this issue about the duplicate tests for SyncOptions, this is not the main topic for this PR right now.  https://github.com/commercetools/commercetools-sync-java/issues/389 

Also @heshamMassoud you add your comments to another PR which I merged when I trying to clean some branches for cart-discounts. Anyway, you can check the changes on these commits. Thanks again for the review 👍

ahmetoz:
changes in this PR is already discussed with @heshamMassoud, I assume everyone checked this PR, because of that I'm merging it. Thanks 👍

PR #388, opened 2019-05-21
#379 add cart discount update action utils and unit tests

ahmetoz:
#### Description
Add update actions utilities for cart discount sync utils. 
https://docs.commercetools.com/http-api-projects-cartDiscounts#update-actions

#### Hints for Review
- Custom type actions and some factory methods are missing in jvm-sdk because of that some actions are waiting for the jvm-sdk release.
https://github.com/commercetools/commercetools-jvm-sdk/issues/1918
https://github.com/commercetools/commercetools-jvm-sdk/issues/1920
- 79d33b2 this commit also in other PR (check out the other PR https://github.com/commercetools/commercetools-sync-java/pull/387 for it.)


src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountSyncUtilsTest.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final SphereClient CTP_CLIENT = mock(SphereClient.class);
+    private static final CartDiscount CART_DISCOUNT_WITH_SHIPPING_TARGET =
+        readObjectFromResource("cart-discount-with-shipping-target.json", CartDiscount.class);
+
+    @Test
+    public void buildActions_FromDraftsWithDifferentNameValues_ShouldBuildUpdateActions() {
+        final LocalizedString newName =
+            LocalizedString.of(Locale.GERMAN, "Neu Name", Locale.ENGLISH, "new name");
+
+        final CartDiscountDraft newCartDiscount = CartDiscountDraftBuilder
```
lojzatran: Is it possible to create a draft from `CART_DISCOUNT_WITH_SHIPPING_TARGET` and change only what you need? It's quite hard to review what's the difference between this newCartDiscount and `CART_DISCOUNT_WITH_SHIPPING_TARGET` and why you have only one action `ChangeName` returned.

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
```diff
@@ -0,0 +1,284 @@
...
+            () -> ChangeSortOrder.of(newCartDiscount.getSortOrder()));
+    }
+
+    /**
+     * Compares the {@link Boolean} requiresDiscountCode values of a {@link CartDiscount}
+     * and a {@link CartDiscountDraft} and returns an {@link UpdateAction}&lt;{@link CartDiscount}&gt; as a result
+     * in an {@link Optional}. If both the {@link CartDiscount} and the {@link CartDiscountDraft} have the same
+     * requiresDiscountCode value, then no update action is needed and hence an empty {@link Optional} is returned.
+     *
+     * <p>Note: A {@code null} {@code requiresDiscountCode} value in the {@link CartDiscount} is treated as a
```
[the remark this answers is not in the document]
ahmetoz: In the API documentation, there is no default value information 
https://docs.commercetools.com/http-api-projects-cartDiscounts#cartdiscountdraft 
but when I create a cart discount without **requiresDiscountCode**, value is false, for **isActive** value is true.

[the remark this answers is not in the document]
ahmetoz: but it's interesting when I try to set an update action with null from jvm-sdk  `ChangeRequiresDiscountCode.of(null)`,  it returned this API response: 
```
{
  "statusCode" : 400,
  "message" : "Request body does not contain valid JSON.",
  "errors" : [ {
    "code" : "InvalidJsonInput",
    "message" : "Request body does not contain valid JSON.",
    "detailedErrorMessage" : "actions -> requiresDiscountCode: Missing required value"
  } ]
}
```

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,550 @@
...
+    public void buildChangeValueUpdateAction_WithSameAbsoluteValues_ShouldNotBuildUpdateAction() {
+        final CartDiscountDraft newCartDiscountDraft = mock(CartDiscountDraft.class);
+        final CartDiscountValue values =
+            CartDiscountValue.ofAbsolute(asList(MoneyImpl.of(10, USD), MoneyImpl.of(10, EUR)));
+        when(newCartDiscountDraft.getValue()).thenReturn(values);
+
+        final Optional<UpdateAction<CartDiscount>> changeValueUpdateAction =
+            buildChangeValueUpdateAction(CART_DISCOUNT_WITH_ABSOLUTE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeValueUpdateAction).contains(ChangeValue.of(values));
```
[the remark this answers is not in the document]
ahmetoz: yes, you are right, nice catch. the values are the same but with a different order, I added an additional test for it but I'm not sure should we create different action for this case ? @heshamMassoud  what do you think about this case ?

Commit: e8f91fb

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,550 @@
...
+        when(newCartDiscountDraft.getTarget()).thenReturn(cartDiscountTarget);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+            buildChangeTargetUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).contains(ChangeTarget.of(cartDiscountTarget));
+    }
+
+    @Test
+    public void buildChangeTargetUpdateAction_WithSameLineItemTargetValues_ShouldBuildUpdateAction() {
```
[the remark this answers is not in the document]
ahmetoz: oh shame, in ChangeTarget action I am comparing the values not targets 😱 
I updated this line, thanks for the catch 🙁 👍 

Commit: 3022650

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountSyncUtilsTest.java
ahmetoz: I could not deep copy from the CART_DISCOUNT_WITH_SHIPPING_TARGET but I hope it's more readable now.
 
Commit: 5b761ff

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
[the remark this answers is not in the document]
lojzatran: Report to support team and ask them so we know for sure?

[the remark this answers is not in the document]
ahmetoz: I reported to the supporting team, (SUPPORT-5817), thanks for the advice. 👍

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        assertThat(changeValueUpdateAction).contains(ChangeValue.of(null));
+    }
+
+    @Test
+    public void buildChangeCartPredicateUpdateAction_WithDifferentValues_ShouldBuildUpdateAction() {
+        final CartDiscountDraft newCartDiscountDraft = mock(CartDiscountDraft.class);
+        when(newCartDiscountDraft.getCartPredicate()).thenReturn("1 = 1");
+
+        final Optional<UpdateAction<CartDiscount>> changeCartPredicateUpdateAction =
+            buildChangeCartPredicateUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
```
heshamMassoud: Don't you think it's simpler when you can see right here both values that are under test/comparison right here, instead of having to go the json file under this constant to see how the cart predicate there looks like? 

This is what I mean when you see the values under test close to each other:

```java
final CartDiscountDraft newCartDiscountDraft = mock(CartDiscountDraft.class);
when(newCartDiscountDraft.getCartPredicate()).thenReturn("foo");

final CartDiscount oldCartDiscount = mock(CartDiscount.class);
when(oldCartDiscount.getCartPredicate()).thenReturn("bar");

final Optional<UpdateAction<CartDiscount>> changeCartPredicateUpdateAction =
buildChangeCartPredicateUpdateAction(oldCartDiscount, newCartDiscountDraft);

assertThat(changeCartPredicateUpdateAction).contains(ChangeCartPredicate.of("foo"));
```

WDYT?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.getTarget()).thenReturn(cartDiscountTarget);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+            buildChangeTargetUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).isNotPresent();
+    }
+
+    @Test
+    public void buildChangeTargetUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud:
```suggestion
    public void buildChangeTargetUpdateAction_WithNullNewValue_ShouldBuildUpdateAction() {
```
What about tests for:

both null values (old and new)?
just old null value?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    @Test
+    public void buildChangeTargetUpdateAction_WithDifferentCustomLineItemValues_ShouldBuildUpdateAction() {
+        final CartDiscountDraft newCartDiscountDraft = mock(CartDiscountDraft.class);
+        final CartDiscountTarget cartDiscountTarget = CustomLineItemsTarget.of("1 = 1");
+        when(newCartDiscountDraft.getTarget()).thenReturn(cartDiscountTarget);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+            buildChangeTargetUpdateAction(CART_DISCOUNT_WITH_CUSTOM_LINE_ITEM_TARGET, newCartDiscountDraft);
```
heshamMassoud: check my comment on this line: https://github.com/commercetools/commercetools-sync-java/pull/388/files#diff-30e0149be0301bb3ff8a918ecc1b08d5R153

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    @Test
+    public void buildChangeTargetUpdateAction_WithLineItemAndShippingTargetValues_ShouldBuildUpdateAction() {
+        final CartDiscountDraft newCartDiscountDraft = mock(CartDiscountDraft.class);
+        final CartDiscountTarget cartDiscountTarget = ShippingCostTarget.of();
+        when(newCartDiscountDraft.getTarget()).thenReturn(cartDiscountTarget);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+            buildChangeTargetUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
```
heshamMassoud: also here: https://github.com/commercetools/commercetools-sync-java/pull/388/files#diff-30e0149be0301bb3ff8a918ecc1b08d5R153

heshamMassoud: There are many cases which are not covered by the unit tests:

`MultiBuyCustomLineItemsTarget`
`MultiBuyLineItemsTarget`

with all the variation of their sub fields! :(

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.getName()).thenReturn(LocalizedString.of(Locale.ENGLISH, "newName"));
+
+        final Optional<UpdateAction<CartDiscount>> changeNameUpdateAction =
+            buildChangeNameUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeNameUpdateAction).contains(ChangeName.of(LocalizedString.of(Locale.ENGLISH, "newName")));
+    }
+
+    @Test
+    public void buildChangeNameUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: I think you miss understood the `WithNullValues` throughout this test. Usually it means that both old and new are `null`. But in these cases it's only then new one which is `null`

heshamMassoud: What about tests for:

both null values (old and new)?
just old null value?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+
+        final Optional<UpdateAction<CartDiscount>> setDescriptionUpdateAction =
+            buildSetDescriptionUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(setDescriptionUpdateAction)
+            .contains(SetDescription.of(LocalizedString.of(Locale.ENGLISH, "new-description")));
+    }
+
+    @Test
+    public void buildSetDescriptionUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: like all the comments above: 

What about tests for:

both null values (old and new)?
just old null value?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.getSortOrder()).thenReturn("0.3");
+
+        final Optional<UpdateAction<CartDiscount>> changeChangeSortOrderUpdateAction =
+            buildChangeSortOrderUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeChangeSortOrderUpdateAction).contains(ChangeSortOrder.of("0.3"));
+    }
+
+    @Test
+    public void buildChangeSortOrderUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: also here

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.isRequiresDiscountCode()).thenReturn(false);
+
+        final Optional<UpdateAction<CartDiscount>> changeRequiresDiscountCodeUpdateAction =
+            buildChangeRequiresDiscountCodeUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeRequiresDiscountCodeUpdateAction).contains(ChangeRequiresDiscountCode.of(false));
+    }
+
+    @Test
+    public void buildChangeRequiresDiscountCodeUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: also here

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.isActive()).thenReturn(true);
+
+        final Optional<UpdateAction<CartDiscount>> changeIsActiveUpdateAction =
+            buildChangeIsActiveUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeIsActiveUpdateAction).contains(ChangeIsActive.of(true));
+    }
+
+    @Test
+    public void buildChangeIsActiveUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: both are not `null` here

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+
+        final Optional<UpdateAction<CartDiscount>> changeStackingModeUpdateAction =
+            buildChangeStackingModeUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeStackingModeUpdateAction)
+            .contains(ChangeStackingMode.of(StackingMode.STACKING));
+    }
+
+    @Test
+    public void buildChangeStackingModeUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: here too

heshamMassoud: ok, I won't comment it again. You get the idea for the rest of the tests.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountSyncUtilsTest.java
heshamMassoud: I definitley agree with @lojzatran comment here. It's really hard throughout the rest of the unit tests.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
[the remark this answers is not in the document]
heshamMassoud: https://github.com/commercetools/commercetools-sync-java/pull/388#discussion_r289374307

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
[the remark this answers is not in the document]
heshamMassoud: > @ahmetoz but it's interesting when I try to set an update action with null from jvm-sdk ChangeRequiresDiscountCode.of(null), it returned this API response:

In the docs it's not optional: https://docs.commercetools.com/http-api-projects-cartDiscounts#change-requires-discountcode

So it makes sense you got this error message, right?

heshamMassoud:
@ahmetoz Please don't forget to add release notes for this PR and also the other which were already merged (sync options, statistics, etc..) It will make it easier not to forget a certain release notes entry.

heshamMassoud:
I am wondering why the test coverage for this PR is not passing 🤔

ahmetoz:
> @ahmetoz Please don't forget to add release notes for this PR and also the other which were already merged (sync options, statistics, etc..) It will make it easier not to forget a certain release notes entry.

My plan was adding the releases notes separately with documentation.. https://github.com/commercetools/commercetools-sync-java/pull/399

ahmetoz:
> I am wondering why the test coverage for this PR is not passing 🤔

Thanks for the notice, I will definitely check, maybe I need to remove CustomType action classes (which is not required in this PR) and comments about it to make PR more easy to check.

ahmetoz:
@heshamMassoud thanks for the comments for this PR, it seems I will update the all tests to make it more easy to read. https://github.com/commercetools/commercetools-sync-java/pull/388#discussion_r289379907

src/main/java/com/commercetools/sync/cartdiscounts/helpers/CartDiscountCustomActionBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+import io.sphere.sdk.cartdiscounts.CartDiscount;
+import io.sphere.sdk.cartdiscounts.commands.updateactions.SetCustomField;
+import io.sphere.sdk.cartdiscounts.commands.updateactions.SetCustomType;
+import io.sphere.sdk.commands.UpdateAction;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Map;
+
+public class CartDiscountCustomActionBuilder implements GenericCustomActionBuilder<CartDiscount> {
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/388#issuecomment-498187027

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountSyncUtils.java
```diff
@@ -0,0 +1,74 @@
...
+import static com.commercetools.sync.cartdiscounts.utils.CartDiscountUpdateActionUtils.buildChangeStackingModeUpdateAction;
+import static com.commercetools.sync.cartdiscounts.utils.CartDiscountUpdateActionUtils.buildChangeTargetUpdateAction;
+import static com.commercetools.sync.cartdiscounts.utils.CartDiscountUpdateActionUtils.buildChangeValueUpdateAction;
+import static com.commercetools.sync.cartdiscounts.utils.CartDiscountUpdateActionUtils.buildSetDescriptionUpdateAction;
+import static com.commercetools.sync.cartdiscounts.utils.CartDiscountUpdateActionUtils.buildSetValidDatesUpdateAction;
+import static com.commercetools.sync.commons.utils.OptionalUtils.filterEmptyOptionals;
+
+public final class CartDiscountSyncUtils {
+    //    private static final CartDiscountCustomActionBuilder cartDiscountCustomActionBuilder =
+    //        new CartDiscountCustomActionBuilder();
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/388#issuecomment-498187027

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
```diff
@@ -0,0 +1,284 @@
...
+        @Nonnull final CartDiscount oldCartDiscount,
+        @Nonnull final CartDiscountDraft newCartDiscount) {
+        final StackingMode stackingMode = ofNullable(newCartDiscount.getStackingMode()).orElse(StackingMode.STACKING);
+
+        return buildUpdateAction(oldCartDiscount.getStackingMode(), stackingMode,
+            () -> ChangeStackingMode.of(stackingMode));
+    }
+
+
+}
```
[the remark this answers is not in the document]
ahmetoz: 6320595

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountSyncUtils.java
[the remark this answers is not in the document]
ahmetoz: 21d82bf

src/main/java/com/commercetools/sync/cartdiscounts/helpers/CartDiscountCustomActionBuilder.java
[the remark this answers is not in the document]
ahmetoz: 21d82bf

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.getValue()).thenReturn(sameValuesWithDifferentOrder);
+
+        final Optional<UpdateAction<CartDiscount>> changeValueUpdateAction =
+            buildChangeValueUpdateAction(CART_DISCOUNT_WITH_ABSOLUTE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeValueUpdateAction).contains(ChangeValue.of(sameValuesWithDifferentOrder));
+    }
+
+    @Test
+    public void buildChangeValueUpdateAction_WithSameValues_ShouldNotBuildUpdateAction() {
```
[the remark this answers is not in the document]
ahmetoz: 171a818

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.getValue()).thenReturn(tenEuro);
+
+        final Optional<UpdateAction<CartDiscount>> changeValueUpdateAction =
+            buildChangeValueUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeValueUpdateAction).contains(ChangeValue.of(tenEuro));
+    }
+
+    @Test
+    public void buildChangeValueUpdateAction_WithDifferentRelativeValues_ShouldBuildUpdateAction() {
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-sync-java/blob/f8651bce054b8f81a08fd8e98391483935ce9a5c/src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java#L91

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: 07723af

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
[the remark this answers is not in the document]
ahmetoz: 9df400c

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: f8651bc

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,563 @@
...
+        when(newCartDiscountDraft.getValue()).thenReturn(tenPercent);
+
+        final Optional<UpdateAction<CartDiscount>> changeValueUpdateAction =
+            buildChangeValueUpdateAction(CART_DISCOUNT_WITH_RELATIVE_VALUE, newCartDiscountDraft);
+
+        assertThat(changeValueUpdateAction).isNotPresent();
+    }
+
+    @Test
+    public void buildChangeValueUpdateAction_WithNullValues_ShouldBuildUpdateAction() {
```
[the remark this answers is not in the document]
ahmetoz: I could not manage to add different commits for null checks, for example, I added tests for these cases: 
``_WithJustNew...HasNullValue, _WithJustOld...HasNullValue_, _WithBoth...HaveNullValues_`` 8260475, maybe it's better if you check it from the latest commit.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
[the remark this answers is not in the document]
heshamMassoud: Please test same values of different instances: for example: `buildChangeValueUpdateAction_WithSameRelativeValuesOfDifferentInstances_ShouldNotBuildUpdateAction` and 
`buildChangeValueUpdateAction_WithSameGiftItemValueOfDifferentInstances_ShouldNotBuildUpdateAction`

and others..

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+        assertThat(changeTargetUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithDifferentMultiBuyLineItemsTargetValues_ShouldBuildUpdateAction() {
+        final CartDiscount oldCartDiscount = mock(CartDiscount.class);
+        // Given a cart with 6 items, the discount can be applied only once.
+        // As a result, 2 items will be discounted
+        // and 4 cheapest items will be marked as participating in this discount.
```
heshamMassoud: I think these comments are confusing to be here. They don't affect the tests in any way. 

That's other than the fact that they are wrong 🙈

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(oldCartDiscount.getTarget())
+                .thenReturn(MultiBuyLineItemsTarget.of("quantity > 0", 6L, 2L,  SelectionMode.CHEAPEST));
+
+        final CartDiscountDraft newCartDiscountDraft = mock(CartDiscountDraft.class);
+        // Given a cart with 6 items, the discount can be applied only once.
+        // As a result, 3 items will be discounted
+        // and 4 most expensive items will be marked as participating in this discount.
```
heshamMassoud: I think these comments are confusing to be here. They don't affect the tests in any way.

That's other than the fact that they are wrong 🙈

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+                buildChangeTargetUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).contains(ChangeTarget.of(newTarget));
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithSameMultiBuyLineItemsTargetValues_ShouldNotBuildUpdateAction() {
+        // Given a cart with 6 items, the discount can be applied only once.
+        // As a result, 3 items will be discounted
+        // and 4 most expensive items will be marked as participating in this discount.
```
heshamMassoud: I think these comments are confusing to be here. They don't affect the tests in any way.

That's other than the fact that they are wrong 🙈

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+        assertThat(changeTargetUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithDifferentMultiBuyCustomLineItemsTargetValues_ShouldBuildUpdateAction() {
+        final CartDiscount oldCartDiscount = mock(CartDiscount.class);
+        // Given a cart with 6 items, the discount can be applied only once.
+        // As a result, 2 custom items will be discounted
+        // and 4 cheapest custom items will be marked as participating in this discount.
```
heshamMassoud: I think these comments are confusing to be here. They don't affect the tests in any way.

That's other than the fact that they are wrong 🙈

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+                buildChangeTargetUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).contains(ChangeTarget.of(newTarget));
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithSameMultiBuyCustomLineItemsTargetValues_ShouldNotBuildUpdateAction() {
+        // Given a cart with 6 items, the discount can be applied only once.
+        // As a result, 3 custom items will be discounted
+        // and 4 most expensive custom items will be marked as participating in this discount.
```
heshamMassoud: I think these comments are confusing to be here. They don't affect the tests in any way.

That's other than the fact that they are wrong 🙈

heshamMassoud: Please add a test for testing same values of different instances (which would be the more common case actually happening)

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getTarget()).thenReturn(target);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+                buildChangeTargetUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithJustNewTargetHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullNewTarget`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getTarget()).thenReturn(null);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+                buildChangeTargetUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).contains(ChangeTarget.of(null));
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithJustOldTargetHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullOldTarget`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getTarget()).thenReturn(newTarget);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+                buildChangeTargetUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).contains(ChangeTarget.of(newTarget));
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithBothTargetHaveNullValues_ShouldNotBuildUpdateAction() {
```
heshamMassoud: `WithBothNullTargetValues`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.isActive()).thenReturn(false);
+
+        final Optional<UpdateAction<CartDiscount>> changeIsActiveUpdateAction =
+                buildChangeIsActiveUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeIsActiveUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeIsActiveUpdateAction_WithJustNewIsActiveHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullNewIsActive`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.isActive()).thenReturn(false);
+
+        final Optional<UpdateAction<CartDiscount>> changeIsActiveUpdateAction =
+                buildChangeIsActiveUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeIsActiveUpdateAction).contains(ChangeIsActive.of(false));
+    }
+
+    @Test
+    void buildChangeIsActiveUpdateAction_WithBothIsActiveHaveNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithBothNullIsActive`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.isActive()).thenReturn(null);
+
+        final Optional<UpdateAction<CartDiscount>> changeIsActiveUpdateAction =
+                buildChangeIsActiveUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeIsActiveUpdateAction).contains(ChangeIsActive.of(true));
+    }
+
+    @Test
+    void buildChangeIsActiveUpdateAction_WithJustOldIsActiveHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: this case cannot happen that you have the old is `null`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: what about the case if new is `null` and old is `true`?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+        final Optional<UpdateAction<CartDiscount>> changeNameUpdateAction =
+                buildChangeNameUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeNameUpdateAction).contains(ChangeName.of(name));
+    }
+
+    @Test
+    void buildChangeNameUpdateAction_WithSameValues_ShouldNotBuildUpdateAction() {
+        final LocalizedString name = LocalizedString.of(Locale.ENGLISH, "name");
```
heshamMassoud: Please add a test for testing same values of different instances (which would be the more common case actually happening)

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getName()).thenReturn(name);
+
+        final Optional<UpdateAction<CartDiscount>> changeNameUpdateAction =
+                buildChangeNameUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeNameUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeNameUpdateAction_WithJustNewNameHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: but you cannot have this case, since name is optional in draft.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getName()).thenReturn(null);
+
+        final Optional<UpdateAction<CartDiscount>> changeNameUpdateAction =
+                buildChangeNameUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeNameUpdateAction).contains(ChangeName.of(null));
+    }
+
+    @Test
+    void buildChangeNameUpdateAction_WithJustOldNameHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: you cannot have this case since name is not optional.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getName()).thenReturn(newName);
+
+        final Optional<UpdateAction<CartDiscount>> changeNameUpdateAction =
+                buildChangeNameUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeNameUpdateAction).contains(ChangeName.of(newName));
+    }
+
+    @Test
+    void buildChangeNameUpdateAction_WithBothNameHaveNullValues_ShouldNotBuildUpdateAction() {
```
heshamMassoud: you cannot have this case.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+        final Optional<UpdateAction<CartDiscount>> setDescriptionUpdateAction =
+                buildSetDescriptionUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(setDescriptionUpdateAction).contains(SetDescription.of(description));
+    }
+
+    @Test
+    void buildSetDescriptionUpdateAction_WithSameValues_ShouldNotBuildUpdateAction() {
+        final LocalizedString description = LocalizedString.of(Locale.ENGLISH, "cart-discount-1-desc");
```
heshamMassoud: Please add a test for testing same values of different instances (which would be the more common case actually happening)

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getDescription()).thenReturn(description);
+
+        final Optional<UpdateAction<CartDiscount>> setDescriptionUpdateAction =
+                buildSetDescriptionUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(setDescriptionUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildSetDescriptionUpdateAction_WithJustNewDescriptionHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullNewDescription`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getDescription()).thenReturn(null);
+
+        final Optional<UpdateAction<CartDiscount>> setDescriptionUpdateAction =
+                buildSetDescriptionUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(setDescriptionUpdateAction).contains(SetDescription.of(null));
+    }
+
+    @Test
+    void buildSetDescriptionUpdateAction_WithJustOldDescriptionHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullOldDescription`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getDescription()).thenReturn(newDesc);
+
+        final Optional<UpdateAction<CartDiscount>> setDescriptionUpdateAction =
+                buildSetDescriptionUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(setDescriptionUpdateAction).contains(SetDescription.of(newDesc));
+    }
+
+    @Test
+    void buildSetDescriptionUpdateAction_WithBothDescriptionHaveNullValues_ShouldNotBuildUpdateAction() {
```
heshamMassoud: `WithBothNullValues`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+        final Optional<UpdateAction<CartDiscount>> changeChangeSortOrderUpdateAction =
+                buildChangeSortOrderUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeChangeSortOrderUpdateAction).contains(ChangeSortOrder.of(sortOrder));
+    }
+
+    @Test
+    void buildChangeSortOrderUpdateAction_WithSameValues_ShouldNotBuildUpdateAction() {
+        final String sortOrder = "0.1";
```
heshamMassoud: Please add a test for testing same values of different instances (which would be the more common case actually happening)

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getSortOrder()).thenReturn(sortOrder);
+
+        final Optional<UpdateAction<CartDiscount>> changeChangeSortOrderUpdateAction =
+                buildChangeSortOrderUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeChangeSortOrderUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeSortOrderUpdateAction_WithJustNewSortOrderHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullNewSortOrder`

heshamMassoud: what about empty string value?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getSortOrder()).thenReturn(null);
+
+        final Optional<UpdateAction<CartDiscount>> changeChangeSortOrderUpdateAction =
+                buildChangeSortOrderUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeChangeSortOrderUpdateAction).contains(ChangeSortOrder.of(null));
+    }
+
+    @Test
+    void buildChangeSortOrderUpdateAction_WithJustOldSortOrderHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullOldSortOrder`

heshamMassoud: what about empty string value?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: but it seems that it is also not optional from docs: https://docs.commercetools.com/http-api-projects-cartDiscounts#cartdiscountdraft

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: but it seems that it is also not optional from docs: https://docs.commercetools.com/http-api-projects-cartDiscounts#cartdiscountdraft

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getSortOrder()).thenReturn(newSortOrder);
+
+        final Optional<UpdateAction<CartDiscount>> changeChangeSortOrderUpdateAction =
+                buildChangeSortOrderUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeChangeSortOrderUpdateAction).contains(ChangeSortOrder.of(newSortOrder));
+    }
+
+    @Test
+    void buildChangeSortOrderUpdateAction_WithBothSortOrderHaveNullValues_ShouldNotBuildUpdateAction() {
```
heshamMassoud: but it seems that it is also not optional from docs: https://docs.commercetools.com/http-api-projects-cartDiscounts#cartdiscountdraft

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.isRequiresDiscountCode()).thenReturn(true);
+
+        final Optional<UpdateAction<CartDiscount>> changeRequiresDiscountCodeUpdateAction =
+                buildChangeRequiresDiscountCodeUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeRequiresDiscountCodeUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeRequiresDiscountCodeUpdateAction_WithJustNewRequiresDisCodeHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: but it seems that it is also not optional from docs: https://docs.commercetools.com/http-api-projects-cartDiscounts#cartdiscountdraft

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.isRequiresDiscountCode()).thenReturn(null);
+
+        final Optional<UpdateAction<CartDiscount>> changeRequiresDiscountCodeUpdateAction =
+                buildChangeRequiresDiscountCodeUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeRequiresDiscountCodeUpdateAction).contains(ChangeRequiresDiscountCode.of(false));
+    }
+
+    @Test
+    void buildChangeRequiresDiscountCodeUpdateAction_WithJustOldRequiresDisCodeHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: but it seems that it is also not optional from docs: https://docs.commercetools.com/http-api-projects-cartDiscounts#cartdiscountdraft

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.isRequiresDiscountCode()).thenReturn(true);
+
+        final Optional<UpdateAction<CartDiscount>> changeRequiresDiscountCodeUpdateAction =
+                buildChangeRequiresDiscountCodeUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeRequiresDiscountCodeUpdateAction).contains(ChangeRequiresDiscountCode.of(true));
+    }
+
+    @Test
+    void buildChangeRequiresDiscountCodeUpdateAction_WithBothReqDisCodeHaveNullValues_ShouldNotBuildUpdateAction() {
```
heshamMassoud: but it seems that it is also not optional from docs: https://docs.commercetools.com/http-api-projects-cartDiscounts#cartdiscountdraft

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getStackingMode()).thenReturn(StackingMode.STOP_AFTER_THIS_DISCOUNT);
+
+        final Optional<UpdateAction<CartDiscount>> changeStackingModeUpdateAction =
+                buildChangeStackingModeUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeStackingModeUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeStackingModeUpdateAction_WithJustNewStackingModeHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: `WithOnlyNullNewStackingMode`

heshamMassoud: what if old value had `StackingMode.STACKING`?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getStackingMode()).thenReturn(null);
+
+        final Optional<UpdateAction<CartDiscount>> changeStackingModeUpdateAction =
+                buildChangeStackingModeUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeStackingModeUpdateAction).contains(ChangeStackingMode.of(StackingMode.STACKING));
+    }
+
+    @Test
+    void buildChangeStackingModeUpdateAction_WithJustOldRequiresDisCodeHasNullValue_ShouldBuildUpdateAction() {
```
heshamMassoud: old cannot be null, right?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+        final Optional<UpdateAction<CartDiscount>> changeStackingModeUpdateAction =
+                buildChangeStackingModeUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeStackingModeUpdateAction)
+                .contains(ChangeStackingMode.of(StackingMode.STOP_AFTER_THIS_DISCOUNT));
+    }
+
+    @Test
+    void buildChangeStackingModeUpdateAction_WithBothReqDisCodeHaveNullValues_ShouldBuildUpdateAction() {
```
heshamMassoud: old cannot be null, right?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+        final Optional<UpdateAction<CartDiscount>> setValidFromUpdateAction =
+                buildSetValidFromUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(setValidFromUpdateAction).contains(SetValidFrom.of(now));
+    }
+
+    @Test
+    void buildSetValidFromUpdateAction_WithSameValues_ShouldNotBuildUpdateAction() {
+        final ZonedDateTime validFrom = ZonedDateTime.parse("2019-04-30T22:00:00.000Z");
```
heshamMassoud: Please add a test for testing same values of different instances (which would be the more common case actually happening)

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        final Optional<UpdateAction<CartDiscount>> setValidFromUpdateAction =
+                buildSetValidFromUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(setValidFromUpdateAction).contains(SetValidFrom.of(null));
+    }
+
+    @Test
+    void buildSetValidFromUpdateAction_WithJustOldSetValidFromDateHasNullValue_ShouldBuildUpdateAction() {
+        final CartDiscount oldCartDiscount = mock(CartDiscount.class);
+        when(oldCartDiscount.getValidFrom()).thenRetu
```
heshamMassoud: Please add a test for testing same values of different instances (which would be the more common case actually happening)

src/test/resources/cart-discount-with-relative-value.json
```diff
@@ -0,0 +1,30 @@
+{
```
heshamMassoud: I don't see where this is used.

src/test/resources/cart-discount-with-gift-item-value.json
```diff
@@ -0,0 +1,31 @@
+{
```
heshamMassoud: I don't see where this is used.

src/test/resources/cart-discount-with-custom-line-item-target.json
```diff
@@ -0,0 +1,32 @@
+{
```
heshamMassoud: I don't see where this is used.

src/test/resources/cart-discount-with-absolute-value.json
```diff
@@ -0,0 +1,43 @@
+{
```
heshamMassoud: I don't see where this is used.

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
```diff
@@ -34,18 +39,59 @@
      * If both the {@link CartDiscount} and the {@link CartDiscountDraft} have the same cart discount value,
      * then no update action is needed and hence an empty {@link Optional} is returned.
      *
+     * <p>Note: Order is not considered as an update action for the {@link AbsoluteCartDiscountValue}s
```
heshamMassoud: `Order is not significant when comparing {@link AbsoluteCartDiscountValue}s`

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
```diff
@@ -34,18 +39,59 @@
...
+    @Nonnull
+    private static Optional<UpdateAction<CartDiscount>> buildChangeAbsoluteValueUpdateAction(
+            @Nonnull final List<MonetaryAmount> amountsFromCartDiscount,
+            @Nonnull final List<MonetaryAmount> amountsFromCartDiscountDraft) {
+
+        final List<MonetaryAmount> newAmounts = new ArrayList<>();
+        boolean isAmountNew;
+        for (MonetaryAmount amountDraft : amountsFromCartDiscountDraft) {
+            isAmountNew = true;
+            for (MonetaryAmount amount : amountsFromCartDiscount) {
```
heshamMassoud: maybe a more efficient implementation than `n^2` implementation is to: 
- Check if they have different size first, 
   - if yes, then build action. 
   - if not, then 1. sort then 2. use equals for comparison.

refernece: https://stackoverflow.com/a/13501200/5654504

ahmetoz:
> ok, about the release notes. Why it's important with every PR is that we don't forget something like the bugfix in this line: https://github.com/commercetools/commercetools-sync-java/pull/388/files#diff-dfda92e7db082817e6e5c96dd809efdeR183

https://github.com/commercetools/commercetools-sync-java/pull/399/files#diff-aa75d2e7837774f5cb565717b6744a35R45
I actually did not forget it, but yes you are right 😉

src/test/resources/cart-discount-with-relative-value.json
ahmetoz: I remove these ones in custom type related branch but make sense to remove it in this PR too.

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
ahmetoz: is the complexity important for this algorithm because N won't be big enough for performance? 

they could be in different sizes but could be the same elements, the draft could have nullable values or duplicate values.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        final GiftLineItemCartDiscountValue oldGiftLineItemCartDiscountValue =
+                GiftLineItemCartDiscountValue.of(Reference.of(Product.referenceTypeId(), "product-1"),
+                        1, null, null);
+
+        final CartDiscount oldCartDiscount = mock(CartDiscount.class);
+        when(oldCartDiscount.getValue()).thenReturn(oldGiftLineItemCartDiscountValue);
+
+        final CartDiscountDraft newCartDiscountDraft = mock(CartDiscountDraft.class);
+        final GiftLineItemCartDiscountValue newGiftLineItemCartDiscountValue =
+                GiftLineItemCartDiscountValue.of(Reference.of(Product.referenceTypeId(), "product-2"),
```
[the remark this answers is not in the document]
ahmetoz: 860c750

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: bd16451

src/test/resources/cart-discount-with-relative-value.json
ahmetoz: 3c6ea89

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getTarget()).thenReturn(cartDiscountTarget);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+                buildChangeTargetUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).contains(ChangeTarget.of(cartDiscountTarget));
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithShippingTargetValues_ShouldNotBuildUpdateAction() {
```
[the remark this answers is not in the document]
ahmetoz: de27f08

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getCartPredicate()).thenReturn(newCartPredicate);
+
+        final Optional<UpdateAction<CartDiscount>> changeCartPredicateUpdateAction =
+                buildChangeCartPredicateUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeCartPredicateUpdateAction).contains(ChangeCartPredicate.of(newCartPredicate));
+    }
+
+    @Test
+    void buildChangeCartPredicateUpdateAction_WithBothCartPredicateHaveNullValues_ShouldNotBuildUpdateAction() {
```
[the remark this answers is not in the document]
ahmetoz: according to the API, cart discount predicate cannot be invalid like (null or ""), so this test is not necessary, I already removed it.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        when(newCartDiscountDraft.getTarget()).thenReturn(cartDiscountTarget);
+
+        final Optional<UpdateAction<CartDiscount>> changeTargetUpdateAction =
+                buildChangeTargetUpdateAction(oldCartDiscount, newCartDiscountDraft);
+
+        assertThat(changeTargetUpdateAction).isNotPresent();
+    }
+
+    @Test
+    void buildChangeTargetUpdateAction_WithDifferentCustomLineItemValues_ShouldBuildUpdateAction() {
```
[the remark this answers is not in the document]
ahmetoz: d68df14

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: Actually, if it's null it will not create update action because the default value for the isActive is true. But I realized, I removed the test for it, I have added it back : 92d640a
https://github.com/commercetools/commercetools-sync-java/blob/860c750f8c6afa2b46ef472eda3c0528961d94fa/src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java#L183-L197

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: c03a57e

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: why is this test not there anymore?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: why is this test not there anymore?

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: 😄  but why did you remove the old test when the old is `false`?  I understand what would happen, but we should have tests for both cases.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: why is this test not there anymore?

heshamMassoud: ah ok, because old cannot be `null`

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: but actually it's not optional so user should not pass null at all.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: according to the API, this is not possible, but both can be null in case of GiftLineItem. If the old is not null, new cannot be un-setted.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: same, https://github.com/commercetools/commercetools-sync-java/pull/388#discussion_r297097344 if the old is null, new should be null, this case can just happen if the value type is GiftLineItem

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: isActive could not be null, it's true or false.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
heshamMassoud: but why did you remove also the previous case when the old value has `StackingMode.STOP_AFTER_THIS_DISCOUNT`

ahmetoz: 🤔 It would be a different test, because if the new value in the draft is not set (null) means it's `Stacking`, I will add these test too, it could be helpful for the active, require discount code case.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: I will add these tests.

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: 5bd6ba4

src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
ahmetoz: 5bd6ba4

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtils.java
heshamMassoud: > they could be in different sizes but could be the same elements, the draft could have nullable values or duplicate values.

actually duplicate (and null) values are not allowed by the platform (I have just tried it now):
```json
"code": "InvalidOperation",
"message": "Duplicate currencies: [EUR]"
```

So If someone is trying to sync in duplicate money values, then (s)he will get this error message from the platform. The sync library is not responsible/expected to remove those duplicate or null values. 

> is the complexity important for this algorithm because N won't be big enough for performance?

I would say the complexity here is very critical, since I think there should not be so many `Money` Objects in the list. **But I only _think_.** So why not just have the better performant one? 

Having efficient code gives a good impression about the code base in general to any coder. Which in turn leads to maintaining it also with quality.  Especially, that is self-understandable and not confusing and I think this better performant one is quite clear to understand.

ahmetoz: Efficient code is related to the N also related to the memory usage ext, anyway I will check this StackOverflow answer if it's more understandable, why not :)

ahmetoz: But this algorithm just checks the equality, we need to find the diff of the two lists too, check this test : https://github.com/commercetools/commercetools-sync-java/blob/5bd6ba4cf38ed2c70670271c091ecc3fa61e76f5/src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java#L285

Anway, let's discuss it together 😄

PR #390, opened 2019-05-24
#379 add projectSource IT tests for CartDiscountSync

ahmetoz:
#### Summary
project source IT's for CartDiscount, these tests check integration between source & target project and asserting the statistics object after sync.


src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/cartdiscounts/CartDiscountSyncIT.java
```diff
@@ -0,0 +1,161 @@
...
+    @AfterClass
+    public static void tearDown() {
+        deleteTypesFromTargetAndSource();
+        deleteCartDiscountsFromTargetAndSource();
+    }
+
+    @Test
+    public void sync_WithoutUpdates_ShouldReturnProperStatistics() {
+        // preparation
+        final List<CartDiscount> cartDiscounts = CTP_SOURCE_CLIENT
```
lojzatran: Why do you need `CTP_SOURCE_CLIENT`? Can't you just create cartDiscounts directly without loading it from a project? Because in your test, you don't need `CTP_SOURCE_CLIENT` anyway.

ahmetoz: nice question, yes I can create the test scenario as you explain, but this way test explain some use case which is different than the externalsource package, the scenario that you explaining it could be in the externalsource package, here we are focused to sync between 2 projects, so first we are getting resources from source and synchronizing with the target. I hope, I understand what you meant in your question.

note: check the `populateSourceProject();` before all test, it's creating some resource both in source and target projects, in the source project, there are 2 cart discounts, cartDiscount1 and cartDiscount2, in the target it's cartDiscount1.

PR #395, opened 2019-05-28
#379 add CartDiscountService IT

ahmetoz:
#### Summary
Add CartDiscountService IT's


src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,425 @@
...
+
+    @Test
+    void fetchCartDiscount_WithBlankKey_ShouldNotFetchCartDiscount() {
+        final Optional<CartDiscount> fetchedCartDiscountOptional =
+                cartDiscountService.fetchCartDiscount("").toCompletableFuture().join();
+        assertThat(fetchedCartDiscountOptional).isEmpty();
+    }
+
+    @Test
+    void fetchCartDiscount_WithNullKey_ShouldNotFetchCartDiscount() {
```
lojzatran: Do you need IT for this case? You already have unit test in the other PR. This IT doesn't add any value. Or you can remove unit tests. WDYT?

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,425 @@
...
+        final Optional<CartDiscount> fetchedCartDiscountOptional =
+                cartDiscountService.fetchCartDiscount(null).toCompletableFuture().join();
+        assertThat(fetchedCartDiscountOptional).isEmpty();
+    }
+
+
+    @Test
+    void fetchMatchingCartDiscountsByKeys_WithEmptySetOfKeys_ShouldReturnEmptySet() {
```
lojzatran: Do you need IT for this case? You already have unit test in the other PR. This IT doesn't add any value. Or you can remove unit tests. WDYT?

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,425 @@
...
+                        .toCompletableFuture()
+                        .join();
+
+        assertThat(matchingCartDiscounts).isEmpty();
+        assertThat(errorCallBackExceptions).isEmpty();
+        assertThat(errorCallBackMessages).isEmpty();
+    }
+
+    @Test
+    void fetchMatchingCartDiscountsByKeys_WithNonExistingKeys_ShouldReturnEmptySet() {
```
lojzatran: Do you need IT for this case? You already have unit test in the other PR. This IT doesn't add any value. Or you can remove unit tests. WDYT?

src/main/java/com/commercetools/sync/services/CartDiscountService.java
```diff
@@ -15,6 +15,26 @@
 import java.util.concurrent.CompletionStage;
 
 public interface CartDiscountService {
+
+    /**
+     * Given a {@code key}, this method first checks if a cached map of CartDiscount keys -&gt; ids is not empty.
```
lojzatran: Implementation is different than what you describe here. Your code:
```
if (keyToIdCache.containsKey(key)) {
```
vs your comment:
`this method first checks if a cached map of CartDiscount keys -> ids is not empty.`

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
ahmetoz: aha, nice catch 👍 makes sense I will remove these tests from IT. 
51088e2, be3a5fc

src/main/java/com/commercetools/sync/services/CartDiscountService.java
ahmetoz: you are right,  this is also missing on some other services, I will open an issue for this and link to this comment.
Commit: 1829967

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
ahmetoz: this is not in unit test just fetchCartDiscount methods.

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
ahmetoz: this is not in unit test just fetchCartDiscount methods.

src/main/java/com/commercetools/sync/services/CartDiscountService.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/issues/397

src/main/java/com/commercetools/sync/services/TypeService.java
```diff
@@ -23,13 +22,13 @@
...
      * in which the result of it's completion could contain an
-     * {@link Optional} with the id inside of it or an empty {@link Optional} if no {@link Channel} was
+     * {@link Optional} with the id inside of it or an empty {@link Optional} if no {@link Type} was
      * found in the CTP project with this key.
      *
      * @param key the key by which a {@link io.sphere.sdk.types.Type} id should be fetched from the CTP project.
      * @return {@link CompletionStage}&lt;{@link Optional}&lt;{@link String}&gt;&gt; in which the result of its
      *         completion could contain an {@link Optional} with the id inside of it or an empty {@link Optional} if no
-     *         {@link Channel} was found in the CTP project with this key.
+     *         {@link Type} was found in the CTP project with this key.
```
heshamMassoud: 👍

src/main/java/com/commercetools/sync/services/CartDiscountService.java
```diff
@@ -15,6 +15,24 @@
...
+     * could contain an {@link Optional} with the id inside of it or an empty {@link Optional}
+     * if no {@link CartDiscount} was found in the CTP project with this key.
+     *
+     * @param key the key by which a {@link CartDiscount} id should be fetched from the CTP project.
+     * @return {@link CompletionStage}&lt;{@link Optional}&lt;{@link String}&gt;&gt; in which the result of it's
+     *         completion could contain an {@link Optional} with the id inside of it or an empty {@link Optional} if no
+     *         {@link CartDiscount} was found in the CTP project with this key.
+     */
+    @Nonnull
+    CompletionStage<Optional<String>> fetchCachedCartDiscountId(@Nonnull final String key);
```
heshamMassoud: why do we need this service at all?

src/main/java/com/commercetools/sync/services/impl/CartDiscountServiceImpl.java
```diff
@@ -73,7 +95,7 @@ private String getKey(@Nonnull final CartDiscountDraft cartDiscountDraft) {
 
 
         return CtpQueryUtils.queryAll(syncOptions.getCtpClient(), cartDiscountQuery, identity())
-                            .thenApply(types -> types
+                            .thenApply(cartDiscounts -> cartDiscounts
```
heshamMassoud: 🙌

src/main/java/com/commercetools/sync/services/impl/CartDiscountServiceImpl.java
```diff
@@ -59,6 +59,28 @@ private String getKey(@Nonnull final CartDiscountDraft cartDiscountDraft) {
...
+            return CompletableFuture.completedFuture(Optional.empty());
+        }
+        if (keyToIdCache.containsKey(key)) {
+            return CompletableFuture.completedFuture(Optional.ofNullable(keyToIdCache.get(key)));
+        }
+        return fetchAndCache(key);
+    }
+
+    @Nonnull
+    private CompletionStage<Optional<String>> fetchAndCache(@Nonnull final String key) {
```
heshamMassoud: Why do we need those two methods?

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+
+    private List<String> errorCallBackMessages;
+    private List<Throwable> errorCallBackExceptions;
+
+    /**
+     * Cleans up the target test data that were built in this test class.
+     */
+    @AfterAll
+    static void tearDown() {
+        deleteTypes(CTP_TARGET_CLIENT);
```
heshamMassoud: why do we need to delete types?

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+        deleteCartDiscountsFromTargetAndSource();
+    }
+
+    /**
+     * Deletes types & cart discounts from the target CTP project, then it populates the project with test data.
+     */
+    @BeforeEach
+    void setup() {
+        deleteTypes(CTP_TARGET_CLIENT);
```
heshamMassoud: why do we need to delete types?

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
heshamMassoud: I agree a unit test for this would be better instead of an IT. Because then we won't need to call the expensive `@BeforeEach` here

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
heshamMassoud: I agree a unit test for this would be better. Because then we won't need to call the expensive @BeforeEach here

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+                .join();
+
+        assertThat(cartDiscountId).isNotEmpty();
+        assertThat(errorCallBackExceptions).isEmpty();
+        assertThat(errorCallBackMessages).isEmpty();
+    }
+
+    @Test
+    void fetchCachedCartDiscountId_WithBlankCartDiscountKey_ShouldReturnEmptyOptional() {
+        final Optional<String> cartDiscountId = cartDiscountService.fetchCachedCartDiscountId("  ")
```
heshamMassoud: also here a unit test would be better.

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+                .isExactlyInstanceOf(BadGatewayException.class);
+    }
+
+    @Test
+    void fetchCachedCartDiscountId_WithNotCachedExistingCartDiscount_ShouldFetchCartDiscount() {
+        final Optional<String> cartDiscountId = cartDiscountService.fetchCachedCartDiscountId(CART_DISCOUNT_KEY_1)
+                .toCompletableFuture()
+                .join();
+
+        assertThat(cartDiscountId).isNotEmpty();
```
heshamMassoud: I have a feeling we are too relaxed on the asserting here. Don't you think we need to assert that this key is actually cached?

heshamMassoud: by either checking the cache or by verifying that on a second fetch, no request is made.

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+        assertThat(fetchedCartDiscounts).hasSize(1);
+
+        final Optional<CartDiscount> cartDiscountOptional = CTP_TARGET_CLIENT
+                .execute(CartDiscountQuery.of()
+                        .withPredicates(queryModel -> queryModel.name().lang(Locale.ENGLISH).is(CART_DISCOUNT_KEY_1)))
+                .toCompletableFuture()
+                .join()
+                .head();
+
+        assertThat(cartDiscountOptional).isNotNull();
```
heshamMassoud: why do we need this?

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+    @Test
+    void fetchMatchingCartDiscountsByKeys_WithAllExistingSetOfKeys_ShouldCacheFetchedCartDiscountIds() {
+        //preparation
+        final Set<CartDiscount> fetchedCartDiscounts =
+                cartDiscountService.fetchMatchingCartDiscountsByKeys(singleton(CART_DISCOUNT_KEY_1))
+                        .toCompletableFuture().join();
+        assertThat(fetchedCartDiscounts).hasSize(1);
+
+        final Optional<CartDiscount> cartDiscountOptional = CTP_TARGET_CLIENT
+                .execute(CartDiscountQuery.of()
```
heshamMassoud: I don't understand why we do this fetch?

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+                        .withPredicates(queryModel -> queryModel.name().lang(Locale.ENGLISH).is(CART_DISCOUNT_KEY_1)))
+                .toCompletableFuture()
+                .join()
+                .head();
+
+        assertThat(cartDiscountOptional).isNotNull();
+
+        // Change cart discount key_1 on ctp
+        final String newKey = "new_key_1";
+        cartDiscountService.updateCartDiscount(cartDiscountOptional.get(),
```
heshamMassoud: it's not a good practice to use an own implementation of another service in the test of another service.

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+                    assertThat(created.getName()).isEqualTo(queried.getName());
+                    assertThat(created.getCartPredicate()).isEqualTo(queried.getCartPredicate());
+                    assertThat(created.getValue()).isEqualTo(queried.getValue());
+                    assertThat(created.getTarget()).isEqualTo(queried.getTarget());
+                    assertThat(created.getSortOrder()).isNotEqualTo(queried.getSortOrder());
+                }));
+
+        // Assert that the created cart discount is cached
+        final Optional<String> cartDiscountId =
+                spyCartDiscountService.fetchCachedCartDiscountId(CART_DISCOUNT_KEY_1).toCompletableFuture().join();
```
heshamMassoud: it's not a good practice to use an own implementation of another service in the test of another service.

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
```diff
@@ -0,0 +1,432 @@
...
+        assertThat(cartDiscountId).isPresent();
+        verify(spyClient, times(0)).execute(any(CartDiscountQuery.class));
+    }
+
+    @Test
+    void createCartDiscount_WithInvalidCartDiscount_ShouldHaveEmptyOptionalAsAResult() {
```
heshamMassoud: a unit test for this is better, since no is call is needed and no mocking is needed.

src/main/java/com/commercetools/sync/services/impl/CartDiscountServiceImpl.java
JudeNiroshan: Those 2 methods use can be placed as a future reference which uses caching CartDiscount IDs. Do you think this is not necessary?

src/main/java/com/commercetools/sync/services/CartDiscountService.java
JudeNiroshan: 5cdd3f9de578f817bda650bce4b2d122c41f0c03

src/main/java/com/commercetools/sync/services/impl/CartDiscountServiceImpl.java
JudeNiroshan: 5cdd3f9de578f817bda650bce4b2d122c41f0c03

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: 5cdd3f9de578f817bda650bce4b2d122c41f0c03

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: 5cdd3f9de578f817bda650bce4b2d122c41f0c03

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: 5cdd3f9de578f817bda650bce4b2d122c41f0c03

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: 5cdd3f9de578f817bda650bce4b2d122c41f0c03

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: a70a13c5d73471fe1e665701cadd9fbd9f295fef

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: a70a13c5d73471fe1e665701cadd9fbd9f295fef

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: 5cdd3f9

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: 5cdd3f9

src/integration-test/java/com/commercetools/sync/integration/services/impl/CartDiscountServiceImplIT.java
JudeNiroshan: f7a82da

ahmetoz:
@JudeNiroshan what is the reason to delete `fetchCachedCartDiscountId` ? 
5cdd3f9

src/test/java/com/commercetools/sync/services/impl/CartDiscountServiceImplTest.java
```diff
@@ -118,8 +137,128 @@ void createCartDiscount_WithEmptyCartDiscountKey_ShouldNotCreateCartDiscount() {
...
+        // test
+        final CompletionStage<CartDiscount> result =
+                cartDiscountService.updateCartDiscount(mockCartDiscount, updateActions);
+
+        // assertions
+        assertThat(result).hasFailedWithThrowableThat()
+                .isExactlyInstanceOf(InternalServerErrorException.class);
+    }
+
+
```
heshamMassoud: nitpick: redundant extra empty line

src/test/java/com/commercetools/sync/services/impl/CartDiscountServiceImplTest.java
```diff
@@ -31,6 +45,12 @@
 
     private String testCartDiscountKey = "ABC123";
 
+    private static final String PREDICATE_1 = "1 = 1";
```
heshamMassoud: seems like these values are only used in one test, then it's better to inline them where they are used.

src/test/java/com/commercetools/sync/services/impl/CartDiscountServiceImplTest.java
ahmetoz: 96337b1

src/test/java/com/commercetools/sync/services/impl/CartDiscountServiceImplTest.java
ahmetoz: 6f256ee

heshamMassoud: even better!

PR #398, opened 2019-05-29
394 fix resourceIdentifier vs reference comparison

### LINKED ISSUE #394: ResourceIdentifier to a Reference comparison is not working as expected
### Description
Comparing a `ResourceIdentifier` to a `Reference` which contain the same id field value should be render as equal. For example:
A resource identifier that looks as follows:
```
{
   "id": "foo",
   "typeId": "category"
}
```
and a reference that looks as follows: 
```
{
   "id": "foo",
   "typeId": "category"
}
```

Should be rendered as equal and not produce a generate an update action for it. 

### Current Behaviour
When comparing the both resource identifier on a draft and a reference on a resource, the update actions utils perceives them as *not* equal. And would build an update action as a result. Which is wrong behaviour.

### Expected Behaviour
They should be perceived as equal and not build an action since it's not required.

### Context
Most references on drafts are resource identifiers. Therefore, in most cases we will be comparing a reference on a resource and resource identifier on the new draft. Therefore, an update action is not needed when the id fields on both are equal.

heshamMassoud:
#### Summary
Fixes #394 

This PR addresses multiple things related to the same issue #394:

1. Add the [`CommonTypeUpdateActionUtils#buildUpdateActionForReferences`](https://github.com/commercetools/commercetools-sync-java/pull/398/files#diff-1e56c52206999ccd8db44d223e81bde7R58) util which is used to correctly compare any two resourceIdentifiers/references. 

2. When attaching resourceIdentifiers to drafts in tests, we now use `ResourceIdentifier#of` instead of `Category#referenceOfId` in case of categories for example. Since, the second one has type `Reference` and the first one has `ResourceIdentifier`. This way is the more common way to do it, since the draft accepts a `ResourceIdentifier` not the more specific `Reference`. The previously done why was also passing some tests positively, since when comparing a draft to a resource, the resource usually has a `Reference` not a `ResourceIdentifier`. So if the draft also had a `Reference` attached instead of a `ResourceIdentifier`, it means we'll be comparing a `Reference` to a `Reference` which basically wasn't catching the case of #394 so the tests were false positives. 

#### Relevant Issues
#394 

- Tests
    - [x] Unit 
    - [x] Integration
- [x] Documentation


src/integration-test/java/com/commercetools/sync/integration/inventories/InventorySyncIT.java
```diff
@@ -135,37 +136,6 @@ void sync_WithNewInventory_ShouldCreateInventory() {
         assertValues(oldInventoryAfterSync.get(), QUANTITY_ON_STOCK_2, EXPECTED_DELIVERY_2, RESTOCKABLE_IN_DAYS_2);
     }
 
-    @Test
```
ahmetoz: why did you need to delete this test method?

src/integration-test/java/com/commercetools/sync/integration/inventories/InventorySyncIT.java
```diff
@@ -135,37 +136,6 @@ void sync_WithNewInventory_ShouldCreateInventory() {
...
-        //Ensure old entry values after sync.
-        final Optional<InventoryEntry> oldInventoryAfterSync =
-            getInventoryEntryBySkuAndSupplyChannel(CTP_TARGET_CLIENT, SKU_1, supplyChannelReference);
-        assertThat(oldInventoryAfterSync).isPresent();
-        assertValues(oldInventoryAfterSync.get(), QUANTITY_ON_STOCK_2, EXPECTED_DELIVERY_2, RESTOCKABLE_IN_DAYS_2);
-        assertThat(oldInventoryAfterSync.get().getSupplyChannel().getId()).isEqualTo(supplyChannelReference.getId());
-    }
-
     @Test
     void sync_WithKeyToExistingSupplyChannelInPlaceOfReferenceId_ShouldUpdateInventory() {
```
ahmetoz: should we change the method name in here from ReferenceId to ResourceIdentifierWithId or Key?

src/integration-test/java/com/commercetools/sync/integration/inventories/InventorySyncIT.java
```diff
@@ -179,9 +149,7 @@ void sync_WithKeyToExistingSupplyChannelInPlaceOfReferenceId_ShouldUpdateInvento
          * Prepare InventoryEntryDraft of sku SKU_1 and reference to supply channel of key SUPPLY_CHANNEL_KEY_1.
          * Please note that the key is provided in place of Referenced id.
          */
-        final Reference<Channel> supplyChannelReference = Channel.referenceOfId(SUPPLY_CHANNEL_KEY_1);
-        assertThat(supplyChannelReference.getObj()).isNull();
-        assertThat(supplyChannelReference.getId()).isEqualTo(SUPPLY_CHANNEL_KEY_1);
+        final ResourceIdentifier<Channel> supplyChannelReference = ResourceIdentifier.ofId(SUPPLY_CHANNEL_KEY_1);
```
ahmetoz: why not `ResourceIdentifier.ofKey(SUPPLY_CHANNEL_KEY_1);` ?

src/main/java/com/commercetools/sync/categories/utils/CategoryReferenceReplacementUtils.java
```diff
@@ -38,8 +38,8 @@
             .map(category -> {
```
ahmetoz: maybe we need to update java docs in here from references to resourceidentifiers like 
`the categories to replace their {@link ResourceIdentifier} ids with keys...`
 wdyt ?

src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
```diff
@@ -52,6 +56,33 @@
...
+     * @param keyInReferenceSupplier the supplier to execute and return its result if the {@code reference} was
+     *                               expanded.
+     *
+     * @return returns the result of the {@code keyInReferenceSupplier} if the {@code reference} was expanded.
+     *         Otherwise, it returns the supplied reference as is.
+     */
+    @Nullable
+    public static <T> ResourceIdentifier<T> getResourceIdentifierWithKeyReplaced(
+        @Nullable final Reference<T> reference,
+        @Nonnull final Supplier<ResourceIdentifier<T>> keyInReferenceSupplier) {
```
ahmetoz: maybe it's better if we use ResourceIdentifier in java doc instead of the reference to make it different with getReferenceWithKeyReplaced's Javadoc. both javadocs looks same except the method name.

src/test/java/com/commercetools/sync/commons/utils/SyncUtilsTest.java
```diff
@@ -117,36 +119,63 @@ void batchElements_WithNegativeSize_ShouldReturnNoBatches() {
...
+    }
+
+    @Test
+    void getReferenceWithKeyReplaced_WithNonExpandedCategoryReference_ShouldReturnReferenceWithoutReplacedKey() {
         final String categoryUuid = UUID.randomUUID().toString();
         final Reference<Category> categoryReference = Reference.ofResourceTypeIdAndId(Category.referenceTypeId(),
             categoryUuid);
 
-        final Reference<Category> keyReplacedReference = replaceReferenceIdWithKey(categoryReference,
+        final ResourceIdentifier<Category> keyReplacedReference = getReferenceWithKeyReplaced(categoryReference,
```
ahmetoz: why not Reference, because we are testing `getReferenceWithKeyReplaced`

src/test/java/com/commercetools/sync/products/ProductSyncMockUtils.java
```diff
@@ -119,6 +121,27 @@ public static ProductDraftBuilder createProductDraftBuilder(@Nonnull final Strin
...
+
+        final List<Reference<Category>> references = categoryResourceIdentifiers
+            .stream()
+            .map(categoryResourceIdentifier -> Category.referenceOfId(categoryResourceIdentifier.getId()))
+            .collect(toList());
+        return createRandomCategoryOrderHints(references);
+    }
 
     /**
      * Given a {@link List} of {@link Category}, this method returns an instance of {@link CategoryOrderHints}
```
ahmetoz: ``Given a {@link Set} of {@link Category} {@link Reference}``

src/test/java/com/commercetools/sync/products/ProductSyncMockUtils.java
```diff
@@ -119,6 +121,27 @@ public static ProductDraftBuilder createProductDraftBuilder(@Nonnull final Strin
...
+     * {@link Set} as a key and a random categoryOrderHint which is a {@link String} containing a random double value
+     * between 0 and 1 (exclusive).
+     *
+     * <p>Note: The random double value is generated by the {@link ThreadLocalRandom#current()} nextDouble method.
+     *
+     * @param categoryResourceIdentifiers set of resource identifiers of categories to build categoryOrderHints for.
+     * @return an instance of {@link CategoryOrderHints} containing a categoryOrderHint for each category in the
+     *         supplied set of category resource identifiers.
+     */
+    public static CategoryOrderHints createRandomCategoryOrderHints(
```
ahmetoz: I could not see any usage of this method from idea.

src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
```diff
@@ -29,13 +31,57 @@
...
+     * @param oldResourceIdentifier the old resource identifier
+     * @param newResourceIdentifier the new resource identifier
+     * @param <T>                   the type of the old resource identifier
+     * @param <S>                   the type of the new resource identifier
+     * @return true or false depending if the resource identifiers have the same id.
+     */
+    public static <T extends ResourceIdentifier, S extends ResourceIdentifier> boolean areResourceIdentifiersEqual(
+        @Nullable final T oldResourceIdentifier, @Nullable final S newResourceIdentifier) {
+
+        final String oldId = ofNullable(oldResourceIdentifier).map(ResourceIdentifier::getId).orElse(null);
```
ahmetoz: what happens if ResourceIdentifiers are key-based ? 
https://docs.commercetools.com/http-api-types.html#resourceidentifier-by-key

src/integration-test/java/com/commercetools/sync/integration/commons/utils/CategoryITUtils.java
```diff
@@ -280,10 +281,11 @@ private static boolean hasADeletedAncestor(@Nonnull final Category category,
      *         {@link Category}.
      */
     @Nonnull
-    public static List<Reference<Category>> getReferencesWithKeys(@Nonnull final List<Category> categories) {
```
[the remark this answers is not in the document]
heshamMassoud: https://github.com/commercetools/commercetools-sync-java/pull/398/commits/6f488a919289755fc53cadb71ce1d4197ade3041

src/integration-test/java/com/commercetools/sync/integration/inventories/InventorySyncIT.java
heshamMassoud: because I noticed its a redundant test, it's testing nothing special compared to `sync_WithKeyToExistingSupplyChannelInPlaceOfReferenceId_ShouldUpdateInventory` for example.

src/integration-test/java/com/commercetools/sync/integration/inventories/InventorySyncIT.java
heshamMassoud: because this is the way we do reference resolution on the sync java library in general: We always provide the key in place of the id field (for consistency) since there still (will change in the future) exists references which are of type `Reference` not `ResourceIdentifier`. And in a `Reference` there is no `key` field, so we misuse the `id` field for it. 

However, this ought to be changed very soon.

src/integration-test/java/com/commercetools/sync/integration/inventories/InventorySyncIT.java
heshamMassoud: 602b8cb

src/main/java/com/commercetools/sync/categories/utils/CategoryReferenceReplacementUtils.java
heshamMassoud: 17ea123

src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
heshamMassoud: but actually the javadoc mentions the reference that is passed not the returned `ResourceIdentifier`, so I am not entirely sure which part of the javadoc should be changed

src/test/java/com/commercetools/sync/commons/utils/SyncUtilsTest.java
heshamMassoud: nice catch: 41e6358

src/test/java/com/commercetools/sync/products/ProductSyncMockUtils.java
heshamMassoud: why? It is still a list here :)

src/test/java/com/commercetools/sync/products/ProductSyncMockUtils.java
heshamMassoud: In this IT https://github.com/commercetools/commercetools-sync-java/pull/398/files#diff-4b6cdd198bf924968ff292406ff87580R138

src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
heshamMassoud: This is not how the sync java should be used (as of now). Check the documentations:

https://commercetools.github.io/commercetools-sync-java/doc/usage/PRODUCT_SYNC/#sync-list-of-product-drafts

src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
ahmetoz: oh you are right 👍

src/test/java/com/commercetools/sync/products/ProductSyncMockUtils.java
ahmetoz: probably I was reading the other one, methods have the same name. 🙈

src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
ahmetoz: hmm, I did not realize that before 👍

PR #404, opened 2019-06-13
#379 use cart discount keys instead of eng name

ahmetoz:
use cart discount keys instead of eng name


src/test/java/com/commercetools/sync/services/impl/CartDiscountServiceImplTest.java
```diff
@@ -31,6 +33,16 @@
 
     private String testCartDiscountKey = "ABC123";
 
+    public static final String PREDICATE_1 = "1 = 1";
```
heshamMassoud: All these fields look like they are not used.

heshamMassoud:
other than that, looks good.

PR #405, opened 2019-06-14
Only notify slack on changes for the failed builds not always.

heshamMassoud:
#### Summary
Only notify slack on changes for the failed builds not always.

- if the build was 'failing' and now is still 'failing' -> no notification
- if the build was 'failing' and now is 'succeeding' ->  notification

#### Description
Attempt to reduce noisy slack room


butenkor:
👍

PR #448, opened 2019-09-23
new CustomObjectService introduced for saving resolved references

JudeNiroshan:
#### Summary

New CustomObjectService has introduce to persist unresolved references in a custom object

#### Description
This PR is the start of introducing persistence using CTP custom objects for unresolved (yet) product references on other product attributes.

#### Todo

- Tests
    - [ ] Unit 
    - [ ] Integration
- [ ] Documentation

#### Hints for Review

Tests are under construction

### DIFF
--- a/src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
+++ b/src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
@@ -7,10 +7,8 @@ import java.util.List;
 
 public class NonResolvedReferencesCustomObject {
 
-    @Nonnull
     private String productKey;
 
-    @Nonnull
     private ProductDraft productDraft;
 
     private List<String> dependantProductKeys;
@@ -26,6 +24,9 @@ public class NonResolvedReferencesCustomObject {
         this.productKey = productKey;
         this.productDraft = productDraft;
         this.dependantProductKeys = dependantProductKeys;
+    }
+
+    public NonResolvedReferencesCustomObject() {
     }
 
     public String getProductKey() {
--- a/src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
+++ b/src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
@@ -1,0 +1,69 @@
+package com.commercetools.sync.commons.models;
+
+import io.sphere.sdk.models.Reference;
+import io.sphere.sdk.models.Resource;
+import io.sphere.sdk.products.ProductDraft;
+
+import javax.annotation.Nonnull;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+public class NonResolvedReferencesCustomObject implements Resource {
+
+    @Nonnull
+    private String productKey;
+
+    @Nonnull
+    private ProductDraft productDraft;
+
+    private List<String> dependantProductKeys;
+
+    public String getProductKey() {
+        return productKey;
+    }
+
+    public void setProductKey(String productKey) {
+        this.productKey = productKey;
+    }
+
+    public ProductDraft getProductDraft() {
+        return productDraft;
+    }
+
+    public void setProductDraft(ProductDraft productDraft) {
+        this.productDraft = productDraft;
+    }
+
+    public List<String> getDependantProductKeys() {
+        return dependantProductKeys;
+    }
+
+    public void setDependantProductKeys(List<String> dependantProductKeys) {
+        this.dependantProductKeys = dependantProductKeys;
+    }
+
+    @Override
+    public String getId() {
+        return null;
+    }
+
+    @Override
+    public Long getVersion() {
+        return null;
+    }
+
+    @Override
+    public ZonedDateTime getCreatedAt() {
+        return null;
+    }
+
+    @Override
+    public ZonedDateTime getLastModifiedAt() {
+        return null;
+    }
+
+    @Override
+    public Reference toReference() {
+        return null;
+    }
+}
--- a/src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
+++ b/src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
@@ -1,14 +1,11 @@
 package com.commercetools.sync.commons.models;
 
-import io.sphere.sdk.models.Reference;
-import io.sphere.sdk.models.Resource;
 import io.sphere.sdk.products.ProductDraft;
 
 import javax.annotation.Nonnull;
-import java.time.ZonedDateTime;
 import java.util.List;
 
-public class NonResolvedReferencesCustomObject implements Resource {
+public class NonResolvedReferencesCustomObject {
 
     @Nonnull
     private String productKey;
@@ -18,11 +15,24 @@ public class NonResolvedReferencesCustomObject implements Resource {
 
     private List<String> dependantProductKeys;
 
+    /**
+     * Object represent the value for the custom object that will hold unresolved references
+     * @param productKey product key that has non-resolved references
+     * @param productDraft product draft which has non-resolved references
+     * @param dependantProductKeys product keys of non-resolved references
+     */
+    public NonResolvedReferencesCustomObject(@Nonnull final String productKey, @Nonnull final ProductDraft productDraft,
+                                             final List<String> dependantProductKeys) {
+        this.productKey = productKey;
+        this.productDraft = productDraft;
+        this.dependantProductKeys = dependantProductKeys;
+    }
+
     public String getProductKey() {
         return productKey;
     }
 
-    public void setProductKey(String productKey) {
+    public void setProductKey(final String productKey) {
         this.productKey = productKey;
     }
 
@@ -30,7 +40,7 @@ public class NonResolvedReferencesCustomObject implements Resource {
         return productDraft;
     }
 
-    public void setProductDraft(ProductDraft productDraft) {
+    public void setProductDraft(final ProductDraft productDraft) {
         this.productDraft = productDraft;
     }
 
@@ -38,32 +48,8 @@ public class NonResolvedReferencesCustomObject implements Resource {
         return dependantProductKeys;
     }
 
-    public void setDependantProductKeys(List<String> dependantProductKeys) {
+    public void setDependantProductKeys(final List<String> dependantProductKeys) {
         this.dependantProductKeys = dependantProductKeys;
     }
 
-    @Override
-    public String getId() {
-        return null;
-    }
-
-    @Override
-    public Long getVersion() {
-        return null;
-    }
-
-    @Override
-    public ZonedDateTime getCreatedAt() {
-        return null;
-    }
-
-    @Override
-    public ZonedDateTime getLastModifiedAt() {
-        return null;
-    }
-
-    @Override
-    public Reference toReference() {
-        return null;
-    }
 }
--- a/src/main/java/com/commercetools/sync/services/CustomObjectService.java
+++ b/src/main/java/com/commercetools/sync/services/CustomObjectService.java
@@ -24,8 +24,8 @@ public interface CustomObjectService {
      * {@link Optional} that contains the matching {@link CustomObject} if exists, otherwise empty.
      */
     @Nonnull
-    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>>
-    fetchCustomObject(@Nullable final String key);
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> fetchCustomObject(
+            @Nullable final String key);
 
     /**
      * Given a resource of type {@link CustomObject}, this method attempts to create a
@@ -43,11 +43,11 @@ public interface CustomObjectService {
      *
      * @param customObject the resource to create/update a resource based off of.
      * @return a {@link CompletionStage} containing an optional with the created resource if successful otherwise an
-     * empty optional.
+     *     empty optional.
      */
     @Nonnull
-    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> createOrUpdateCustomObject
-    (@Nonnull final CustomObjectDraft<NonResolvedReferencesCustomObject> customObject);
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> createOrUpdateCustomObject(
+            @Nonnull final CustomObjectDraft<NonResolvedReferencesCustomObject> customObject);
 
     /**
      * Given a resource of type {@link CustomObject}, this method attempts to delete a
@@ -65,10 +65,10 @@ public interface CustomObjectService {
      *
      * @param customObject the resource to delete a resource based off of.
      * @return a {@link CompletionStage} containing an optional with the deleted resource if successful otherwise an
-     * empty optional.
+     *     empty optional.
      */
     @Nonnull
-    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> deleteCustomObject
-    (@Nonnull final CustomObject<NonResolvedReferencesCustomObject> customObject);
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> deleteCustomObject(
+            @Nonnull final CustomObject<NonResolvedReferencesCustomObject> customObject);
 
 }
--- a/src/main/java/com/commercetools/sync/services/CustomObjectService.java
+++ b/src/main/java/com/commercetools/sync/services/CustomObjectService.java
@@ -1,0 +1,74 @@
+package com.commercetools.sync.services;
+
+
+import com.commercetools.sync.commons.models.NonResolvedReferencesCustomObject;
+import io.sphere.sdk.client.SphereClient;
+import io.sphere.sdk.customobjects.CustomObject;
+import io.sphere.sdk.customobjects.CustomObjectDraft;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+public interface CustomObjectService {
+
+    /**
+     * Given a custom object key, this method fetches a custom object that matches this given key in the CTP project
+     * defined in a potentially injected {@link SphereClient}. If there is no matching custom object an empty
+     * {@link Optional} will be returned in the returned future. A mapping of the key to the id of the fetched
+     * custom object is persisted in an in -memory map.
+     *
+     * @param key the key of the custom object to fetch.
+     * @return {@link CompletionStage}&lt;{@link Optional}&gt; in which the result of it's completion contains an
+     * {@link Optional} that contains the matching {@link CustomObject} if exists, otherwise empty.
+     */
+    @Nonnull
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>>
+    fetchCustomObject(@Nullable final String key);
+
+    /**
+     * Given a resource of type {@link CustomObject}, this method attempts to create a
+     * {@link CustomObject} based on it in the CTP project defined by the sync options.
+     *
+     * <p>A completion stage containing an empty option and the error callback will be triggered in those cases:
+     * <ul>
+     *     <li>the draft has a blank key</li>
+     *     <li>the create request fails on CTP</li>
+     * </ul>
+     *
+     * <p>On the other hand, if the resource gets created/updated successfully on CTP, then the created resource
+     * of {@link CompletionStage} in which the result of it's completion
+     * contains an instance {@link Optional} of the resource which was created/updated.
+     *
+     * @param customObject the resource to create/update a resource based off of.
+     * @return a {@link CompletionStage} containing an optional with the created resource if successful otherwise an
+     * empty optional.
+     */
+    @Nonnull
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> createOrUpdateCustomObject
+    (@Nonnull final CustomObjectDraft<NonResolvedReferencesCustomObject> customObject);
+
+    /**
+     * Given a resource of type {@link CustomObject}, this method attempts to delete a
+     * {@link CustomObject} based on it in the CTP project defined by the sync options.
+     *
+     * <p>A completion stage containing an empty option and the error callback will be triggered in those cases:
+     * <ul>
+     *     <li>the draft has a blank key</li>
+     *     <li>the delete request fails on CTP</li>
+     * </ul>
+     *
+     * <p>On the other hand, if the resource gets deleted successfully on CTP, then the created resource
+     * of {@link CompletionStage} in which the result of it's completion
+     * contains an instance {@link Optional} of the resource which was deleted.
+     *
+     * @param customObject the resource to delete a resource based off of.
+     * @return a {@link CompletionStage} containing an optional with the deleted resource if successful otherwise an
+     * empty optional.
+     */
+    @Nonnull
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> deleteCustomObject
+    (@Nonnull final CustomObject<NonResolvedReferencesCustomObject> customObject);
+
+}
--- a/src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
+++ b/src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
@@ -32,21 +32,21 @@ public class CustomObjectServiceImpl
 
     private static final String CREATE_FAILED = "Failed to create draft with key: '%s'. Reason: %s";
 
-    public CustomObjectServiceImpl(BaseSyncOptions baseSyncOptions) {
+    public CustomObjectServiceImpl(final BaseSyncOptions baseSyncOptions) {
         this.syncOptions = baseSyncOptions;
     }
 
     @Nonnull
     @Override
     public CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>>
-    fetchCustomObject(@Nullable String key) {
+        fetchCustomObject(@Nullable final String key) {
 
         if (isBlank(key)) {
             return CompletableFuture.completedFuture(Optional.empty());
         }
 
-        final CustomObjectQuery<NonResolvedReferencesCustomObject> customObjectQuery = CustomObjectQuery.
-                of(NonResolvedReferencesCustomObject.class)
+        final CustomObjectQuery<NonResolvedReferencesCustomObject> customObjectQuery = CustomObjectQuery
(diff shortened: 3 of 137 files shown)


src/main/java/com/commercetools/sync/services/CustomObjectService.java
```diff
@@ -0,0 +1,74 @@
...
+import io.sphere.sdk.client.SphereClient;
+import io.sphere.sdk.customobjects.CustomObject;
+import io.sphere.sdk.customobjects.CustomObjectDraft;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+public interface CustomObjectService {
```
heshamMassoud: I would name it e.g. `LazyResolutionService`. The service itself is about keeping track of `productDrafts` with references that cannot be resolved yet. So the service can be used to lazily create/update those drafts as soon as the references can be resolved.

So the fact that the inner implementation uses CTP `CustomObject`s is an implementation specific/detail. It is not the main focus of the service.

src/main/java/com/commercetools/sync/services/CustomObjectService.java
```diff
@@ -0,0 +1,74 @@
...
+     * defined in a potentially injected {@link SphereClient}. If there is no matching custom object an empty
+     * {@link Optional} will be returned in the returned future. A mapping of the key to the id of the fetched
+     * custom object is persisted in an in -memory map.
+     *
+     * @param key the key of the custom object to fetch.
+     * @return {@link CompletionStage}&lt;{@link Optional}&gt; in which the result of it's completion contains an
+     * {@link Optional} that contains the matching {@link CustomObject} if exists, otherwise empty.
+     */
+    @Nonnull
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> fetchCustomObject(
```
heshamMassoud: `fetch` is enough

src/main/java/com/commercetools/sync/services/CustomObjectService.java
```diff
@@ -0,0 +1,74 @@
...
+     * <p>On the other hand, if the resource gets created/updated successfully on CTP, then the created resource
+     * of {@link CompletionStage} in which the result of it's completion
+     * contains an instance {@link Optional} of the resource which was created/updated.
+     *
+     * @param customObject the resource to create/update a resource based off of.
+     * @return a {@link CompletionStage} containing an optional with the created resource if successful otherwise an
+     *     empty optional.
+     */
+    @Nonnull
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> createOrUpdateCustomObject(
```
heshamMassoud: I would name it `save`

src/main/java/com/commercetools/sync/services/CustomObjectService.java
```diff
@@ -0,0 +1,74 @@
...
+            @Nonnull final CustomObjectDraft<NonResolvedReferencesCustomObject> customObject);
```
heshamMassoud: It should take the `NonResolvedReferencesCustomObject` POJO as a parameter. Again, the usage of CustomObject is an implementation detail not the main purpose of the service. Maybe at some point in the future, we use sth completley different than CTP customObjects for persistence.

src/main/java/com/commercetools/sync/services/CustomObjectService.java
```diff
@@ -0,0 +1,74 @@
...
+     * of {@link CompletionStage} in which the result of it's completion
+     * contains an instance {@link Optional} of the resource which was deleted.
+     *
+     * @param customObject the resource to delete a resource based off of.
+     * @return a {@link CompletionStage} containing an optional with the deleted resource if successful otherwise an
+     *     empty optional.
+     */
+    @Nonnull
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> deleteCustomObject(
+            @Nonnull final CustomObject<NonResolvedReferencesCustomObject> customObject);
```
heshamMassoud: why not just pass key here?

src/main/java/com/commercetools/sync/services/CustomObjectService.java
```diff
@@ -0,0 +1,74 @@
...
+     * <p>On the other hand, if the resource gets deleted successfully on CTP, then the created resource
+     * of {@link CompletionStage} in which the result of it's completion
+     * contains an instance {@link Optional} of the resource which was deleted.
+     *
+     * @param customObject the resource to delete a resource based off of.
+     * @return a {@link CompletionStage} containing an optional with the deleted resource if successful otherwise an
+     *     empty optional.
+     */
+    @Nonnull
+    CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>> deleteCustomObject(
```
heshamMassoud: `delete` is enough

src/main/java/com/commercetools/sync/services/CustomObjectService.java
```diff
@@ -0,0 +1,74 @@
...
+import io.sphere.sdk.customobjects.CustomObjectDraft;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+public interface CustomObjectService {
+
+    /**
```
heshamMassoud: Please adjust the Javadocs accordingly after addressing the review. Keep in mind, there are a lot of implementation detail that should not be in the Javadoc in the `interface` which does not know about the implementation (there could be multiple implementations)

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static java.util.Collections.singleton;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+public class CustomObjectServiceImpl
+        implements CustomObjectService {
+
+    private final BaseSyncOptions syncOptions;
```
heshamMassoud: We know for sure that it is `ProductSyncOptions`

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+                .of(NonResolvedReferencesCustomObject.class)
+                .withPredicates(buildCustomObjectKeysQueryPredicate(singleton(key)));
+
+        return syncOptions
+                .getCtpClient()
+                .execute(customObjectQuery)
+                .thenApply(PagedResult::head);
+    }
+
+    private QueryPredicate<CustomObject<NonResolvedReferencesCustomObject>> buildCustomObjectKeysQueryPredicate(
```
heshamMassoud: Why do we need this method that takes set of keys? You are always calling this with `singleton(key)`!

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    public CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>>
+        fetchCustomObject(@Nullable final String key) {
+
+        if (isBlank(key)) {
+            return CompletableFuture.completedFuture(Optional.empty());
+        }
+
+        final CustomObjectQuery<NonResolvedReferencesCustomObject> customObjectQuery = CustomObjectQuery
+                .of(NonResolvedReferencesCustomObject.class)
+                .withPredicates(buildCustomObjectKeysQueryPredicate(singleton(key)));
```
heshamMassoud: Consider using `CustomObjectByKeyGet`
```java
        return syncOptions
            .getCtpClient()
            .execute(CustomObjectByKeyGet
                .of(CUSTOM_OBJECT_CONTAINER_KEY, key, NonResolvedReferencesCustomObject.class))
            .thenApply(Optional::ofNullable);
```

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+
+import static java.lang.String.format;
+import static java.util.Collections.singleton;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+public class CustomObjectServiceImpl
+        implements CustomObjectService {
+
+    private final BaseSyncOptions syncOptions;
+
```
heshamMassoud: Please use a constant `CustomObject` container key that will always be used for all the custom objects created by this service. It should be a name that is understood from a user that doesn't know about this service. For example:

```java
private static final String CUSTOM_OBJECT_CONTAINER_KEY = "commercetools-sync-java.LazyResolutionService";
```

heshamMassoud: then when saving, deleting or fetching in the service, this container key should always be used.

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>>
+        deleteCustomObject(@Nonnull final CustomObject<NonResolvedReferencesCustomObject> customObject) {
+
+        return syncOptions
+                .getCtpClient()
+                .execute(CustomObjectDeleteCommand.of(customObject, NonResolvedReferencesCustomObject.class))
```
heshamMassoud: for better user experience delete by key:

```java
        return syncOptions
            .getCtpClient()
            .execute(CustomObjectDeleteCommand
                .of(CUSTOM_OBJECT_CONTAINER_KEY, key, NonResolvedReferencesCustomObject.class))
```

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+                .handle((resource, exception) -> {
+                    if (exception == null) {
+                        return Optional.of(resource);
+                    } else {
+                        syncOptions.applyErrorCallback(
+                                format(CREATE_FAILED, customObject.getKey(), exception.getMessage()), exception);
```
heshamMassoud: `CREATE_FAILED` is wrong here.

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        String keysQueryString = keysSurroundedWithDoubleQuotes.toString();
+        // Strip square brackets from list string. For example: ["key1", "key2"] -> "key1", "key2"
+        keysQueryString = keysQueryString.substring(1, keysQueryString.length() - 1);
+        return QueryPredicate.of(format("key in (%s)", keysQueryString));
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Optional<CustomObject<NonResolvedReferencesCustomObject>>>
+        createOrUpdateCustomObject(@Nonnull final CustomObjectDraft<NonResolvedReferencesCustomObject>
```
heshamMassoud: we should pass the `NonResolvedReferencesCustomObject` instance here and the service body should take care of how the custom object is created not the outside (out of this service) user.


````java
        final CustomObjectDraft<ProductWithUnResolvedProductReferences> customObjectDraft = CustomObjectDraft
            .ofUnversionedUpsert(
                CUSTOM_OBJECT_CONTAINER_KEY,
                draftWithUnresolvedReferences.getProductDraft().getKey(),
                draftWithUnresolvedReferences,
                ProductWithUnResolvedProductReferences.class);

        return syncOptions
            .getCtpClient()
            .execute(CustomObjectUpsertCommand.of(customObjectDraft))
            .handle((resource, exception) -> {
                if (exception == null) {
                    return Optional.of(resource);
                } else {
                    syncOptions.applyErrorCallback(
                        format(SAVE_FAILED, customObjectDraft.getKey(), exception.getMessage()), exception);
                    return Optional.empty();
                }
            });
````

File: src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java [hunk 15]
```diff
@@ -0,0 +1,56 @@
+package com.commercetools.sync.commons.models;
+
+import io.sphere.sdk.products.ProductDraft;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+public class NonResolvedReferencesCustomObject {
```
heshamMassoud: I would name it `ProductWithUnResolvedProductReferences`. This class has/knows nothing about CTP CustomObject. We should loosen this coupling of concepts.

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private String productKey;
```
heshamMassoud: Why do we need this field?

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private ProductDraft productDraft;
+
+    private List<String> dependantProductKeys;
```
heshamMassoud: `dependant` is a bit confusing who is dependant on who?

I would name it: `missingReferencedProductKeys`

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Object represent the value for the custom object that will hold unresolved references
+     * @param productKey product key that has non-resolved references
+     * @param productDraft product draft which has non-resolved references
+     * @param dependantProductKeys product keys of non-resolved references
+     */
+    public NonResolvedReferencesCustomObject(@Nonnull final String productKey, @Nonnull final ProductDraft productDraft,
+                                             final List<String> dependantProductKeys) {
```
heshamMassoud: We can we accept null here? should be `@Nonnull`

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public NonResolvedReferencesCustomObject() {
+    }
+
+    public String getProductKey() {
+        return productKey;
+    }
+
+    public void setProductKey(final String productKey) {
```
heshamMassoud: `@Nonnull`

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public void setProductDraft(final ProductDraft productDraft) {
+        this.productDraft = productDraft;
+    }
+
+    public List<String> getDependantProductKeys() {
+        return dependantProductKeys;
+    }
+
+    public void setDependantProductKeys(final List<String> dependantProductKeys) {
```
heshamMassoud: `@Nonnull`

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public void setProductKey(final String productKey) {
+        this.productKey = productKey;
+    }
+
+    public ProductDraft getProductDraft() {
+        return productDraft;
+    }
+
+    public void setProductDraft(final ProductDraft productDraft) {
```
heshamMassoud: `@Nonnull`

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
heshamMassoud: also `final` class if we know that we won't extend it.

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
heshamMassoud: better `set` since it should only contain unique keys?

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: yes, It makes sense :)

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: In this case, we have to create the CustomObjectDraft inside the implementation. Then, how do we go with defining the CO `container` and `key`?

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: There is no builder method in `CustomObjectDeleteCommand` that only accepts `key`. We have to decide whether we want to persist *versioned* custom object or not. If we need *versioned* custom objects, we need to always take care of **versionId** as well. Currently we do not care about *versioning* the custom objects.

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: yes, this method seems to be unnecessary.

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: Then again we have to answer the same question, what do we consider to be the `container` of the COs we create through this service?

JudeNiroshan: Please ignore the above comment. I have the answer in below 👍

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: `draftWithUnresolvedReferences.getProductDraft().getKey()` or `draftWithUnresolvedReferences.getProductKey()` would be the same right? Aren't they?

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: 63ff515cdd97b4350b6cc8d1b9fee25c6b532345

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: c031550e391deee0adff7c5299d099062963158a

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: 4cad76f0f8f148941c375a74f9d5b669439032d6

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: a081501113b7ca7fa8c716a7491c56a081edb0f0

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: aa512c4722b4ae605af634c22f3f7618f4fba4ff

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: 517f5b392446c4cd021757488a341d509aa4dfba

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: f3a06bbe350aa6fc0f780a15c05b455d580e7491

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: 9d4151a7bce2e91d561be72574ac917aa961d204

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: 2ed582ee415adbad4cb71228089cb5f21fd9568a

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: a081501113b7ca7fa8c716a7491c56a081edb0f0

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: 2ed582ee415adbad4cb71228089cb5f21fd9568a

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: aa512c4722b4ae605af634c22f3f7618f4fba4ff

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: 7aab8777db23834065e24ebdf96cf42e286ba22e

src/main/java/com/commercetools/sync/services/impl/CustomObjectServiceImpl.java
JudeNiroshan: Please ignore the above comment. :)

JudeNiroshan: a081501113b7ca7fa8c716a7491c56a081edb0f0

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: 723572d6bbf71ee61b733d15542162ed09bea278

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: b18e91fd54a0dbf32f25970b1f82f0577d88eeb0

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: dd6d0e61e82fb516b8e36b4aa9f2106f6dab0b55

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: b2bfdb115c99e4c56a775cb0dd143f44ec2dd250

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: We can ignore this; as we have already removed this particular variable

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: b2bfdb115c99e4c56a775cb0dd143f44ec2dd250

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: b4ecc8c49c15182fea9dffd364df66f172dbf6a1

src/main/java/com/commercetools/sync/services/CustomObjectService.java
heshamMassoud: please also check: https://github.com/commercetools/commercetools-sync-java/commit/c031550e391deee0adff7c5299d099062963158a#r35291991

src/main/java/com/commercetools/sync/commons/models/NonResolvedReferencesCustomObject.java
JudeNiroshan: 831f16069df3110e04543ecebebccafc293a4899

src/main/java/com/commercetools/sync/services/CustomObjectService.java
heshamMassoud: @JudeNiroshan not sure what you mean by your first comment. Is it still valid?

JudeNiroshan:
> Where is this constant defined? 🤔 🤔 🤔

It is being defined inside `LazyResolutionServiceImpl`.

src/main/java/com/commercetools/sync/services/CustomObjectService.java
heshamMassoud: Please also check: please also check: https://github.com/commercetools/commercetools-sync-java/commit/c031550e391deee0adff7c5299d099062963158a#r35291991

heshamMassoud:
@JudeNiroshan Looks good! thanks. The only thing left is I think the unit tests are missing in this PR for both new classes.

src/main/java/com/commercetools/sync/services/CustomObjectService.java
JudeNiroshan: Currently we do not support versioning for CustomObjects that we created through this service.

heshamMassoud:
Hi @JudeNiroshan. I have merged the changes here to: https://github.com/commercetools/commercetools-sync-java/pull/452

I would appreciate it if you add unit and integration tests covering the cases that are not covered in a branch off of: https://github.com/commercetools/commercetools-sync-java/pull/452

I will close this PR for now.

Issue #447, opened 2019-09-23
Sync products in any order even if they have self-references


heshamMassoud:
Since a productDraft X could be supplied in with an attribute referencing productDraft Y. It could be that Y is not supplied before X, which means the sync could fail creating/updating X. It could also be that Y is not supplied at all in this batch but at a later batch.

The product sync would fail updating/creating this draft immediately, since the reference resolver component would fail to resolve such reference since it does not exist (yet). However, the library should keep track of such "referencing" drafts like X and persist them in storage (CTP customObjects in this case) to keep them and create/update them accordingly when the references are there.

lojzatran:
Description for the issue is missing @heshamMassoud

heshamMassoud:
> Description for the issue is missing @heshamMassoud

now it's not 😛

heshamMassoud:
released in https://github.com/commercetools/commercetools-sync-java/releases/tag/1.6.0

PR #454, opened 2019-10-07
#447: unit tests for LazyResolutionService

JudeNiroshan:
#### Summary

Unit tests for LazyResolutionService

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [x] Documentation

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
@@ -38,7 +38,7 @@ class UnresolvedReferencesServiceImplIT {
 
 
     @AfterEach
-    void setup() {
+    void tearDown() {
         deleteWaitingToBeResolvedCustomObjects(CTP_TARGET_CLIENT);
     }
 
--- a/src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
@@ -9,6 +9,7 @@ import io.sphere.sdk.customobjects.CustomObject;
 import io.sphere.sdk.customobjects.queries.CustomObjectByKeyGet;
 import io.sphere.sdk.json.SphereJsonUtils;
 import io.sphere.sdk.products.ProductDraft;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -36,7 +37,7 @@ class UnresolvedReferencesServiceImplIT {
         "commercetools-sync-java.UnresolvedReferencesService.productDrafts";
 
 
-    @BeforeEach
+    @AfterEach
     void setup() {
         deleteWaitingToBeResolvedCustomObjects(CTP_TARGET_CLIENT);
     }
@@ -90,6 +91,44 @@ class UnresolvedReferencesServiceImplIT {
             .join();
 
         assertThat(createdCustomObject.getValue()).isEqualTo(productDraftWithUnresolvedRefs);
+    }
+
+    @Test
+    void save_ExistingProductDraftWithoutException_overwriteOldCustomObjectValue() {
+        // preparation
+        final ProductDraft productDraft =
+            SphereJsonUtils.readObjectFromResource(PRODUCT_KEY_1_RESOURCE_PATH, ProductDraft.class);
+
+        final WaitingToBeResolved productDraftWithUnresolvedRefs =
+                new WaitingToBeResolved(productDraft, asSet("foo", "bar"));
+
+        unresolvedReferencesService
+                .save(productDraftWithUnresolvedRefs)
+                .toCompletableFuture()
+                .join();
+
+        final WaitingToBeResolved productDraftWithUnresolvedNewRefs =
+                new WaitingToBeResolved(productDraft, asSet("foo123", "bar123"));
+
+        // test
+        final Optional<WaitingToBeResolved> latestResult = unresolvedReferencesService
+            .save(productDraftWithUnresolvedNewRefs)
+            .toCompletableFuture()
+            .join();
+
+
+        // assertions
+        assertThat(latestResult).hasValueSatisfying(waitingToBeResolved ->
+            assertThat(waitingToBeResolved.getProductDraft()).isEqualTo(productDraft));
+
+        final CustomObjectByKeyGet<WaitingToBeResolved> customObjectByKeyGet = CustomObjectByKeyGet
+            .of(CUSTOM_OBJECT_CONTAINER_KEY, productDraft.getKey(), WaitingToBeResolved.class);
+        final CustomObject<WaitingToBeResolved> createdCustomObject = CTP_TARGET_CLIENT
+            .execute(customObjectByKeyGet)
+            .toCompletableFuture()
+            .join();
+
+        assertThat(createdCustomObject.getValue()).isEqualTo(productDraftWithUnresolvedNewRefs);
     }
 
     @Test
--- a/src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
@@ -36,6 +36,13 @@ class UnresolvedReferencesServiceImplIT {
     private static final String CUSTOM_OBJECT_CONTAINER_KEY =
         "commercetools-sync-java.UnresolvedReferencesService.productDrafts";
 
+
+
+    @AfterEach
+    void tearDown() {
+        deleteWaitingToBeResolvedCustomObjects(CTP_TARGET_CLIENT);
+    }
+
     @BeforeEach
     void setupTest() {
         errorCallBackMessages = new ArrayList<>();
@@ -54,42 +61,8 @@ class UnresolvedReferencesServiceImplIT {
         unresolvedReferencesService = new UnresolvedReferencesServiceImpl(productSyncOptions);
     }
 
-    @AfterEach
-    void tearDown() {
-        deleteWaitingToBeResolvedCustomObjects(CTP_TARGET_CLIENT);
-    }
-
-
     @Test
-    void fetchCustomObject_shouldReturnCorrectCustomObject() {
-        // preparation
-        final ProductDraft productDraft =
-            SphereJsonUtils.readObjectFromResource(PRODUCT_KEY_1_RESOURCE_PATH, ProductDraft.class);
-
-        final WaitingToBeResolved productDraftWithUnresolvedRefs =
-            new WaitingToBeResolved(productDraft, asSet("foo", "bar"));
-
-        unresolvedReferencesService
-            .save(productDraftWithUnresolvedRefs)
-            .toCompletableFuture()
-            .join();
-
-        // test
-        final Set<WaitingToBeResolved> result = unresolvedReferencesService
-            .fetch(singleton(productDraft.getKey()))
-            .toCompletableFuture()
-            .join();
-
-        // assertions
-        assertThat(result).hasOnlyOneElementSatisfying(waitingToBeResolved ->
-            assertThat(waitingToBeResolved).isEqualTo(productDraftWithUnresolvedRefs));
-        assertThat(errorCallBackMessages).isEmpty();
-        assertThat(errorCallBackExceptions).isEmpty();
-        assertThat(warningCallBackMessages).isEmpty();
-    }
-
-    @Test
-    void save_WithoutException_createsNewCustomObject() {
+    void saveFetchAndDelete_WithoutExceptions_shouldWorkCorrectly() {
         // preparation
         final ProductDraft productDraft =
             SphereJsonUtils.readObjectFromResource(PRODUCT_KEY_1_RESOURCE_PATH, ProductDraft.class);
@@ -107,17 +80,30 @@ class UnresolvedReferencesServiceImplIT {
         assertThat(result).hasValueSatisfying(waitingToBeResolved ->
             assertThat(waitingToBeResolved.getProductDraft()).isEqualTo(productDraft));
 
-        final CustomObjectByKeyGet<WaitingToBeResolved> customObjectByKeyGet = CustomObjectByKeyGet
-            .of(CUSTOM_OBJECT_CONTAINER_KEY, productDraft.getKey(), WaitingToBeResolved.class);
-        final CustomObject<WaitingToBeResolved> createdCustomObject = CTP_TARGET_CLIENT
-            .execute(customObjectByKeyGet)
+        // test
+        final Set<WaitingToBeResolved> waitingDrafts = unresolvedReferencesService
+            .fetch(singleton(productDraft.getKey()))
             .toCompletableFuture()
             .join();
 
-        assertThat(createdCustomObject.getValue()).isEqualTo(productDraftWithUnresolvedRefs);
+        // assertions
+        assertThat(waitingDrafts).containsExactly(productDraftWithUnresolvedRefs);
+
+        // test
+        final Optional<WaitingToBeResolved> deletionResult = unresolvedReferencesService
+            .delete(productDraft.getKey())
+            .toCompletableFuture()
+            .join();
+
+        // assertions
+        assertThat(deletionResult).hasValueSatisfying(waitingToBeResolved ->
+            assertThat(waitingToBeResolved.getProductDraft()).isEqualTo(productDraft));
+
         assertThat(errorCallBackMessages).isEmpty();
-        assertThat(errorCallBackExceptions).isEmpty();
         assertThat(warningCallBackMessages).isEmpty();
+        assertThat(errorCallBackExceptions).isEmpty();
+
+
     }
 
     @Test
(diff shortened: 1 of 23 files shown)


src/main/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImpl.java
```diff
@@ -60,6 +63,13 @@ public UnresolvedReferencesServiceImpl(@Nonnull final ProductSyncOptions baseSyn
     @Nonnull
     @Override
     public CompletionStage<Optional<WaitingToBeResolved>> save(@Nonnull final WaitingToBeResolved draft) {
+
+        if (isBlank(draft.getProductDraft().getKey())) {
+            syncOptions.applyErrorCallback(format(SAVE_FAILED, draft.getProductDraft().getKey(),
+                    "Draft key is blank!"));
```
heshamMassoud: 1. It's nice to have but we don't need it at the moment (https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) since we know we always call it with a non-blank key. So there is no need to add the logic and test it and so on. 
2. "Draft key is blank!" is not very helpful for the user in this case, some information in the draft like the value maybe could be useful for the user to know more information about this case.

heshamMassoud:
Unit tests are also missing for the `WaitingToBeResolved` class.

src/main/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImpl.java
```diff
@@ -37,6 +39,7 @@ public UnresolvedReferencesServiceImpl(@Nonnull final ProductSyncOptions baseSyn
     @Nonnull
     @Override
     public CompletionStage<Set<WaitingToBeResolved>> fetch(@Nonnull final Set<String> keys) {
+        keys.remove(EMPTY);
```
heshamMassoud: - It's not a good practice to mutate in the user input.
- What is the purpose of removing `EMPTY` at all? We should not be expecting it to be passed.

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+        assertThat(errorMessages).hasSize(1);
+        assertThat(errorExceptions).hasSize(1);
+        assertThat(errorMessages.get(0))
+                .isEqualTo("Failed to save CustomObject with key: 'null'. Reason: Draft key is blank!");
+    }
+
+    @Test
+    void fetch_WithEmptyKeySet_ShouldReturnEmptySet() {
+        Set<String> keys = new HashSet<>();
+        keys.add("");
```
heshamMassoud: why do you add it `""`, why not test with an originally empty set?

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+        final Set<WaitingToBeResolved> beResolvedOptional = service
+                .fetch(keys).toCompletableFuture().join();
+
+        assertThat(beResolvedOptional).isEmpty();
+    }
+
+
+    @Test
+    void fetch_WithSuccessfulMockCtpResponse_ShouldReturnMock() {
```
heshamMassoud: what about fetch with unsuccessful?

heshamMassoud: I find tests to read when the bodies are separated with comments:

```java
// preparation
...
// test
...
// assertion
...
```
What do you think about this?

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+        final CustomObject mock = mock(CustomObject.class);
+        when(mock.getContainer()).thenReturn("commercetools-sync-java.UnresolvedReferencesService.productDrafts");
+
+        final ProductDraft productDraft = mock(ProductDraft.class);
+        when(productDraft.getKey()).thenReturn("product-draft-key");
+        final Set<String> missingRefs = new HashSet<>();
+        missingRefs.add("test-ref");
+        WaitingToBeResolved valueObj = new WaitingToBeResolved(productDraft, missingRefs);
```
heshamMassoud: can be simplified to:

```java
final WaitingToBeResolved waitingDraft = new WaitingToBeResolved(productDraft, singleton("test-ref"));
```

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+        when(mock.getValue()).thenReturn(valueObj);
+
+        Set<String> keys = new HashSet<>();
+        keys.add("product-draft-key");
```
heshamMassoud: You can inline this directly on line 158 with just

```java
final Set<WaitingToBeResolved> toBeResolvedOptional = service.fetch(singleton("product-draft-key")).toCompletableFuture().join();
```

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .fetch(keys).toCompletableFuture().join();
+
+        assertThat(beResolvedOptional).isEmpty();
+    }
+
+
+    @Test
+    void fetch_WithSuccessfulMockCtpResponse_ShouldReturnMock() {
+        final CustomObject mock = mock(CustomObject.class);
+        when(mock.getContainer()).thenReturn("commercetools-sync-java.UnresolvedReferencesService.productDrafts");
```
heshamMassoud: this is not needed, right?

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+        missingRefs.add("test-ref");
+        WaitingToBeResolved valueObj = new WaitingToBeResolved(productDraft, missingRefs);
+        when(mock.getValue()).thenReturn(valueObj);
+
+        Set<String> keys = new HashSet<>();
+        keys.add("product-draft-key");
+
+        PagedQueryResult pagedQueryResult = mock(PagedQueryResult.class);
+        when(pagedQueryResult.getCount()).thenReturn(1L);
+        when(pagedQueryResult.getOffset()).thenReturn(0L);
```
heshamMassoud: is this needed?

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+        when(productDraft.getKey()).thenReturn("product-draft-key");
+        final Set<String> missingRefs = new HashSet<>();
+        missingRefs.add("test-ref");
+        WaitingToBeResolved valueObj = new WaitingToBeResolved(productDraft, missingRefs);
+        when(mock.getValue()).thenReturn(valueObj);
+
+        Set<String> keys = new HashSet<>();
+        keys.add("product-draft-key");
+
+        PagedQueryResult pagedQueryResult = mock(PagedQueryResult.class);
```
heshamMassoud: `final`

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+        when(productSyncOptions.getCtpClient().execute(any())).thenReturn(completedFuture(pagedQueryResult));
+
+        final Set<WaitingToBeResolved> toBeResolvedOptional = service.fetch(keys).toCompletableFuture().join();
+
+        assertThat(toBeResolvedOptional).isNotEmpty();
+        assertThat(toBeResolvedOptional).containsOnly(valueObj);
+    }
+
+    @Test
+    void delete_WithUnsuccessfulMockCtpResponse_ShouldReturnProperException() {
```
heshamMassoud: What about successful delete?

lojzatran:
Where is `LazyResolutionService`? I don't see any class with this name.

heshamMassoud:
> Where is `LazyResolutionService`? I don't see any class with this name.

@lojzatran `UnresolvedReferencesService` used to be `LazyResolutionService`

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+                    errorMessages.add(errorMessage);
+                    errorExceptions.add(errorException);
+                })
+                .build();
+        service = new UnresolvedReferencesServiceImpl(productSyncOptions);
+    }
+
+    @Test
+    void save_WithSuccessfulMockCtpResponse_ShouldReturnMock() {
+        final CustomObject mock = mock(CustomObject.class);
```
lojzatran: Here the mock var is called `mock`, but later you have `final ProductDraft productDraft = mock(ProductDraft.class);`. Pls make it consistent.

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .errorCallback((errorMessage, errorException) -> {
+                    errorMessages.add(errorMessage);
+                    errorExceptions.add(errorException);
+                })
+                .build();
+        service = new UnresolvedReferencesServiceImpl(productSyncOptions);
+    }
+
+    @Test
+    void save_WithSuccessfulMockCtpResponse_ShouldReturnMock() {
```
lojzatran:
```suggestion
    void save_WithSuccessfulMockCtpResponse_ShouldSaveMock() {
```

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: > what about fetch with unsuccessful?

As we are calling this static method `QueryExecutionUtils.queryAll` and it does not possible to mock the handle the failed called directly.

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: do you mean the this particular test case?

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: yes, as we are using the static method of `QueryExecutionUtils.queryAll` which internally calls the ctpClient.execute() we need to return a `PagedQueryResult` to handle this result inside `queryAll` method.

src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
```diff
@@ -92,6 +93,44 @@ void save_WithoutException_createsNewCustomObject() {
...
+            .join();
+
+
+        // assertions
+        assertThat(latestResult).hasValueSatisfying(waitingToBeResolved ->
+            assertThat(waitingToBeResolved.getProductDraft()).isEqualTo(productDraft));
+
+        final CustomObjectByKeyGet<WaitingToBeResolved> customObjectByKeyGet = CustomObjectByKeyGet
+            .of(CUSTOM_OBJECT_CONTAINER_KEY, productDraft.getKey(), WaitingToBeResolved.class);
+        final CustomObject<WaitingToBeResolved> createdCustomObject = CTP_TARGET_CLIENT
```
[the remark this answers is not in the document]
JudeNiroshan: It is to assert the entire `WaitingToBeResolved` object has changed to `productDraftWithUnresolvedNewRefs` in the CTP project. Basically this assert whether the old value has been replaced with new value. This test case is made because we do not maintain versioned custom objects.

src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
```diff
@@ -92,6 +93,44 @@ void save_WithoutException_createsNewCustomObject() {
...
+        final WaitingToBeResolved productDraftWithUnresolvedRefs =
+                new WaitingToBeResolved(productDraft, asSet("foo", "bar"));
+
+        unresolvedReferencesService
+                .save(productDraftWithUnresolvedRefs)
+                .toCompletableFuture()
+                .join();
+
+        final WaitingToBeResolved productDraftWithUnresolvedNewRefs =
+                new WaitingToBeResolved(productDraft, asSet("foo123", "bar123"));
```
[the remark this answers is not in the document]
JudeNiroshan: Because this test case is to verify the replacing the value of an existing custom object value. This `save_WithoutException_createsNewCustomObject()` particular test case assert the custom object creation for the first time.

src/main/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImpl.java
JudeNiroshan: 3dea2d39e4f5b8ddd76371f5a685e88647732b0a

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: 22ce5b06a4ecb648f8606fd083faa2a6d8983714

src/main/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImpl.java
JudeNiroshan: 1e2a4526c537a0e4abe2a909bfa0b007a23ec7f5

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: cc8958914da78b9df1fa42bb9ae1cf36db93b815

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: 7c7f6be17f49411d13eaae28ea4d03483b85e152

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: baef4a8b2da882e0e29202ebf1305d035702d94a

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: 025f8449e3edd339aa0bd73514c9514109a99517

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: b0a11c56132d37f29326ab6ff749f0b5f1531b39

src/integration-test/java/com/commercetools/sync/integration/services/impl/UnresolvedReferencesServiceImplIT.java
```diff
@@ -36,7 +37,7 @@
         "commercetools-sync-java.UnresolvedReferencesService.productDrafts";
 
 
-    @BeforeEach
+    @AfterEach
```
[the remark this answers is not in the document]
JudeNiroshan: f7bb2a9607ea2fe1a25c412b1262f002e5c41b47

src/test/java/com/commercetools/sync/commons/models/WaitingToBeResolvedTest.java
```diff
@@ -0,0 +1,69 @@
...
+        // assertions
+        assertTrue(equalsResult);
+    }
+
+    @Test
+    void testEquals_withSimilarObjects() {
+        // preparation
+        when(productDraft.getKey()).thenReturn("test-product-key");
+        refSet.add("test-reference-key");
+        WaitingToBeResolved obj2 = new WaitingToBeResolved(productDraft, refSet);
```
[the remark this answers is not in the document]
JudeNiroshan: 3cf9cabbcd01cc748d49870ad84f31700a52ff15

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: 87a27419aff2220080dd3e5ce20fa9f182e6fa27

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: 95578f1b9bfe5a0363c6be232839f7fd23bda287

src/test/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImplTest.java
JudeNiroshan: 236b3d90326e21fc4c0e2c919cc4ffc03017fdb4

PR #464, opened 2019-10-15
463 fix query all

### LINKED ISSUE #463: QueryAll bug
**Opened from a support ticket**

I found a bug in the commercetools-sync-java:1.5.0 library.

I use the

CtpQueryUtils.queryAll(sphereClient, productQuery, pageCallback, pageSize).toCompletableFuture().join();

Static method to do a paged query.
My productQuery is:

ProductReferenceReplacementUtils.buildProductQuery()
.plusPredicates(m -> m.productType().is(ProductType.reference(productType)))

The result is, that I get in the FIRST page loop only products with the right product type and in all other loops all products. The reason is that the line

QueryAll.java:167 return client.execute(queryPredicate != null ? query.withPredicates(queryPredicate) : query);

Overwrites MY predicate. Therefore a solution would be to change the withPredicate to plusPredicate.

heshamMassoud:
#### Summary
Fixes #463 

- Tests
    - [x] Unit 
    ~- [ ] Integration~
~- [ ] Documentation~

### DIFF
--- a/docs/RELEASE_NOTES.md
+++ b/docs/RELEASE_NOTES.md
@@ -29,9 +29,9 @@
 -->
 <!--
 ### 1.6.1 -  Oct 16, 2019
-[Commits](https://github.com/commercetools/commercetools-sync-java/compare/1.6.0...1.7.0) |
-[Javadoc](https://commercetools.github.io/commercetools-sync-java/v/1.7.0/) | 
-[Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/1.7.0)
+[Commits](https://github.com/commercetools/commercetools-sync-java/compare/1.6.0...1.1.0) |
+[Javadoc](https://commercetools.github.io/commercetools-sync-java/v/1.6.1/) | 
+[Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/1.6.1)
 
 - 🐞 **Bug Fixes** (1)
     - **Commons** - Fixed a bug in the `CtpQueryUtils` which was overwriting the query input query for every page after
--- a/docs/RELEASE_NOTES.md
+++ b/docs/RELEASE_NOTES.md
@@ -28,10 +28,15 @@
 
 -->
 <!--
-### 1.7.0 -  Oct 17, 2019
+### 1.6.1 -  Oct 16, 2019
 [Commits](https://github.com/commercetools/commercetools-sync-java/compare/1.6.0...1.7.0) |
 [Javadoc](https://commercetools.github.io/commercetools-sync-java/v/1.7.0/) | 
 [Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/1.7.0)
+
+- 🐞 **Bug Fixes** (1)
+    - **Commons** - Fixed a bug in the `CtpQueryUtils` which was overwriting the query input query for every page after
+    first page fetched, eventually fetching more than needed.  [#463](https://github.com/commercetools/commercetools-sync-java/issues/463)
+
 -->
 
 ### 1.6.0 -  Oct 10, 2019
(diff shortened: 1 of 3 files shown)


src/main/java/com/commercetools/sync/commons/utils/QueryAll.java
```diff
@@ -164,7 +164,7 @@ private void mapOrConsume(@Nonnull final List<T> pageElements) {
     @Nonnull
     private CompletionStage<PagedQueryResult<T>> queryPage(@Nullable final QueryPredicate<T> queryPredicate) {
         final QueryDsl<T, C> query = baseQuery.withLimit(pageSize);
```
butenkor: As discussed maybe we should always take queryPredicate and not assume that it might be null. It also means that here https://github.com/commercetools/commercetools-sync-java/pull/464/files#diff-eebb1f732a21746c9d39441c25721442L177 you also do not pass a null but users query.

heshamMassoud: thanks for the idea: 52918d7

docs/RELEASE_NOTES.md
```diff
@@ -28,10 +28,15 @@
 
 -->
 <!--
-### 1.7.0 -  Oct 17, 2019
+### 1.6.1 -  Oct 16, 2019
 [Commits](https://github.com/commercetools/commercetools-sync-java/compare/1.6.0...1.7.0) |
```
lojzatran:
```suggestion
[Commits](https://github.com/commercetools/commercetools-sync-java/compare/1.6.0...1.6.1) |
```

src/main/java/com/commercetools/sync/commons/utils/QueryAll.java
```diff
@@ -107,12 +114,14 @@ private QueryAll(@Nonnull final SphereClient client,
      *         the method returns a completed future containing null.
      */
     @Nonnull
+    @SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION") // `https://github.com/findbugsproject/findbugs/issues/79
     private CompletionStage<PagedQueryResult<T>> processPageAndGetNext(@Nonnull final PagedQueryResult<T> page) {
         final List<T> currentPageElements = page.getResults();
-        if (currentPageElements.size() > 0) {
+        if (!currentPageElements.isEmpty()) {
```
lojzatran: Why this change?

src/main/java/com/commercetools/sync/commons/utils/QueryAll.java
```diff
@@ -148,32 +157,8 @@ private void mapOrConsume(@Nonnull final List<T> pageElements) {
         if (pageElements.size() == pageSize) {
             final String lastElementId = pageElements.get(pageElements.size() - 1).getId();
             final QueryPredicate<T> queryPredicate = QueryPredicate.of(format("id > \"%s\"", lastElementId));
-            return queryPage(queryPredicate);
+            return client.execute(baseQuery.plusPredicates(queryPredicate));
```
lojzatran: If you have multiple pages, will it keep appending the query?

src/main/java/com/commercetools/sync/commons/utils/QueryAll.java
heshamMassoud: it was sonarlint suggestion:
> Using Collection.size() to test for emptiness works, but using Collection.isEmpty() makes the code more readable and can be more performant. The time complexity of any isEmpty() method implementation should be O(1) whereas some implementations of size() can be O(n).

https://rules.sonarsource.com/java/tag/clumsy/RSPEC-1155

src/main/java/com/commercetools/sync/commons/utils/QueryAll.java
heshamMassoud: nope, because `plusPredicates` does not mutate the  `baseQuery` but copies and returns new instance.

docs/RELEASE_NOTES.md
heshamMassoud: bf7d8e3

src/main/java/com/commercetools/sync/commons/utils/QueryAll.java
heshamMassoud: approved f2f.

Issue #494, opened 2020-03-18
CartDiscountSync generates a changeValue action when there is no change


michaelbannister:
**Describe the bug**
CartDiscountSync generates a changeValue action when there is no change, causing the sync to fail for that cart discount.

**To Reproduce**
Using commercetools-java-client 1.50.0 and commercetools-sync-java 1.8.0

I have a CartDiscount: [cartdiscount.json.txt](https://github.com/commercetools/commercetools-sync-java/files/4350455/cartdiscount.json.txt)

I run this Kotlin code:
```kotlin
val cartDiscountSyncOptions = CartDiscountSyncOptionsBuilder.of(client)
        .errorCallback { msg, e -> logger.error(msg, e) }
        .warningCallback { msg -> logger.warn(msg) }
        .build()
val cartDiscountSync = CartDiscountSync(cartDiscountSyncOptions)

runBlocking {
    val bundleId = "238594087"
    val bundleDiscount = CartDiscountValue.ofAbsolute(
            Money.of(50, Monetary.getCurrency("GBP")))
    val target = LineItemsTarget.of("""sku in ("238005351","238298178")""")

    val cartDiscountDraft = CartDiscountDraftBuilder.of(
                    "bundle-${bundleId}".localizedUK(),
                    CartPredicate.of("1 = 1"),
                    bundleDiscount,
                    target,
                    "0.9238594087288394",
                    true
            ).key("bundle-${bundleId}")
            .description("sonos-bundle".localizedUK())
            .active(true)
            .build()
    val cartDiscountStats = cartDiscountSync.sync(listOf(cartDiscountDraft)).await()
    logger.info(cartDiscountStats.reportMessage)
}
```

The sync fails with this output:
```
Failed to update cart discount with key: 'bundle-238594087'. Reason: io.sphere.sdk.client.ErrorResponseException: detailMessage: 'value' has no changes.
summary: POST https://commercetools-instance-redacted/project-redacted/cart-discounts/b2dae369-c3e7-4a04-9a40-975aef0a5175 failed  with response code 400 with X-Correlation-ID `project-redacted/04072b1b-e035-479b-be3f-5db07a14d632` on 2020-03-18T17:37:29.000790Z
http response formatted body: {
  "statusCode" : 400,
  "message" : "'value' has no changes.",
  "errors" : [ {
    "code" : "InvalidOperation",
    "message" : "'value' has no changes.",
    "action" : {
      "action" : "changeValue",
      "value" : {
[...]

PR #495, opened 2020-03-20
494 improve CartDiscountSync handling of javax.money.MonetaryAmount

michaelbannister:
#### Summary
Compare AbsoluteCartDiscountValue list elements using `isEqualTo` instead of (implicitly) `equals`.

#### Description
CartDiscountSync generated a setValue action if the CartDiscountDraft values were built using a different `javax.money.MonetaryAmount` implementation. This meant that a sync would generate setValue actions that made commercetools throw an InvalidOperation error saying "'value' has no changes."

The JSR 354 spec for [MonetaryAmount](https://javamoney.github.io/apidocs/javax/money/MonetaryAmount.html) explicitly states that the `equals` method should not consider instances from two different implementation to be equal, and the `isEqualTo` method is the appropriate comparison in this case.
`io.sphere.sdk.utils.MoneyImpl` is actually not compliant with the spec in this regard, since its `equals` method is more permissive, but `org.javamoney.moneta.Money` is compliant and hence we get some surprising behaviour. It's probably not desirable to change the implementation of `io.sphere.sdk.utils.MoneyImpl` as people may be relying on it with its current behaviour, so better to fix here in CartDiscountUpdateActionUtils.

#### Relevant Issues
#494 
JavaMoney/jsr354-ri#328

#### Todo

- Tests
    - [x] Unit 
    - [ ] Integration
- [ ] Documentation

### DIFF
--- a/src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
+++ b/src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
@@ -308,7 +308,7 @@ class CartDiscountUpdateActionUtilsTest {
     }
 
     @Test
-    void buildChangeValueUpdateAction_WithSameAbsoluteValuesUsingDifferentMonetaryAmountImplementation_ShouldNotBuildUpdateAction() {
+    void buildChangeValueUpdateAction_WithSameValuesUsingDifferentMonetaryAmount_ShouldNotBuildUpdateAction() {
         final CartDiscountValue values =
                 CartDiscountValue.ofAbsolute(asList(MoneyImpl.of(10, EUR), MoneyImpl.of(10, USD)));
         final CartDiscountValue values2 =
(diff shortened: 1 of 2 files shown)


src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java
```diff
@@ -306,6 +307,25 @@ void buildChangeValueUpdateAction_WithSameAbsoluteValues_ShouldNotBuildUpdateAct
         assertThat(changeValueUpdateAction).isNotPresent();
     }
 
+    @Test
+    void buildChangeValueUpdateAction_WithSameAbsoluteValuesUsingDifferentMonetaryAmountImplementation_ShouldNotBuildUpdateAction() {
```
ahmetoz: Could you please check the `checkstyleTest` task of the build that causes the build to failed ?

````
> Task :checkstyleTest
[ant:checkstyle] [WARN] /home/travis/build/commercetools/commercetools-sync-java/src/test/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountUpdateActionUtilsTest.java:311: Line is longer than 120 characters (found 133). [LineLength]
````

ahmetoz:
@michaelbannister thanks for fixing the tests, as it's a fork build there are no credentials on Travis so ITs did not work, but it seems all unit tests are working, I will check the ITs on my local and let you know if sth needs to be fixed or I will merge it. 

Again thanks for the contribution 🙏

Issue #466, opened 2019-10-16
ClassCastException while getting response


jarzynp:
**Describe the bug**
While doing ProductType synchronization following error appeared in our logs: 
```
java.util.concurrent.CompletionException: java.lang.ClassCastException: java.util.concurrent.CompletionException cannot be cast to com.commercetools.sync.commons.exceptions.ReferenceResolutionException
```

Bug exists in this line: https://github.com/commercetools/commercetools-sync-java/blob/d0b2a9877b5e45b025ba370de05de100a89eaa0a/src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java#L234

**To Reproduce**
Unfortunately we don't have any details what was the cause of it and don't know how to reproduce.

**Expected behavior**
More detailed exception about what actually happened (maybe exception could contain response?).

**Additional context**
Stacktrace:
```
java.util.concurrent.CompletionException: java.lang.ClassCastException: java.util.concurrent.CompletionException cannot be cast to com.commercetools.sync.commons.exceptions.ReferenceResolutionException
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:874)
	at java.util.concurrent.CompletableFuture.uniExceptionallyStage(CompletableFuture.java:884)
	at java.util.concurrent.CompletableFuture.exceptionally(CompletableFuture.java:2196)
	at java.util.concurrent.CompletableFuture.exceptionally(CompletableFuture.java:110)
	at com.commercetools.sync.producttypes.ProductTypeSync.lambda$syncBatch$7(ProductTypeSync.java:233)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1556)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:546)
[...]

PR #505, opened 2020-04-14
#466: Fix ClassCastException caused by ReferenceResolutionException

leungkinghin:
Issue #466 : ClassCastException occurs when handling ReferenceResolutionException

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -23,7 +23,6 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-import java.util.concurrent.CompletionException;
 
 import static com.commercetools.sync.commons.asserts.statistics.AssertionsForStatistics.assertThat;
 import static com.commercetools.sync.commons.helpers.BaseReferenceResolver.BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER;
@@ -168,15 +167,18 @@ class ProductReferenceResolverIT {
 
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
+        assertThat(errorCallBackExceptions).hasSize(1);
+        if (errorCallBackExceptions.get(0) instanceof ReferenceResolutionException) {
             assertThat(errorCallBackMessages).containsExactly(
                     format("Failed to resolve references on ProductDraft with"
-                    + " key:'%s'. Reason: "+ ReferenceResolutionException.class.getCanonicalName() + ": "
-                    + "Failed to resolve 'product-type' resource identifier on ProductDraft with "
+                    + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
                     + "key:'%s'. Reason: %s",
                 productDraft.getKey(), productDraft.getKey(),
                 BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER));
-        assertThat(errorCallBackExceptions).hasSize(1);
-        assertThat(errorCallBackExceptions.get(0)).isExactlyInstanceOf(CompletionException.class);
+        } else {
+            assertThat(errorCallBackMessages).contains(
+                format("Failed to sync draft on ProductDraft with key:'%s'.", productDraft.getKey()));
+        }
         assertThat(warningCallBackMessages).isEmpty();
     }
 }
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -23,6 +23,7 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.CompletionException;
 
 import static com.commercetools.sync.commons.asserts.statistics.AssertionsForStatistics.assertThat;
 import static com.commercetools.sync.commons.helpers.BaseReferenceResolver.BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER;
@@ -168,12 +169,14 @@ class ProductReferenceResolverIT {
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
         assertThat(errorCallBackMessages).containsExactly(format("Failed to process the ProductDraft with"
-                + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
+                + " key:'%s'. Reason: "
+                + ReferenceResolutionException.class.getCanonicalName() + ": "
+                + "Failed to resolve 'product-type' resource identifier on ProductDraft with "
                 + "key:'%s'. Reason: %s",
             productDraft.getKey(), productDraft.getKey(),
             BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER));
         assertThat(errorCallBackExceptions).hasSize(1);
-        assertThat(errorCallBackExceptions.get(0)).isExactlyInstanceOf(ReferenceResolutionException.class);
+        assertThat(errorCallBackExceptions.get(0)).isExactlyInstanceOf(CompletionException.class);
         assertThat(warningCallBackMessages).isEmpty();
     }
 }
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -23,7 +23,6 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-import java.util.concurrent.CompletionException;
 
 import static com.commercetools.sync.commons.asserts.statistics.AssertionsForStatistics.assertThat;
 import static com.commercetools.sync.commons.helpers.BaseReferenceResolver.BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER;
@@ -168,15 +167,18 @@ class ProductReferenceResolverIT {
 
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
-        assertThat(errorCallBackMessages).containsExactly(format("Failed to resolve references on ProductDraft with"
-                    + " key:'%s'. Reason: "
-                    +  ReferenceResolutionException.class.getCanonicalName() + ": "
-                    + "Failed to resolve 'product-type' resource identifier on ProductDraft with "
+        assertThat(errorCallBackExceptions).hasSize(1);
+        if (errorCallBackExceptions.get(0) instanceof ReferenceResolutionException) {
+            assertThat(errorCallBackMessages).containsExactly(
+                    format("Failed to resolve references on ProductDraft with"
+                    + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
                     + "key:'%s'. Reason: %s",
                 productDraft.getKey(), productDraft.getKey(),
                 BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER));
-        assertThat(errorCallBackExceptions).hasSize(1);
-        assertThat(errorCallBackExceptions.get(0)).isExactlyInstanceOf(CompletionException.class);
+        } else {
+            assertThat(errorCallBackMessages).contains(
+                format("Failed to sync draft on ProductDraft with key:'%s'.", productDraft.getKey()));
+        }
         assertThat(warningCallBackMessages).isEmpty();
     }
 }
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -23,6 +23,7 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.CompletionException;
 
 import static com.commercetools.sync.commons.asserts.statistics.AssertionsForStatistics.assertThat;
 import static com.commercetools.sync.commons.helpers.BaseReferenceResolver.BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER;
@@ -167,18 +168,15 @@ class ProductReferenceResolverIT {
 
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
-        assertThat(errorCallBackExceptions).hasSize(1);
-        if (errorCallBackExceptions.get(0) instanceof ReferenceResolutionException) {
-            assertThat(errorCallBackMessages).containsExactly(
-                    format("Failed to resolve references on ProductDraft with"
-                    + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
+        assertThat(errorCallBackMessages).containsExactly(format("Failed to resolve references on ProductDraft with"
+                    + " key:'%s'. Reason: "
+                    +  ReferenceResolutionException.class.getCanonicalName() + ": "
+                    + "Failed to resolve 'product-type' resource identifier on ProductDraft with "
                     + "key:'%s'. Reason: %s",
                 productDraft.getKey(), productDraft.getKey(),
                 BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER));
-        } else {
-            assertThat(errorCallBackMessages).contains(
-                format("Failed to sync draft on ProductDraft with key:'%s'.", productDraft.getKey()));
-        }
+        assertThat(errorCallBackExceptions).hasSize(1);
+        assertThat(errorCallBackExceptions.get(0)).isExactlyInstanceOf(CompletionException.class);
         assertThat(warningCallBackMessages).isEmpty();
     }
 }
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -23,6 +23,7 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.CompletionException;
 
 import static com.commercetools.sync.commons.asserts.statistics.AssertionsForStatistics.assertThat;
 import static com.commercetools.sync.commons.helpers.BaseReferenceResolver.BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER;
@@ -168,17 +169,14 @@ class ProductReferenceResolverIT {
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
         assertThat(errorCallBackExceptions).hasSize(1);
-        if (errorCallBackExceptions.get(0) instanceof ReferenceResolutionException) {
-            assertThat(errorCallBackMessages).containsExactly(
+        assertThat(errorCallBackExceptions.get(0)).isExactlyInstanceOf(CompletionException.class);
+        assertThat(errorCallBackMessages).containsExactly(
                     format("Failed to resolve references on ProductDraft with"
-                    + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
+                    + " key:'%s'. Reason: " + ReferenceResolutionException.class.getCanonicalName() + ": "
+                    + "Failed to resolve 'product-type' resource identifier on ProductDraft with "
                     + "key:'%s'. Reason: %s",
                 productDraft.getKey(), productDraft.getKey(),
                 BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER));
-        } else {
-            assertThat(errorCallBackMessages).contains(
-                format("Failed to sync draft on ProductDraft with key:'%s'.", productDraft.getKey()));
-        }
         assertThat(warningCallBackMessages).isEmpty();
     }
 }
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -167,7 +167,7 @@ class ProductReferenceResolverIT {
 
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
-        assertThat(errorCallBackMessages).containsExactly(format("Failed to run sync draft process on ProductDraft with"
+        assertThat(errorCallBackMessages).containsExactly(format("Failed to process the ProductDraft with"
                 + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
                 + "key:'%s'. Reason: %s",
             productDraft.getKey(), productDraft.getKey(),
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -167,7 +167,7 @@ class ProductReferenceResolverIT {
 
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
-        assertThat(errorCallBackMessages).containsExactly(format("Failed to resolve references on ProductDraft with"
+        assertThat(errorCallBackMessages).containsExactly(format("Failed to run sync draft process on ProductDraft with"
                     + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
                         + "key:'%s'. Reason: %s",
                 productDraft.getKey(), productDraft.getKey(),
--- a/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
@@ -23,7 +23,6 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-import java.util.concurrent.CompletionException;
 
 import static com.commercetools.sync.commons.asserts.statistics.AssertionsForStatistics.assertThat;
 import static com.commercetools.sync.commons.helpers.BaseReferenceResolver.BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER;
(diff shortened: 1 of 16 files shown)


src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
```diff
@@ -167,13 +167,18 @@ void sync_withNewProductWithNoProductTypeKey_ShouldFailCreatingTheProduct() {
...
         // assertion
         assertThat(syncStatistics).hasValues(1, 0, 0, 1);
-        assertThat(errorCallBackMessages).containsExactly(format("Failed to resolve references on ProductDraft with"
-                + " key:'%s'. Reason: Failed to resolve 'product-type' resource identifier on ProductDraft with "
-                + "key:'%s'. Reason: %s",
-            productDraft.getKey(), productDraft.getKey(),
-            BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER));
         assertThat(errorCallBackExceptions).hasSize(1);
-        assertThat(errorCallBackExceptions.get(0)).isExactlyInstanceOf(ReferenceResolutionException.class);
+        if (errorCallBackExceptions.get(0) instanceof ReferenceResolutionException) {
```
ahmetoz, line 171: a test assertion should not expect it with if, you need to expect it directly, please add additional tests for different cases.

src/integration-test/java/com/commercetools/sync/integration/externalsource/cartdiscounts/CartDiscountSyncIT.java
```diff
@@ -308,20 +308,23 @@ void sync_WithUpdatedCartDiscount_WithNewCustomTypeWithWrongResIdentifier_Should
...
-        assertThat(errorMessages).containsExactly(
-            "Failed to resolve references on CartDiscountDraft with key:'key_1'. Reason: "
-                + "Failed to resolve custom type reference on CartDiscountDraft with key:'key_1'. Reason: "
-                + "The value of the 'id' field of the Resource Identifier/Reference is blank (null/empty).");
-        assertThat(exceptions)
-            .hasSize(1)
-            .hasOnlyOneElementSatisfying(exception -> {
-                assertThat(exception).isInstanceOf(ReferenceResolutionException.class);
+        assertThat(exceptions).hasSize(1);
+        assertThat(errorMessages).hasSize(1);
```
ahmetoz, line 312: a test assertion should not expect it with if, you need to expect it directly, please add additional tests for different cases.

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductReferenceResolverIT.java
```diff
@@ -107,15 +107,22 @@ void sync_withNewProductWithInvalidCategoryReferences_ShouldFailCreatingTheProdu
...
         assertThat(syncStatistics).hasValues(1, 0, 0, 1, 0);
         assertThat(errorCallBackExceptions).hasSize(1);
+        assertThat(errorCallBackMessages).hasSize(1);
         final Throwable exception = errorCallBackExceptions.get(0);
-        assertThat(exception).isExactlyInstanceOf(ReferenceResolutionException.class)
-                             .hasMessageContaining("Failed to resolve 'category' resource identifier on ProductDraft "
-                                 + "with key:'productKey1'")
-                             .hasMessageContaining(format("Reason: %s", BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER));
+        final String errorMessage = errorCallBackMessages.get(0);
+        if (exception instanceof ReferenceResolutionException) {
```
ahmetoz, line 113: a test assertion should not expect it with if, you need to expect it directly, please add additional tests for different cases.

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
```diff
@@ -202,12 +204,21 @@ private void handleError(@Nonnull final String errorMessage,
                         .resolveReferences(newCartDiscount)
                         .thenCompose(resolvedDraft -> syncDraft(oldCartDiscountMap, resolvedDraft))
                         .exceptionally(completionException -> {
-                            final ReferenceResolutionException referenceResolutionException =
-                                (ReferenceResolutionException) completionException.getCause();
-                            final String errorMessage = format(FAILED_TO_RESOLVE_REFERENCES, newCartDiscount.getKey(),
-                                             referenceResolutionException.getMessage());
-                            handleError(errorMessage, referenceResolutionException, 1);
-                            return null;
+                            if (completionException.getCause() instanceof ReferenceResolutionException) {
```
ahmetoz, line 207: why do you need getCause in here? checking the expectation itself was not enough?

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
```diff
@@ -202,12 +204,21 @@ private void handleError(@Nonnull final String errorMessage,
...
+                                final ReferenceResolutionException referenceResolutionException =
```
ahmetoz, line 208: please refactor the lambda expression in here, this is totally not easy to understand, my suggestion is moving those checks to another method with a good name.

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
```diff
@@ -107,7 +107,7 @@ protected CustomReferenceResolver(@Nonnull final S options, @Nonnull final TypeS
         try {
             final String customTypeKey = getKeyFromResourceIdentifier(custom.getType());
             return typeService.fetchCachedTypeId(customTypeKey);
-        } catch (ReferenceResolutionException exception) {
+        } catch (Exception exception) {
```
ahmetoz, line 110: this is not correct! it seems the code will swallow the other exceptions types as ReferenceResolutionException.

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -332,16 +334,21 @@ public ProductSync(@Nonnull final ProductSyncOptions productSyncOptions) {
...
             .exceptionally(completionException -> {
-
-                final ReferenceResolutionException referenceResolutionException =
-                    (ReferenceResolutionException) completionException.getCause();
-                final String errorMessage = format(FAILED_TO_RESOLVE_REFERENCES, newProductDraft.getKey(),
-                    referenceResolutionException.getMessage());
-                handleError(errorMessage, referenceResolutionException, 1);
-
-                return null;
+                if (completionException.getCause() instanceof ReferenceResolutionException) {
```
ahmetoz, line 338: the part related to the exception handling seems duplicated with the CartDiscountSync logic except for different keys and different messages, maybe avoid the duplication so when we needed to change that logic again we would not need to refactor all Sync instances.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
```diff
@@ -147,10 +147,10 @@ public ProductReferenceResolver(@Nonnull final ProductSyncOptions productSyncOpt
             return resolveResourceIdentifier(draftBuilder, draftBuilder.getProductType(),
                 productTypeService::fetchCachedProductTypeId, ResourceIdentifier::ofId,
                 ProductDraftBuilder::productType);
-        } catch (ReferenceResolutionException referenceResolutionException) {
+        } catch (Exception exception) {
```
ahmetoz, line 150: That does not seem right to me. The catch block is only handling the reference resolution exception, maybe we would need another handler for generic Exception, instead of swallowing it.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
```diff
@@ -205,20 +205,25 @@ public ProductReferenceResolver(@Nonnull final ProductSyncOptions productSyncOpt
...
+                                    ofNullable(categoryOrderHints.get(category.getKey()))
+                                            .ifPresent(orderHintValue -> categoryOrderHintsMap
+                                                    .put(category.getId(), orderHintValue));
+                                }
+                                return category.toReference();
+                            }).collect(toList()))
+                    .thenApply(categoryReferences -> draftBuilder
+                            .categories(categoryReferences)
+                            .categoryOrderHints(CategoryOrderHints.of(categoryOrderHintsMap)));
+        } catch (Exception exception) {
```
ahmetoz, line 222: How do you know the exception is related to reference? maybe NPE happened in that block of code ?

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
leungkinghin: java.util.concurrent.CompletionException is thrown, and the actual cause is wrapped inside.

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
leungkinghin: This is the purpose. ReferenceResolutionException is considered as exception wrapper to indicate any exception thrown during resolveReference(), including any unchecked exception occurs in runtime.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
leungkinghin: Suppose to handle runtime exception, since categoryService.fetchMatchingCategoriesByKeys(...) goes to make query over network

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
leungkinghin: This is the purpose. ReferenceResolutionException is considered as exception wrapper to indicate any exception thrown during resolveReference(), thus any unchecked exception occurs in runtime should be included.

ahmetoz: if it's not ReferenceResolutionException but let's say NPE (null pointer exception)? IMO, you added that part, but handling the exception is not changed, thus changing only catch block is not enough.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
ahmetoz: then why in catch block we're putting it into a Reference exception with a message ?

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
ahmetoz: what if it's not caused with that way, we would not know it before testing it with different exception classes with different cause types, I also think that it would work without a problem, but a code should always be running against an automation test rather than expecting that it would work without a problem.

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
butenkor: I wonder, why do we need to catch here at all? I see ReferenceResolutionException only as an additional wrapper which adds additional context/details to the error. Important is that such error should not break the whole batch processing but rather be passed to the error callback so that the user can react to it.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
butenkor: Why do we need to catch here? See also my previous related comment.

src/main/java/com/commercetools/sync/producttypes/helpers/AttributeDefinitionReferenceResolver.java
```diff
@@ -112,12 +112,12 @@ public AttributeDefinitionReferenceResolver(@Nonnull final ProductTypeSyncOption
         final String resourceKey;
         try {
             resourceKey = getKeyFromResourceIdentifier(typeReference);
-        } catch (ReferenceResolutionException exception) {
+            return productTypeService.fetchCachedProductTypeId(resourceKey)
+                    .thenApply(optionalId -> optionalId.map(ProductType::referenceOfId));
+        } catch (Exception exception) {
```
butenkor, line 117: Same as above.

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
leungkinghin: It is behaviour of CompletableFuture class. The wrapping occurs in the following line
https://github.com/commercetools/commercetools-sync-java/blob/7baae24007cbb49afe7a6345bfee535765f3ac7b/src/main/java/com/commercetools/sync/producttypes/helpers/AttributeDefinitionReferenceResolver.java#L116

imo if we don't return the exception in such way, we have to change it as try-catch pattern simply in order to get rid of completionException.getCause(). Any other suggestion?

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
leungkinghin: I think, In line 109, the service class attempts to make query via jvm-sdk. If HttpException is suspected to occur, this coding line potentially involves. That's why I catch it here, or otherwise runtime exception is directly thrown to sync class without any handling taking place.

leungkinghin:
**Description** 
1. Remove the exception casting in ProductSync, ProductTypeSync and CartDiscountSync. Directly pass the CompletionException returned into handleError method.
2. Updated corresponding unit test and integration test because CompletionException is thrown instead of ReferenceResolutionException.

**Commit**
https://github.com/commercetools/commercetools-sync-java/pull/505/commits/b6d96a84e6d2ebcf370f8145904cdecc894b8cd2

src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/products/ProductReferenceResolverIT.java
leungkinghin: The if-else block has been removed in latest change.

src/integration-test/java/com/commercetools/sync/integration/externalsource/cartdiscounts/CartDiscountSyncIT.java
leungkinghin: The if-else block has been removed in latest change.

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductReferenceResolverIT.java
leungkinghin: The if-else block has been removed in latest change.

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
leungkinghin: As discussed with Roman, the thrown exception will be passed into handleError directly and let user to deal with it. If-else checking is no longer used in Sync classes.

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
leungkinghin: As discussed with Roman, the thrown exception will be passed into handleError directly and let user to deal with it. If-else checking is no longer used in Sync classes.

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
leungkinghin: The change has been rolled back

src/main/java/com/commercetools/sync/products/ProductSync.java
leungkinghin: As discussed with Roman, the thrown exception will be passed into handleError directly and let user to deal with it. If-else checking is no longer used in Sync classes.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
leungkinghin: This change has been rolled back.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
leungkinghin: This change has been rolled back.

src/main/java/com/commercetools/sync/producttypes/helpers/AttributeDefinitionReferenceResolver.java
leungkinghin: These changes has been obsoleted. Please refer to followings
https://github.com/commercetools/commercetools-sync-java/commit/b6d96a84e6d2ebcf370f8145904cdecc894b8cd2

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -223,7 +225,16 @@ private void handleError(@Nonnull final String errorMessage, @Nullable final Thr
                 removeAndKeepTrackOfMissingNestedAttributes(newProductType, keyToIdCache))
             .map(draftWithoutMissingRefAttrs -> referenceResolver
                 .resolveReferences(draftWithoutMissingRefAttrs)
-                .thenCompose(resolvedDraft -> syncDraft(oldProductTypeMap, resolvedDraft)))
+                .thenCompose(resolvedDraft -> syncDraft(oldProductTypeMap, resolvedDraft))
+                .exceptionally(completionException -> {
+                    final String errorMessage = format(FAILED_TO_RESOLVE_REFERENCES,
```
butenkor, line 230: Why do you assume that there will be a reference resolution exception only? What about `syncDraft` method which might throw a different type of error (400 or 409 for example)?

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -324,26 +325,17 @@ public ProductSync(@Nonnull final ProductSyncOptions productSyncOptions) {
...
                     .map(product -> fetchProductAttributesMetadataAndUpdate(oldProduct, resolvedDraft))
                     .orElseGet(() -> applyCallbackAndCreate(resolvedDraft));
-
             })
-
             .exceptionally(completionException -> {
-
-                final ReferenceResolutionException referenceResolutionException =
-                    (ReferenceResolutionException) completionException.getCause();
                 final String errorMessage = format(FAILED_TO_RESOLVE_REFERENCES, newProductDraft.getKey(),
```
butenkor, line 334: `fetchProductAttributesMetadataAndUpdate` and `applyCallbackAndCreate` might throw an exception which is not related to ReferenceResolution? It means that this message would be wrong/confusing.

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
```diff
@@ -202,12 +200,12 @@ private void handleError(@Nonnull final String errorMessage,
                         .resolveReferences(newCartDiscount)
                         .thenCompose(resolvedDraft -> syncDraft(oldCartDiscountMap, resolvedDraft))
                         .exceptionally(completionException -> {
-                            final ReferenceResolutionException referenceResolutionException =
-                                (ReferenceResolutionException) completionException.getCause();
-                            final String errorMessage = format(FAILED_TO_RESOLVE_REFERENCES, newCartDiscount.getKey(),
-                                             referenceResolutionException.getMessage());
-                            handleError(errorMessage, referenceResolutionException, 1);
+                            final String errorMessage = format(FAILED_TO_RESOLVE_REFERENCES,
```
butenkor, line 203: `syncDraft ` might throw an exception which is not related to ReferenceResolution? It means that this message would be wrong/confusing.

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
leungkinghin: Since our approach now is to pass the raw exception to handleError(..) no matter which type it is, the program would not distinguish exception type inside exceptionally(..). How about re-defining the error message content not to be resolve references specific?

src/main/java/com/commercetools/sync/products/ProductSync.java
leungkinghin: Since our approach now is to pass the raw exception to handleError(..) no matter which type it is, the program would not distinguish exception type inside exceptionally(..). How about re-defining the error message content not to be resolve references specific?

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -62,6 +61,8 @@
     private static final String UPDATE_FAILED = "Failed to update Product with key: '%s'. Reason: %s";
     private static final String FAILED_TO_RESOLVE_REFERENCES = "Failed to resolve references on "
         + "ProductDraft with key:'%s'. Reason: %s";
+    private static final String FAILED_TO_SYNC_DRAFT  = "Failed to sync draft on "
```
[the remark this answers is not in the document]
leungkinghin: This variable has been obsoleted in new commit.

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
leungkinghin: Since our approach now is to pass the raw exception to handleError(..) no matter which type it is, the program would not distinguish exception type inside exceptionally(..). How about re-defining the error message content not to be resolve references specific?

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
butenkor: > Since our approach now is to pass the raw exception to handleError(..) no matter which type it is, the program would not distinguish exception type inside exceptionally

The new approach is better but the error message is misleading if the root cause is not related to reference resolution.

> How about re-defining the error message content not to be ReferenceResolution-specific?

Please make a suggestion

leungkinghin: > Please make a suggestion

Because the entire process is inovked inside ProductTypeSync.syncBatch(..), we can simply change as below 

`private static final String FAILED_TO_RUN_SYNC_BATCH = "Failed to run sync batch process on productTypeDraft with key:'%s'. Reason: %s";`

The error message contains reason part, which is enough to show the actual cause. It would be like following

_Failed to run sync batch process productTypeDraft with key:'001'. Reason: Failed to resolve references on attribute definition with name 'xxx'_

It can also apply in case runtime exception is thrown

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
```diff
@@ -107,10 +107,11 @@ protected CustomReferenceResolver(@Nonnull final S options, @Nonnull final TypeS
         try {
             final String customTypeKey = getKeyFromResourceIdentifier(custom.getType());
             return typeService.fetchCachedTypeId(customTypeKey);
-        } catch (ReferenceResolutionException exception) {
+        } catch (ReferenceResolutionException referenceResolutionException) {
```
ahmetoz, line 110: What is the point of your change in here, it does not seem necessary to me. You are not improving anything in here, the handler was also simple and easy to read before.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
```diff
@@ -205,20 +205,19 @@ public ProductReferenceResolver(@Nonnull final ProductSyncOptions productSyncOpt
...
-                                          ofNullable(categoryOrderHints.get(category.getKey()))
-                                              .ifPresent(orderHintValue -> categoryOrderHintsMap
-                                                  .put(category.getId(), orderHintValue));
-                                      }
-                                      return category.toReference();
-                                  }).collect(toList()))
-                              .thenApply(categoryReferences -> draftBuilder
-                                  .categories(categoryReferences)
-                                  .categoryOrderHints(CategoryOrderHints.of(categoryOrderHintsMap)));
+                    .thenApply(categories ->
```
ahmetoz, line 209: I could not see any change in here, please do not override the history.

src/main/java/com/commercetools/sync/producttypes/helpers/AttributeDefinitionReferenceResolver.java
```diff
@@ -112,12 +112,14 @@ public AttributeDefinitionReferenceResolver(@Nonnull final ProductTypeSyncOption
         final String resourceKey;
         try {
             resourceKey = getKeyFromResourceIdentifier(typeReference);
-        } catch (ReferenceResolutionException exception) {
+
+        } catch (ReferenceResolutionException referenceResolutionException) {
```
ahmetoz, line 116: What is the point of your change in here, it does not seem necessary to me. You are not improving anything in here, the handler was also simple and easy to read before.

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
```diff
@@ -223,7 +225,15 @@ private void handleError(@Nonnull final String errorMessage, @Nullable final Thr
                 removeAndKeepTrackOfMissingNestedAttributes(newProductType, keyToIdCache))
             .map(draftWithoutMissingRefAttrs -> referenceResolver
                 .resolveReferences(draftWithoutMissingRefAttrs)
-                .thenCompose(resolvedDraft -> syncDraft(oldProductTypeMap, resolvedDraft)))
+                .thenCompose(resolvedDraft -> syncDraft(oldProductTypeMap, resolvedDraft))
+                .exceptionally(completionException -> {
```
ahmetoz, line 229: I am not sure from that type `CompletionException` from a library design perspective, it's itself related to the java futures because the task is running concurrently and exceptions happening in the future stages, so the exception is wrapped by CompletionException.class always, what about returning getCause, I would suggest some investigation on that: 

From the https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html:
``` 
 In all other cases, if a stage's computation terminates abruptly with an 
(unchecked) exception or error, then all dependent stages requiring its completion complete exceptionally as well, with a 
CompletionException holding the exception as its cause. If a stage is dependent on both of two stages, and both 
complete exceptionally, then the CompletionException may correspond to either one of these exceptions. If a stage is 
dependent on either of two others, and only one of them completes exceptionally, no guarantees are made about whether 
the dependent stage completes normally or exceptionally. 
```

leungkinghin: That's what I have explained two weeks ago, turn out our conclusion is to get rid of returning .getCause(). Anyway when we put the raw exception into handleError(..) method, it doesn't matter what the actual exception class it wraps.

src/main/java/com/commercetools/sync/producttypes/helpers/AttributeDefinitionReferenceResolver.java
leungkinghin: Would base on what implemented in master branch, so I would rollback this change.

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
leungkinghin: Would base on what implemented in master branch, so I would rollback this change.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
leungkinghin: Removed the try-catch block here. Sorry for the indentation change in line 209.

src/main/java/com/commercetools/sync/producttypes/ProductTypeSync.java
leungkinghin: As suggested, it now returns CompletionException.getCause().

src/main/java/com/commercetools/sync/commons/helpers/CustomReferenceResolver.java
leungkinghin: As suggested, change here has been rolled back.

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
```diff
@@ -46,9 +45,8 @@
         "Failed to process cart discount draft without key.";
     private static final String CART_DISCOUNT_DRAFT_IS_NULL =
         "Failed to process null cart discount draft.";
-    private static final String FAILED_TO_RESOLVE_REFERENCES = "Failed to resolve references on "
-        + "CartDiscountDraft with key:'%s'. Reason: %s";
-
+    private static final String FAILED_TO_RUN_SYNC_DRAFT = "Failed to run sync draft process on "
```
butenkor, line 48: May be better - as it indicates than not the whole process but a single draft is affected: `Failed to process the CartDiscountDraft with key..., Reason...`

src/main/java/com/commercetools/sync/cartdiscounts/CartDiscountSync.java
```diff
@@ -202,11 +200,9 @@ private void handleError(@Nonnull final String errorMessage,
...
                         .thenCompose(resolvedDraft -> syncDraft(oldCartDiscountMap, resolvedDraft))
                         .exceptionally(completionException -> {
-                            final ReferenceResolutionException referenceResolutionException =
-                                (ReferenceResolutionException) completionException.getCause();
-                            final String errorMessage = format(FAILED_TO_RESOLVE_REFERENCES, newCartDiscount.getKey(),
-                                             referenceResolutionException.getMessage());
-                            handleError(errorMessage, referenceResolutionException, 1);
+                            final String errorMessage = format(FAILED_TO_RUN_SYNC_DRAFT, newCartDiscount.getKey(),
+                                completionException.getCause().getMessage());
+                            handleError(errorMessage, completionException.getCause(), 1);
```
butenkor, line 205: Why do path the getCause() here and not the whole completionException?

leungkinghin: CompletionException has a behaviour which always wrap the actual cause of exception. Here, if we use the whole completionException, .getMessage() method returns the wrapped exception classname in addition to the original message. It may bring the impact to current library consumer.

ahmetoz:
This is also related to your PR, I guess you did not change it because there is no cast in here:

https://github.com/commercetools/commercetools-sync-java/blob/70bb9eefb79d2a00f9780e42e1c78c43eb248376/src/main/java/com/commercetools/sync/categories/CategorySync.java#L294-L301

ahmetoz:
Do we know that it is always ReferenceResolutionException, if yes how?

https://github.com/commercetools/commercetools-sync-java/blob/70bb9eefb79d2a00f9780e42e1c78c43eb248376/src/main/java/com/commercetools/sync/inventories/InventorySync.java#L187-L190

Issue #499, opened 2020-04-03
ensurePriceChannels in ProductSyncOptions does not create a missing channel


butenkor:
**Describe the bug**
ensurePriceChannels in ProductSyncOptions does not create a missing channel used on product variant price draft.

**To Reproduce**
Use sync to/create a variant price with a missing price distribution channel and activate ensurePriceChannels in ProductSyncOptions.

Product-Sync produces the following request body:
```
{"key":"0000009039","roles":[]}]]
http request formatted body: {
  "key" : "0000009039",
  "roles" : [ ]
}
```

API response:
```
POST https://api.sphere.io/test-project/channels failed  with response code 400 on 2020-03-30T13:00:50.464094Z
http response formatted body: {
  "statusCode" : 400,
  "message" : "'roles' should not be empty.",
  "errors" : [ {
    "code" : "InvalidOperation",
    "message" : "'roles' should not be empty."
  } ]
}
```

**Expected behavior**
For Product-Sync, a channel with `ProductDistribution` role should be created. Depending on sync type it should assign different roles (inventory sync -> `InventorySupply`)

**Additional context**
SUPPORT-8414

Right now Product-Sync [creates](https://github.com/commercetools/commercetools-sync-java/blob/b152aa945df1a868ef7477ad2b6316c296120090/src/main/java/com/commercetools/sync/products/ProductSync.java#L88) the channel service with syncOptions only.

It leads that on channel [create](https://github.com/commercetools/commercetools-sync-java/blob/b152aa945df1a868ef7477ad2b6316c296120090/src/main/java/com/commercetools/sync/products/helpers/PriceReferenceResolver.java#L195) it would pass to the API an empty array (which in not allowed):
https://github.com/commercetools/commercetools-sync-java/blob/b152aa945df1a868ef7477ad2b6316c296120090/src/main/java/com/commercetools/sync/services/impl/ChannelServiceImpl.java#L37

Issue #449, opened 2019-09-24
Remove reference replacement of attribute references on ProductReferenceReplacement.


heshamMassoud:
### Description
Remove reference replacement of attribute references on ProductReferenceReplacement.

Related to https://github.com/commercetools/commercetools-project-sync/issues/38

PR #507, opened 2020-04-16
#449 ensurePriceChannels in ProductSyncOptions does not create a missing channel

JudeNiroshan:
#### Summary

ensurePriceChannels in ProductSyncOptions does not create a missing channel

#### Description

It was reported that we are submitting an invalid http request to CTP platform with roles: []. This happens when `ensureChannel` enabled in syncOptions, we are using ProductSync having a variant which has a price channel which cannot be resolved.

#### Relevant Issues

https://github.com/commercetools/commercetools-sync-java/issues/499
#### Todo

- Tests
    - [ ] Unit 
    - [x] Integration
- [ ] Documentation

#### Hints for Review

https://github.com/commercetools/commercetools-sync-java/blob/b152aa945df1a868ef7477ad2b6316c296120090/src/main/java/com/commercetools/sync/products/ProductSync.java#L88


src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncIT.java
```diff
@@ -179,6 +184,43 @@ void sync_withNewProduct_shouldCreateProduct() {
...
+
+        ProductSyncOptions syncOptions = ProductSyncOptionsBuilder.of(CTP_TARGET_CLIENT)
+                .errorCallback(this::collectErrors)
+                .warningCallback(warningCallBack)
+                .ensurePriceChannels(true)
+                .build();
+
+        final ProductSync productSync = new ProductSync(syncOptions);
+        final ProductSyncStatistics syncStatistics = executeBlocking(productSync.sync(singletonList(productDraft)));
+
```
ahalberkamp, line 217: It is possible, to check if the product of the target project really has the synced price with the required channelrole.

JudeNiroshan: de21697658fc7e5b6fa2828b295d98f7b57dc8c3

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -85,7 +87,7 @@ public ProductSync(@Nonnull final ProductSyncOptions productSyncOptions) {
             new ProductTypeServiceImpl(productSyncOptions),
             new CategoryServiceImpl(CategorySyncOptionsBuilder.of(productSyncOptions.getCtpClient()).build()),
             new TypeServiceImpl(productSyncOptions),
-            new ChannelServiceImpl(productSyncOptions),
+            new ChannelServiceImpl(productSyncOptions, Collections.singleton(ChannelRole.PRODUCT_DISTRIBUTION)),
```
ahmetoz, line 90: I wonder why not changing the default constructor instead of changing it here, as described on the issue before?

https://github.com/commercetools/commercetools-sync-java/blob/7baae24007cbb49afe7a6345bfee535765f3ac7b/src/main/java/com/commercetools/sync/services/impl/ChannelServiceImpl.java#L35-L38

ahmetoz: Also codecov report is pointing the same place:

Screenshot 2020-04-21 at 09 14 47

JudeNiroshan: I think enforcing the user to aware about the mandatory fields is better rather setting those values inside the constructor. With the current approach, user knows that is not possible to instantiate a ChannelService without knowing channel roles.

JudeNiroshan: abec68c

JudeNiroshan:
Closing this PR as there is a non-related commit has added in this branch. Check out this new PR which contains exactly only changes within this PR -> https://github.com/commercetools/commercetools-sync-java/pull/510

PR #508, opened 2020-04-17
Adding fix to sameForAll attributes + unit testing

ahmed-ali225:
#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [ ] Documentation

### DIFF
--- a/src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
+++ b/src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
@@ -37,7 +37,7 @@ import static com.commercetools.sync.products.utils.ProductUpdateActionUtils.bui
 
 public final class ProductSyncUtils {
 
-    public static final String REMOVE_VARIANT_ACTION_NAME = "removeVariant";
+    private static final String REMOVE_VARIANT_ACTION_NAME = "removeVariant";
     public static final String SET_ATTRIBUTE_IN_ALL_VARIANTS_ACTION_NAME = "setAttributeInAllVariants";
     public static final String ADD_VARIANT_ACTION_NAME = "addVariant";
     public static final String CHANGE_MASTER_VARIANT_ACTION_NAME = "changeMasterVariant";
--- a/src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
+++ b/src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
@@ -39,7 +39,7 @@ public final class ProductSyncUtils {
 
     private static final String REMOVE_VARIANT_ACTION_NAME = "removeVariant";
     private static final String SET_ATTRIBUTE_IN_ALL_VARIANTS_ACTION_NAME = "setAttributeInAllVariants";
-    public static final String ADD_VARIANT_ACTION_NAME = "addVariant";
+    private static final String ADD_VARIANT_ACTION_NAME = "addVariant";
     private static final String CHANGE_MASTER_VARIANT_ACTION_NAME = "changeMasterVariant";
 
     /**
--- a/src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
+++ b/src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
@@ -7,11 +7,14 @@ import com.commercetools.sync.products.SyncFilter;
 import io.sphere.sdk.commands.UpdateAction;
 import io.sphere.sdk.products.Product;
 import io.sphere.sdk.products.ProductDraft;
+import io.sphere.sdk.products.ProductVariant;
+import io.sphere.sdk.products.commands.updateactions.RemoveVariant;
 
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static com.commercetools.sync.commons.utils.CollectionUtils.emptyIfNull;
@@ -34,6 +37,12 @@ import static com.commercetools.sync.products.utils.ProductUpdateActionUtils.bui
 import static com.commercetools.sync.products.utils.ProductUpdateActionUtils.buildVariantsUpdateActions;
 
 public final class ProductSyncUtils {
+
+    public static final String REMOVE_VARIANT_ACTION_NAME = "removeVariant";
+    public static final String SET_ATTRIBUTE_IN_ALL_VARIANTS_ACTION_NAME = "setAttributeInAllVariants";
+    public static final String ADD_VARIANT_ACTION_NAME = "addVariant";
+    public static final String CHANGE_MASTER_VARIANT_ACTION_NAME = "changeMasterVariant";
+
     /**
      * Compares all the fields (including the variants see
      * {@link ProductUpdateActionUtils#buildVariantsUpdateActions(Product, ProductDraft, ProductSyncOptions, Map)})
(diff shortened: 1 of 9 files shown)


ahmed-ali225:
The issue briefly is that for the attributes with **“sameForAll”** constraint we found that there would be a generated UpdateAction per variant however in this case only one UpdateAction of type **SetAttributeInAllVariants** is sufficient.

Having more than UpdateAction generated for a product with _big number of variants_ will cause the **target product's version** to jump very high and will also cause an overhead on the APIs knowing that we sync a lot of products during the day.

Also, found that if there is any **"SetAttributeInAllVariants"** actions generated it has to be executed first before **"addVariant"** actions executed, otherwise this will cause an error in the target product as the old variants are still having a different value for the attribute of constraint type **"sameForAll"** and that's why **"SetAttributeInAllVariants"** has to be on top of the list of actions to be executed.

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
```diff
@@ -416,8 +418,8 @@
                 } else {
                     final ProductVariant matchingOldVariant = oldProductVariantsWithMaster.get(newProductVariantKey);
                     final List<UpdateAction<Product>> updateOrAddVariant = ofNullable(matchingOldVariant)
-                        .map(oldVariant -> collectAllVariantUpdateActions(oldProduct, oldVariant, newProductVariant,
-                            attributesMetaData, syncOptions))
+                        .map(oldVariant -> collectAllVariantUpdateActions(updateActions, oldProduct, oldVariant,
```
JudeNiroshan, line 421: Wouldn't it be better if we pass only the `SetAttributeInAllVariants` update actions? passing down the `updateActions` from this place will make additional iterations inside the inner method calls. We can even remove casting in filter operations. wdyt?

src/test/java/com/commercetools/sync/products/utils/ProductSyncUtilsTest.java
```diff
@@ -230,4 +237,43 @@ void buildActions_FromDraftsWithSameNameValues_ShouldNotBuildUpdateActions() {
 
         assertThat(updateActions).isEmpty();
     }
+
+    @Test
+    void buildActions_FromDraftsWithSameForAllAttribute_ShouldBuildUpdateActions() {
```
JudeNiroshan, line 242: It's better if we can have a test case which has no SameForAll attributes. Just a notify that we don't reduce our code coverage 😄

src/test/java/com/commercetools/sync/products/utils/ProductUpdateActionUtilsTest.java
```diff
@@ -66,6 +68,10 @@
     private static final String NEW_PROD_DRAFT_WITH_VARIANTS_REMOVE_MASTER =
         RES_ROOT + "productDraftNew_changeRemoveMasterVariant.json";
 
+    // this product's variants contain only attribute update
+    private static final String NEW_PROD_DRAFT_WITH_MATCHING_VARIANTS =
```
JudeNiroshan, line 72:
```suggestion
    private static final String NEW_PROD_DRAFT_WITH_MATCHING_VARIANTS_WITH_UPDATED_ATTR_VALUES =
```

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
ahmed-ali225: Yes sure, good idea.

src/test/java/com/commercetools/sync/products/utils/ProductSyncUtilsTest.java
ahmed-ali225: I think the rest of the test cases are not having sameForAll attributes.
Or did i miss your point?

ahmed-ali225: Yeah, i agree this code coverage was driving me crazy.
I added one more test "buildActions_FromDraftsWithDifferentAttributes_ShouldBuildUpdateActions()" for that in order to make sure other attribute updates can be duplicated unlike the sameForAll attributes.

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
```diff
@@ -416,8 +418,8 @@
                 } else {
                     final ProductVariant matchingOldVariant = oldProductVariantsWithMaster.get(newProductVariantKey);
                     final List<UpdateAction<Product>> updateOrAddVariant = ofNullable(matchingOldVariant)
-                        .map(oldVariant -> collectAllVariantUpdateActions(oldProduct, oldVariant, newProductVariant,
-                            attributesMetaData, syncOptions))
+                        .map(oldVariant -> collectAllVariantUpdateActions(getSameForAllUpdateActions(updateActions),
```
ahmetoz, line 421: I could not see what are you trying to do with reading that code, could you please consider refactoring the new collectAllVariantUpdateActions with a better name and handle the collect operation there 🤔 I assume there is a method overloading there but could not see the expected result 🤔

src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
```diff
@@ -104,7 +106,24 @@
...
-        return updateActions;
+        return prioritizeUpdateActions(updateActions);
+    }
+
+    private static List<UpdateAction<Product>> prioritizeUpdateActions(
+            final List<UpdateAction<Product>> updateActions) {
+
+        final List<UpdateAction<Product>> sameForAllUpdateActions =
+                emptyIfNull(updateActions)
+                        .parallelStream()
```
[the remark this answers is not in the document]
ahmed-ali225: ok, sure

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
ahmed-ali225: I tried to keep the same naming convention since it's and overload as you mentioned and only takes one extra parameter.

src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
```diff
@@ -104,7 +106,24 @@
         // lastly publish/unpublish product
         buildPublishUpdateAction(oldProduct, newProduct).ifPresent(updateActions::add);
 
-        return updateActions;
+        return prioritizeUpdateActions(updateActions);
+    }
+
+    private static List<UpdateAction<Product>> prioritizeUpdateActions(
```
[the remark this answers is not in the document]
ahmed-ali225: No, it doesn't make any problems.
I tried it my self and e2e is working fine in this regard.
Basically making the "removeVariant" on top of the list of the update actions will cause some issues when it comes to changing the master variant on the target product.
For instance, changing the master variant would involve the following set of actions:
"addVariant"
"changeMasterVariant"
"removeVariant"

so, in this case making "removeVariant" will cause an error saying that you can't remove the variant because it's a master variant.

So, i saw that the most important action to be on top of the list of update actions is the "setAttributeInAllVariants" because you want to keep the target product consistent first before applying "addVariant" actions for example and it doesn't cause any problems whe it comes to "removeVariant".

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
```diff
@@ -441,6 +450,36 @@
         return allVariants;
     }
 
+    private static boolean filterDuplicateSameForAllAction(
```
butenkor, line 453: Why not name parameter `updateActions` the same way as it is passed to the method: `sameForAllUpdateActions`? Also, the method returns a boolean - maybe then we should name it accordingly i.e.: `hasDuplicateSameForAllAction`?

butenkor: In general, it feels like duplicate sameForAll actions should be omitted here (https://github.com/commercetools/commercetools-sync-java/pull/508/files#diff-540361586e271c4157d1afe06f9c8636R494) instead of first generating them and later filtering them out?

src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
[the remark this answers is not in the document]
ahmed-ali225: I totally agree to this, I just wanted to keep it simple for now but i can make the changes accordingly.

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
ahmed-ali225: ok, sounds gound.

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
ahmed-ali225: I'm sorry the link provided doesn't give me a clue as it opens the pull request again

src/main/java/com/commercetools/sync/products/utils/ProductSyncUtils.java
```diff
@@ -104,7 +114,69 @@
...
+                getActionsByActionName(updateActions, action ->
+                        action.getAction().equals(REMOVE_VARIANT_ACTION_NAME)
+                                && action.equals(removeMasterVariantUpdateAction));
+
+        final List<UpdateAction<Product>> updateActionList = new ArrayList<>(removeVariantUpdateActionsNoMaster);
+        updateActionList.addAll(sameForAllUpdateActions);
+        updateActionList.addAll(addVariantUpdateActions);
+        updateActionList.addAll(changeMasterUpdateActions);
+        updateActionList.addAll(removeOldMasterVariantUpdateAction);
+        updateActionList.addAll(updateActions);
```
ahmetoz, line 164: I wonder why we are applying other actions after removeMaster, isn't it better after SameForAllUpdateActions?

ahmed-ali225: In order to remove the old master variant you have to add the new variant first and change it to master then the CTP will allow the removal of the old master variant as it's not a master anymore...the rest of update actions comes latest after the removal of the old master variant just to avoid applying any other updates to a variant that is meant to be removed later.
Hope that this answers your question.
Thanks

PR #517, opened 2020-06-12
update benchmark triggering flow

ahmetoz:
Before to able to trigger a benchmark test, we would need to push a commit into master like `Prepare for release v1.8.2`, This is not easy to do manage, also when this is designed in the first place benchmarks were running with 10.000 thresholds, now it's 1000. so it should not be an issue for the platform. In this PR also there is a change related to benchmarkCommit task (`gh-pages/benchmarks.json` commit changes should be trigger only with a release tag)

Todo:
- [ ] It's tested locally but not in Travis and with a release tag, those tests could be done when the PR is merged to master branch, then we could try a temporary/pre-release.

- [x] Update all dependency references in documentations.
- [x] Update build & release documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ install: true # skips travis' default installation step which executes gradle as
 jobs:
   include:
     - stage: benchmark
-      if: branch = master AND commit_message =~ /^(Prepare for release .*)/
-      script: ./gradlew clean setLibraryVersion benchmark benchmarkCommit
+      if: branch = master
+      script: ./gradlew clean setLibraryVersion benchmark
     - stage: full build
       script: ./gradlew clean dependencyUpdates setLibraryVersion build
       after_success:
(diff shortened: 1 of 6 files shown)


ahmetoz:
The flaky test found in :

    ✘ sync_WithErrorCreatingTheProductType_ShouldExecuteCallbackOnErrorAndIncreaseFailedCounter()
      java.util.concurrent.CompletionException: io.sphere.sdk.client.ErrorResponseException: detailMessage: A duplicate value '"key_1"' exists for field 'key'.
      summary: POST https://api.europe-west1.gcp.commercetools.com/java-sync-target/product-types failed  with response code 400 with X-Correlation-ID `java-sync-target/4a676b7d-8b83-4a4c-b761-a66c5b5aad65` on 2020-07-02T08:12:11.307Z
        "statusCode" : 400,
        "message" : "A duplicate value '"key_1"' exists for field 'key'.",
        "errors" : [ {
          "code" : "DuplicateField",
          "message" : "A duplicate value '"key_1"' exists for field 'key'.",
          "duplicateValue" : "key_1",
          "field" : "key"
        } ]
      }

Issue #231, opened 2018-01-16
New CTP Feature: parent of CategoryDraft is now a ResourceIdentifier instead of a Reference


heshamMassoud:
### Description
https://dev.commercetools.com/release-notes.html#release-notes---commercetools-platform---version-release-15-january-2018

### Consequences

- Reference of category parents of keys -> ids should not be needed anymore.
- Consistent logic to category type reference resolution.
- possible increase in efficiency reference resolution, since there is no need to fetch ids.

### Related
#166

butenkor:
> possible increase in efficiency reference resolution, since there is no need to fetch ids

Not sure about that as you still need to fetch in order to know that desired key reference exists.

ahmetoz:
Released with: https://github.com/commercetools/commercetools-sync-java/blob/master/docs/RELEASE_NOTES.md#200---sept-14-2020

PR #541, opened 2020-08-25
231 category identifier

ahalberkamp:
#### Summary
In this pr, the category-parent resolution was refactored, so that it is done using keys instead of id

#### Relevant Issues
https://github.com/commercetools/commercetools-sync-java/issues/231

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [ ] Documentation

#### Hints for Review
When getting  the key of a reference to a parent-category a "null" is always returned (see:https://github.com/commercetools/commercetools-jvm-sdk/blob/93b924410f3ae1900baf0232223e098638a0e869/commercetools-sdk-base/src/main/java/io/sphere/sdk/models/Reference.java#L103)

The reason for that is that the  API returns only references, which only has an ID. When creating a category draft this is not a problem, because the draft supports resource identifier, which supports keys. But if the key of an existing Reference should be fetched, this reference has to be exspanded, and the key has to be fetched from its object


src/integration-test/java/com/commercetools/sync/integration/commons/utils/CategoryITUtils.java
```diff
@@ -130,24 +129,27 @@
...
+            CategoryDraftBuilder childBuilder = CategoryDraftBuilder
                 .of(LocalizedString.of(Locale.ENGLISH, categoryName),
                     LocalizedString.of(Locale.ENGLISH, categoryName))
                 .key(categoryName)
-                .parent(toResourceIdentifierIfNotNull(parent))
                 .custom(CustomFieldsDraft.ofTypeKeyAndJson(OLD_CATEGORY_CUSTOM_TYPE_KEY, createCustomFieldsJsonMap()))
-                .orderHint("sameOrderHint")
-                .build();
+                .orderHint("sameOrderHint");
+
```
ahalberkamp, line 138:
```suggestion
```

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -272,20 +271,20 @@ private int getNumberOfDraftsToProcess(@Nonnull final List<CategoryDraft> catego
      * resolution, the error callback is triggered and the category is skipped.
      *
      * @param categoryDrafts the input list of category drafts in the sync batch.
-     * @param keyToIdCache   the cache containing the mapping of all existing category keys to ids.
+     * @param existingCategoryKeys  List of keys of the ca.
```
ahalberkamp, line 274:
```suggestion
     * @param existingCategoryKeys  List of keys of categories.
```

src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
```diff
@@ -71,19 +72,28 @@
...
+     * @param useKeys               if true, the keys of the identifier are used for the comparison
      * @param <T>                   the type of the old resource identifier
      * @param <S>                   the type of the new resource identifier
      * @return true or false depending if the resource identifiers have the same id.
      */
     public static <T extends ResourceIdentifier, S extends ResourceIdentifier> boolean areResourceIdentifiersEqual(
-        @Nullable final T oldResourceIdentifier, @Nullable final S newResourceIdentifier) {
+        @Nullable final T oldResourceIdentifier,
+        @Nullable final S newResourceIdentifier,
+        boolean useKeys) {
```
ahmetoz, line 83: I would add a new method for this instead of passing a boolean check. So we could remove the old method when we don't need reference resolution after all resources support ResourceIdentifiers. It's also a smell in here, I could not see a reason to use a generic static utility method to get the key of the ResourceIdentifier with the method getKeyOfResourceIdentifier before we don't know the exact type but now we know the type. Am I wrong ? 🤔

src/main/java/com/commercetools/sync/services/CategoryService.java
```diff
@@ -24,7 +24,7 @@
      *         category keys -&gt; ids
      */
     @Nonnull
-    CompletionStage<Map<String, String>> cacheKeysToIds();
+    CompletionStage<Set<String>> loadExistingCategoryKeys();
```
ahmetoz, line 27: I would not change this method's name and structure as this method will be changed later and we need to fetch only keys from the batch for better performance. If you need a set maybe change it in the CategorySync.

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -184,14 +185,14 @@ public CategorySync(@Nonnull final CategorySyncOptions syncOptions) {
                         return CompletableFuture.completedFuture(null);
                     }
 
-                    prepareDraftsForProcessing(categoryDrafts, keyToIdCache);
```
ahmetoz, line 187: javadocs are not up to date... check keyToIdCaches in code.

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
```diff
@@ -18,9 +18,10 @@
  *            specified by the user, on reference resolution.
  */
 public abstract class BaseReferenceResolver<T, S extends BaseSyncOptions> {
+    public static final String BLANK_KEY_VALUE_ON_RESOURCE_IDENTIFIER = "The value of the 'key' field of the Resource"
```
ahmetoz, line 21: The message could be improved. Because there is no key field on the Reference. 
`The value of the 'key' field of the ResourceIdentifier is blank (null/empty).`

src/main/java/com/commercetools/sync/categories/CategorySync.java
```diff
@@ -465,23 +463,24 @@ private void processFetchedCategories(@Nonnull final Set<Category> fetchedCatego
...
-                              categoryDraftBuilder
-                                  .parent(Category.referenceOfId(parentId).toResourceIdentifier());
+                              categoryDraftBuilder.parent(ofKey(missingParentKey));
                           });
             }
             categoryDraftsToUpdate.put(categoryDraftBuilder.build(), fetchedCategory);
         });
     }
 
+    private <T extends Reference<Category>> ResourceIdentifier<Category> getKeyReferenceIfNotNull( final T parent) {
```
ahmetoz, line 481: why you need this utility ?

src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/categories/updateactionutils/ChangeParentIT.java
```diff
@@ -42,10 +42,12 @@ static void setup() {
         deleteAllCategories(CTP_SOURCE_CLIENT);
```
ahmetoz, line 42: Those ITs that you changed is removed in master, so you need to sync branch with master

ahmetoz (changes requested):
Could you give an example integration test to ensure that is working with the correct drafts, also the build is failing and no clue about the code coverage.

Also, there is no documentation and release notes, are you planning to add it with a different PR?

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
```diff
@@ -96,27 +91,6 @@ public CategoryReferenceResolver(@Nonnull final CategorySyncOptions options,
             format(FAILED_TO_RESOLVE_CUSTOM_TYPE, draftBuilder.getKey()));
     }
 
-    /**
```
ahmetoz, line 99: We should not remove the reference resolution for parent category

src/main/java/com/commercetools/sync/categories/utils/CategoryReferenceReplacementUtils.java
```diff
@@ -46,7 +46,7 @@ private CategoryReferenceReplacementUtils() {
                 final CustomFieldsDraft customTypeWithKeysInReference = replaceCustomTypeIdWithKeys(category);
                 @SuppressWarnings("ConstantConditions") // NPE checked in replaceReferenceIdWithKey
                 final ResourceIdentifier<Category> parentWithKeyInReference = getResourceIdentifierWithKeyReplaced(
-                    category.getParent(), () -> ResourceIdentifier.ofId(category.getParent().getObj().getKey()));
+                    category.getParent(), () -> ResourceIdentifier.ofKey(category.getParent().getObj().getKey()));
```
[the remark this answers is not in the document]
ahalberkamp: 277b2da8

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
```diff
@@ -48,16 +49,39 @@ protected BaseReferenceResolver(@Nonnull final S options) {
...
+            throw new ReferenceResolutionException(BLANK_ID_VALUE_ON_RESOURCE_IDENTIFIER);
+        }
+        return id;
+    }
+
+    /**
+     * This method fetches the key value on the passed {@link ResourceIdentifier}, if valid. If it is not valid, a
+     * {@link ReferenceResolutionException} will be thrown. The validity checks are:
+     * <ul>
+     * <li>Checks if the id value is not null or not empty.</li>
```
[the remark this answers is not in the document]
ahalberkamp: 9831d358

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
```diff
@@ -48,16 +49,39 @@ protected BaseReferenceResolver(@Nonnull final S options) {
      * If the above checks pass, the key value is returned. Otherwise a {@link ReferenceResolutionException} is thrown.
      *
      * @param resourceIdentifier the reference from which the key value is validated and returned.
+     * @return the Id value on the {@link ResourceIdentifier}
+     * @throws ReferenceResolutionException if any of the validation checks fail.
+     */
+    @Nonnull
+    protected static String getdIdFromResourceIdentifier(@Nonnull final ResourceIdentifier resourceIdentifier)
```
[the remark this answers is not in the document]
ahalberkamp: Yes we can get rid of the method, when we have refactored the reference resolution concept completly, but for now we have to keep this method 715ecd1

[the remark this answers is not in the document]
ahalberkamp: 715ecd1

src/main/java/com/commercetools/sync/commons/utils/CommonTypeUpdateActionUtils.java
ahalberkamp: it doesn`t smell because ,we don`t now if the given Resource identifier is a Reference. if yes, we have to fetch its key from its objects, otherwise we can get the key from the resource identifier directly

ahalberkamp: I would leave the boolean check in the method, because right now we can use the same method, for key-  and id comparison. And when we have finished the refactoring of the resolutionresolving, we can simply remove the boolean check than.

src/main/java/com/commercetools/sync/services/CategoryService.java
ahalberkamp: 85e84de1

src/main/java/com/commercetools/sync/categories/CategorySync.java
ahalberkamp: 94a6821d

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
ahalberkamp: 184e2e5f

src/main/java/com/commercetools/sync/categories/CategorySync.java
ahalberkamp: .560bdfe

src/integration-test/java/com/commercetools/sync/integration/ctpprojectsource/categories/updateactionutils/ChangeParentIT.java
ahalberkamp: done !!

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
ahmetoz: `ResourceIdentifier` is a type, so I would use the type name instead of writing as `resource identifier`

src/integration-test/java/com/commercetools/sync/integration/externalsource/categories/CategorySyncIT.java
```diff
@@ -351,6 +357,24 @@ void syncDrafts_WithNewCategoryWithExistingParent_ShouldCreateCategory() {
...
     }
 
+    @Test
+    void syncDrafts_WithNewCategoryWithExistingParentWithKey_ShouldNotCreateCategory() {
+        // Category draft coming from external source.
+        final CategoryDraft categoryDraft = CategoryDraftBuilder
+            .of(LocalizedString.of(Locale.ENGLISH, "Modern Furniture"),
+                LocalizedString.of(Locale.ENGLISH, "modern-furniture"))
+            .key("newCategory")
+            .parent(ResourceIdentifier.ofId(oldCategoryKey))
```
ahmetoz, line 367: this test is confusing, why you are adding parent key into id field ?

ahalberkamp: 8b42fb59

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
ahalberkamp: c1b406d4

src/main/java/com/commercetools/sync/commons/helpers/BaseReferenceResolver.java
ahalberkamp: ea7bb7e4

src/main/java/com/commercetools/sync/categories/helpers/CategoryReferenceResolver.java
```diff
@@ -164,26 +160,4 @@ public CategoryReferenceResolver(@Nonnull final CategorySyncOptions options,
...
-     * id. If the id is not found in cache nor the CTP project, the resultant draft builder
-     * would remain exactly the same as the passed category draft (without parent reference resolution).
-     *
-     * @param draftBuilder      the category draft builder to accept resolved references values.
-     * @param parentCategoryKey the parent category key of to resolve it's actual id on the draft.
-     * @return a {@link CompletionStage} that contains as a result the same {@code draftBuilder} category draft builder
-     *         instance with resolved parent category references or an exception.
-     */
-    @Nonnull
-    private CompletionStage<CategoryDraftBuilder> fetchAndResolveParentReference(
```
ahmetoz, line 179: Why do you want to remove this code?

src/main/java/com/commercetools/sync/services/impl/CategoryServiceImpl.java
```diff
@@ -75,7 +77,9 @@ public CategoryServiceImpl(@Nonnull final CategorySyncOptions syncOptions) {
         return fetchMatchingResources(categoryKeys,
             () -> CategoryQuery
                 .of()
-                .plusPredicates(categoryQueryModel -> categoryQueryModel.key().isIn(categoryKeys)));
+                .plusPredicates(categoryQueryModel -> categoryQueryModel.key().isIn(categoryKeys))
+                .plusExpansionPaths(CategoryExpansionModel::parent));
```
ahmetoz, line 81: The expansion should effect performance, I assume we don't need to change CategoryService to use key instead of Id.

ahalberkamp:
1) example integration test 

1.1) Syncing is working ,all parentreferences has keys
https://github.com/commercetools/commercetools-sync-java/pull/541/files#diff-7a2564580078e482081eea430630ad3dR344
1.2) Syncing is not working ,the  parentreference has only an ID but no key
https://github.com/commercetools/commercetools-sync-java/pull/541/files#diff-7a2564580078e482081eea430630ad3dR344

2)Release note

-> 49bbfa97

3) Documentation

-> 6cbe987c

ahmetoz:
PR will be handled by, #551

Issue #138, opened 2017-10-10
Support reference resolution in the key field not only id in ResourceIdentifiers


heshamMassoud:
**Edit: The task is resolved, so only parts are related to the API, not SDK anymore, when API adds support of ResourceIdentifier for PriceDraft#customerGroup we could close the issue. (Currently it's not clear when API will support this.)**

**Remaining work:**

Support requests (SUPPORT-9797, SUPPORT-9798) are opened for the problematic types/fields both API and SDK. Additionally, a GitHub issue is opened for jvm-sdk types: commercetools/commercetools-jvm-sdk#2052 (**Edit: The task is resolved, so only parts are related to the API, not SDK anymore.)**

----
### Description
Right now there are references which are of type `Reference` and other references which are of type `ResourceIdentifier`
1. `Reference` only supports `id` as a field.
2. `ResourceIdentifier`  supports `id` or `key` as a field.

On CTP in the future, all references should be changed to `ResourceIdentifier`.

It's quite confusing for the user of the sync library to put the key in the id field for reference resolution.

### Current Behaviour
Whether it's a resourceIdentifier or reference reference resolution is always done on the id field.

### Expected Behaviour
1. When the reference is of type `ResourceIdentifier`:
     - If the key field is set, sync lib should check if resource exists or not, and keep track of missing references (as done in cat sync) as it might be resolved in later batches.
     - If the id field is set, then it is an evidence of resource existence on CTP, so we can issue an update/create API request right away.
2. When the reference is of type `Reference`:
     - Do reference resolution as it currently is (key -> id mapping).

Related: #231

butenkor:
As defined by @ahmetoz 

# References on commercetools-sync-java 

## [ProductDraft](https://docs.commercetools.com/http-api-projects-products.html#productdraft)
| field | commercetools type | required/optional | JVM SDK draft type | status | 
| --- | --- | --- | -- | -- | 
| productType  | ResourceIdentifier to a ProductType | Required   | `ResourceIdentifier<ProductType>`  | :heavy_check_mark: |
| categories  | Array of ResourceIdentifier for a Category  | Optional  | `Set<ResourceIdentifier<Category>>` | :heavy_check_mark: |
| taxCategory | ResourceIdentifier to a TaxCategory   | Optional |` ResourceIdentifier<TaxCategory>` | :heavy_check_mark: |
| state  | ResourceIdentifier to a State | Optional | ` Reference<State>` | :warning:  |

[...]

PR #545, opened 2020-08-27
Use ResourceIdentifier's key field for CustomType reference resolution instead of using id field as a workaround.

ahmetoz:
#### Description
For reference resolution, the library user needs to replace Id with Key for Reference and ResourceIdentifier types **(or the user could use our exposed utilities like replaceCartDiscountsReferenceIdsWithKeys)** 
For CustomFieldsDraft of field type (ResourceIdentifier) it's not needed anymore because ResourceIdentifier has a key field already so that confusing replacement is not needed anymore as a workaround for most of the types. 

This PR is targeting to change only the custom fields, as it's used by almost all sync classes. Then with other PRs, we will be targeting to change other replacement utils. This branch will be a base branch for other resources like CategorySync, ProductSync, and InventorySync.

For more details refer to the ticket: #138 also check the https://github.com/commercetools/commercetools-sync-java/issues/138#issuecomment-680761023 for the latest map.

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [ ] Documentation
    - [ ] Ensure javadocs are clear.
    - [ ] Prepare the migration guide/update docs for the breaking changes. 
- [ ] Complete the parts commented with `todo (ahmetoz)`

### DIFF
--- a/src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
@@ -7,7 +7,7 @@ import io.sphere.sdk.models.AssetDraftBuilder;
 import javax.annotation.Nonnull;
 import java.util.List;
 
-import static com.commercetools.sync.commons.utils.CustomTypeReferenceReplacementUtils.replaceCustomTypeIdWithKeys;
+import static com.commercetools.sync.commons.utils.CustomTypeReferenceReplacementUtils.mapToCustomFieldsDraft;
 import static java.util.stream.Collectors.toList;
 
 /**
@@ -31,7 +31,7 @@ public final class AssetReferenceReplacementUtils {
     public static List<AssetDraft> replaceAssetsReferencesIdsWithKeys(@Nonnull final List<Asset> assets) {
         return assets.stream().map(asset ->
             AssetDraftBuilder.of(asset)
-                             .custom(replaceCustomTypeIdWithKeys(asset)).build())
+                             .custom(mapToCustomFieldsDraft(asset)).build())
                      .collect(toList());
     }
 
--- a/src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
+++ b/src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
@@ -18,17 +18,16 @@ public final class AssetReferenceReplacementUtils {
 
     /**
      * Takes an asset list that is supposed to have all its assets' custom references expanded in order to be able to
-     * fetch the keys and replace the reference ids with the corresponding keys for the custom references. This method
-     * returns as a result a {@link List} of {@link AssetDraft} that has all custom references with keys replacing the
-     * ids.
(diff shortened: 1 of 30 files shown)


src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
```diff
@@ -18,20 +18,19 @@
 
     /**
      * Takes an asset list that is supposed to have all its assets' custom references expanded in order to be able to
-     * fetch the keys and replace the reference ids with the corresponding keys for the custom references. This method
-     * returns as a result a {@link List} of {@link AssetDraft} that has all custom references with keys replacing the
-     * ids.
+     * fetch the keys for the custom references. This method returns as a result a {@link List} of {@link AssetDraft}
```
lojzatran, line 21:
```suggestion
     * get the keys for the custom references. This method returns as a result a {@link List} of {@link AssetDraft}
```

lojzatran: `fetch` means go somewhere and get it (https://www.google.com/search?q=define%20fetch). I would use it for methods where you need to fetch the resource from CTP, but here it doesn't do that. I saw lots of methods use fetch in this context, so it's up to you if you want to fix it.

src/test/java/com/commercetools/sync/commons/utils/ResourceCustomUpdateActionUtilsTest.java
```diff
@@ -96,8 +98,13 @@ void buildResourceCustomUpdateActions_WithNullOldCustomFields_ShouldBuildUpdateA
         final Category oldCategory = mock(Category.class);
         when(oldCategory.getCustom()).thenReturn(null);
 
-        final CategoryDraft newCategoryDraft = CategorySyncMockUtils.getMockCategoryDraft(Locale.ENGLISH, "name",
-            "key", "parentId", "customTypeId", new HashMap<>());
+        final CategoryDraft newCategoryDraft =
+            CategoryDraftBuilder.of(LocalizedString.ofEnglish("name"), LocalizedString.ofEnglish("testSlug"))
```
lojzatran, line 102: Why not using the method `CategorySyncMockUtils.getMockCategoryDraft`?

lojzatran (approved):

src/test/java/com/commercetools/sync/commons/utils/ResourceCustomUpdateActionUtilsTest.java
ahmetoz: It's used for creating a mock with type id (to created directly without sync methods), so I use the typical draft builder, creating a common utility is not needed for this case.

src/main/java/com/commercetools/sync/commons/utils/AssetReferenceReplacementUtils.java
ahmetoz: Thanks, I would note this is not the only place that we have this misusage.

ahmetoz:
PR will be handled by, #551

PR #575, opened 2020-09-22
Implement customer service

salander85:
#### Summary
Implementation of customer service which is acting as REST client to fetch, update, and create resources in commercetools API with considering sync needs.

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [x] Documentation


src/main/java/com/commercetools/sync/services/impl/CustomerServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    @Override
+    public CompletionStage<Customer> updateCustomer(@Nonnull final Customer customer,
+                                                    @Nonnull final
+                                                    List<UpdateAction<Customer>> updateActions) {
+        return updateResource(customer, CustomerUpdateCommand::of, updateActions);
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Customer> changePassword(@Nonnull final Customer customer,
```
ahmetoz, line 117: A change password call is not needed for the sync.

src/main/java/com/commercetools/sync/services/impl/CustomerServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+            @Nonnull final Set<String> customerKeys) {
+        return fetchMatchingResources(customerKeys, () -> CustomerQuery
+                        .of()
+                        .withPredicates(buildResourceKeysQueryPredicate(customerKeys)));
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Optional<Customer>> fetchCustomerByKey(@Nullable final String key) {
+        return fetchResource(key, () -> CustomerQuery
```
ahmetoz, line 59: JVM Sdk supports the key escaping in default so we might not to do it by ourselves, maybe it's better to use it as below:

````java
    public CompletionStage<Optional<Customer>> fetchCustomerByKey(@Nullable final String key) {
        return fetchResource(key, () -> CustomerQueryBuilder
            .of()
            .plusPredicates(queryModel -> queryModel.key().is(key))
            .build());
    }
````

File: src/main/java/com/commercetools/sync/services/impl/CustomerServiceImpl.java [hunk 3]
```diff
@@ -0,0 +1,128 @@
...
+                .of()
+                .withPredicates(buildResourceKeysQueryPredicate(singleton(key))));
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Optional<String>> fetchCachedCustomerId(@Nonnull final String key) {
+        return fetchCachedResourceId(key, () -> CustomerQuery
```
ahmetoz, line 67: JVM Sdk supports the key escaping in default so we might not to do it by ourselves, maybe it's better to use it as below:

````java
    public CompletionStage<Optional<String>> fetchCachedCustomerId(@Nonnull final String key) {
        return fetchCachedResourceId(
            key,
            () -> CustomerQueryBuilder
                .of()
                .plusPredicates(queryModel -> queryModel.key().is(key))
                .build());
    }
````

File: src/main/java/com/commercetools/sync/services/impl/CustomerServiceImpl.java [hunk 4]
```diff
@@ -0,0 +1,128 @@
...
+        super(syncOptions);
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Map<String, String>> cacheKeysToIds(
+            @Nonnull final Set<String> keysToCache) {
+        return cacheKeysToIds(keysToCache, keysNotCached -> CustomerQuery
+                        .of()
+                        .withPredicates(buildResourceKeysQueryPredicate(keysNotCached)));
```
ahmetoz, line 44: same here.

src/main/java/com/commercetools/sync/services/impl/CustomerServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Set<Customer>> fetchMatchingCustomersByKeys(
+            @Nonnull final Set<String> customerKeys) {
+        return fetchMatchingResources(customerKeys, () -> CustomerQuery
+                        .of()
+                        .withPredicates(buildResourceKeysQueryPredicate(customerKeys)));
```
ahmetoz, line 53: same here.

src/main/java/com/commercetools/sync/services/impl/CustomerServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    @Override
+    public CompletionStage<Optional<String>> fetchCachedCustomerId(@Nonnull final String key) {
+        return fetchCachedResourceId(key, () -> CustomerQuery
+                .of()
+                .withPredicates(buildResourceKeysQueryPredicate(singleton(key))));
+    }
+
+    @Nonnull
+    @Override
+    public CompletionStage<Optional<Customer>> createCustomer(
```
ahmetoz, line 74: I guess you need this because the CustomerCreateCommand does not have the right type and can not be changed because it's a different library. But I would suggest here to add a comment on why you have not use the base implementation.

src/main/java/com/commercetools/sync/services/CustomerService.java
```diff
@@ -0,0 +1,123 @@
...
+    @Nonnull
+    CompletionStage<Map<String, String>> cacheKeysToIds(@Nonnull Set<String> keysToCache);
+
+    /**
+     * Given a {@link Set} of customer keys, this method fetches a set of all the customers, matching the given set of
+     * keys in the CTP project, defined in an injected {@link SphereClient}. A mapping of the key to the id of the
+     * fetched customers is persisted in an in-memory map.
+     *
+     * @param customerKeys set of customer keys to fetch matching resources by.
+     * @return {@link CompletionStage}&lt;{@link Set}&lt;{@code U}&gt;&gt; in which the result of it's completion
```
ahmetoz, line 34: Actually I don't know what is {@code U} could be a copy-paste issue from the old docs 🤔

src/main/java/com/commercetools/sync/services/CustomerService.java
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param customer  the {@link Customer} to change the password.
+     * @param currentPassword the current password of the customer
+     * @param newPassword the new password to change to.
+     * @return  {@link CompletionStage}&lt;{@link Customer}&gt; containing as a result of it's completion an instance
+     *          of the {@link Customer} which was updated in the CTP project or a
+     *          {@link io.sphere.sdk.models.SphereException}.
+     */
+    @Nonnull
+    CompletionStage<Customer> changePassword(@Nonnull Customer customer, @Nonnull String currentPassword,
```
ahmetoz, line 121: sorry but we don't need this 🤷🏽

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
```diff
@@ -20,9 +18,8 @@
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
-import java.util.stream.Collectors;
 
-import static java.lang.String.format;
+import static com.commercetools.sync.commons.utils.CtpQueryUtils.buildResourceKeysQueryPredicate;
```
ahmetoz, line 22: Honestly, I am not seeing any reason to change the ProductService also the need for making the function common, maybe you could explain ?

ahmetoz (changes requested):

src/integration-test/java/com/commercetools/sync/integration/services/impl/CustomerServiceImplIT.java
```diff
@@ -0,0 +1,463 @@
...
+        final Optional<String> customerId = customerService.fetchCachedCustomerId(EXISTING_CUSTOMER_KEY)
+                .toCompletableFuture()
+                .join();
+        assertThat(customerId).isNotEmpty();
+        assertThat(errorCallBackExceptions).isEmpty();
+        assertThat(errorCallBackMessages).isEmpty();
+    }
+
+    @Test
+    void fetchCachedCustomerId_WithCachedExistingCustomer_ShouldFetchFromCache() {
```
lojzatran, line 116: I think some of these tests should rather be unit tests. Integration tests should be only minimum amount of tests as they're slow. In these tests you always need to create before the tests start a new customer and then clean up, and some tests don't need that (like `fetchCachedCustomerId_WithUnexpectedException_ShouldFail`). Also with this test `fetchCachedCustomerId_WithCachedExistingCustomer_ShouldFetchFromCache` you don't know if it really fetches from cache or not. Additionally in the last sprint we just cleaned up some integration tests as they were too slow, so it's good not to add too many of them again.

So I recommend to move most of the tests here to unit test and mock the request/response and leave here only a few tests to ensure that the functionality works.

lojzatran (changes requested):

lojzatran:
Btw not sure what is the policy here, but maybe don't merge it to master until the whole feature is finished? What is the process @ahmetoz ?

src/main/java/com/commercetools/sync/services/CustomerService.java
```diff
@@ -0,0 +1,123 @@
...
+     * <p>On the other hand, if the resource gets created successfully on CTP, the created resource's id and
+     * key are cached and the method returns a {@link CompletionStage} in which the result of it's completion
+     * contains an instance {@link Optional} of the resource which was created.
+     *
+     * @param customerDraft the resource draft to create a resource based off of.
+     * @return a {@link CompletionStage} containing an optional with the created resource if successful otherwise an
+     *         empty optional.
+     */
+    @Nonnull
+    CompletionStage<Optional<Customer>> createCustomer(@Nonnull CustomerDraft customerDraft);
```
lojzatran, line 88: There's also create customer in a store. Do we need to consider that as well? https://docs.commercetools.com/http-api-projects-customers#create-customer-sign-up-in-a-store-beta

salander85: haven't considered it yet because its in a beta state. @ahmetoz do we need to consider that?

ahmetoz: We will do the syncing for customers not store customers. Some endpoints in customer is more related to the UI usages. (/{projectKey}/in-store/key={storeKey}/customers) as you could see the endpoint is not directly a customer endpoint like (/{projectKey}/customers/{id})

ahmetoz:
> Btw not sure what is the policy here, but maybe don't merge it to master until the whole feature is finished? What is the process @ahmetoz ?

Correct, it's not complex process, and imo similar for all types of open source projects, don't merge anything into master if it's not ready to be deployed.

cc: @salander85

src/main/java/com/commercetools/sync/services/impl/CustomerServiceImpl.java
salander85: @lojzatran cannot reuse it, because `CustomerCreateCommand` replies with `CustomerSignInResult` which does not extend `ResourceView` which is used in `DraftBasedCreateCommand` to resolve the result into a resource object. So, @ahmetoz is right; will add a comment.

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
salander85: My intension was to re-use in several areas, but as im using built-in predicates from the JVM SDK now i reverted this change.

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
```diff
@@ -47,15 +50,26 @@ public ProductServiceImpl(@Nonnull final ProductSyncOptions syncOptions) {
...
                 .of()
-                .withPredicates(buildResourceKeysQueryPredicate(keysNotCached)));
+                .withPredicates(buildProductKeysQueryPredicate(keysNotCached)));
+    }
+
+    QueryPredicate<Product> buildProductKeysQueryPredicate(@Nonnull final Set<String> productKeys) {
+        final List<String> keysSurroundedWithDoubleQuotes = productKeys.stream()
+                                                                       .filter(StringUtils::isNotBlank)
+                                                                       .map(productKey -> format("\"%s\"", productKey))
+                                                                       .collect(Collectors.toList());
```
lojzatran: Can you change this to `Collectors.joining(",")` and then you can remove the lines:
```
        String keysQueryString = keysSurroundedWithDoubleQuotes.toString();
        // Strip square brackets from list string. For example: ["key1", "key2"] -> "key1", "key2"
        keysQueryString = keysQueryString.substring(1, keysQueryString.length() - 1);
```

lojzatran (approved):

ahmetoz (approved):

Issue #580, opened 2020-09-29
Bug in product sync while fixing unresolved references


ahmetoz:
**Describe the bug**
ProductSync is not resolving references correctly because of the missing id in the cache.
We need all cache of the key to ids for finding missing references during product sync, test cases were missing the case because for the product sync test cases were not working with multiple batches.

**To Reproduce**
Run ProductSync with batches or update batchSize of the syncOptions to ensure references exist in different parts.

ahmetoz:
Released with v2.2.0

PR #581, opened 2020-09-29
Fix in product sync while fixing unresolved references.

ahmetoz:
#580 

- Tests
    - [X] Unit 
    - [X] Integration
- [X] Documentation

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncWithReferencedProductsInAnyOrderIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncWithReferencedProductsInAnyOrderIT.java
@@ -112,6 +112,7 @@ class ProductSyncWithReferencedProductsInAnyOrderIT {
                                         .errorCallback((syncException, draft, product, updateActions)
                                             -> collectErrors(syncException.getMessage(), syncException))
                                         .beforeUpdateCallback(this::collectActions)
+                                        .batchSize(1)
                                         .warningCallback(warningCallback)
                                         .build();
     }
--- a/src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
+++ b/src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
@@ -26,7 +26,6 @@ import io.sphere.sdk.taxcategories.TaxCategory;
 
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -386,9 +385,7 @@ public final class ProductReferenceResolver extends BaseReferenceResolver<Produc
         final List<CompletionStage<Map<String, String>>> futures = new ArrayList<>();
 
         final Set<String> productKeys = referencedKeys.getProductKeys();
-        if (!productKeys.isEmpty()) {
-            futures.add(productService.cacheKeysToIds(productKeys));
-        }
+        futures.add(productService.cacheKeysToIds(productKeys));
 
         final Set<String> productTypeKeys = referencedKeys.getProductTypeKeys();
         if (!productTypeKeys.isEmpty()) {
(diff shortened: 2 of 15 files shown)


src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
```diff
@@ -386,7 +386,7 @@ public ProductReferenceResolver(@Nonnull final ProductSyncOptions productSyncOpt
         final List<CompletionStage<Map<String, String>>> futures = new ArrayList<>();
 
         final Set<String> productKeys = referencedKeys.getProductKeys();
-        if (!productKeys.isEmpty()) {
+        if (productKeys.isEmpty()) {
```
JudeNiroshan, line 389: why execute the reference fetch when we do not have any keys?

ahmetoz: no reason, it's typo, reverted: https://github.com/commercetools/commercetools-sync-java/pull/581/commits/5c22820a06e5c339a127aa7a6a5b6d8f30ee7f60

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
```diff
@@ -432,6 +432,6 @@ public ProductReferenceResolver(@Nonnull final ProductSyncOptions productSyncOpt
         }
 
         return collectionOfFuturesToFutureOfCollection(futures, toList())
-            .thenApply(maps -> productKeys.isEmpty() ? Collections.emptyMap() : maps.get(0));
+            .thenCompose(ignored -> productService.cacheKeysToIds(Collections.emptySet())); // return all cache.
```
JudeNiroshan, line 435: I guess it's the time to introduce a getter for the caches in the service layer. wdyt?

JudeNiroshan, line 435: also naming is confusing. `productService.cacheKeysToIds` when the user ask nothing, it shouldn't return everything it has.`getKeysToIdsCache` would be good candidate name for the new method

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncWithReferencedProductsInAnyOrderIT.java
```diff
@@ -112,6 +112,7 @@ private ProductSyncOptions buildSyncOptions() {
                                         .errorCallback((syncException, draft, product, updateActions)
                                             -> collectErrors(syncException.getMessage(), syncException))
                                         .beforeUpdateCallback(this::collectActions)
+                                        .batchSize(1)
```
JudeNiroshan, line 115: why we need to make batch size to 1? how does it relate to productWith500ProductReferences test case?

JudeNiroshan (changes requested):

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncWithReferencedProductsInAnyOrderIT.java
ahmetoz: caches are based on the batch, so it's ensuring the keyToId is considering other batches too, before it was not.

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
ahmetoz: For now, the cache is an internal part of the service, for this PR, I will focus the part of the quickly fixing the issues with returning cache. We will improve the caching with different PR soon.

ahmetoz: Same here, maybe I could add those points to a separate GitHub issue, soon we will describe the new concept for the cache implementation,  https://github.com/commercetools/commercetools-sync-java/pull/581#discussion_r496616879

JudeNiroshan (approved):

lojzatran (approved):

src/main/java/com/commercetools/sync/products/helpers/ProductReferenceResolver.java
ahmetoz: Added as a context: https://github.com/commercetools/commercetools-sync-java/issues/582

ahmetoz: Added as a context: https://github.com/commercetools/commercetools-sync-java/issues/582

Issue #579, opened 2020-09-28
Support CustomerSync.


ahmetoz:
### Description
The library should support the syncing of `Customer` resources.
 
1. Sync a list of `customerDraft` elements into a target commercetools project.
1. Use utils to compare differences in field values of `customerDraft` to be able to decide the custom object needs to be updated. 
   1. Matching should be based on key.
   1. Addresses should be matched by key 
   1. Import of anonymousCartId and anonymous id should be ignored
   1. Customer password creation or update should be skipped
1. CustomerReferenceResolver: Both customer group and custom type will be provided as ResourceIdentifier with the key set
1. CustomerBatchValidator validates the batch of drafts.
1. CustomerReferenceResolution should map from Customer to CustomerDraft.

ahmetoz:
Release with https://github.com/commercetools/commercetools-sync-java/blob/master/docs/RELEASE_NOTES.md#230---oct-15-2020

PR #585, opened 2020-09-29
Add sync statistics and sync options unit tests

salander85:
#### Summary
In order to configure customer sync behaviour and access statistics some helpers are required.

Related task: #579 

#### Description
Add **CustomerSyncOptionsBuilderTest** .
Add **CustomerSyncStatistics** including UnitTests

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [x] Documentation

### DIFF
--- a/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
+++ b/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
@@ -2,8 +2,6 @@ package com.commercetools.sync.customers.helpers;
 
 import com.commercetools.sync.commons.helpers.BaseSyncStatistics;
 
-import static java.lang.String.format;
-
 public class CustomerSyncStatistics extends BaseSyncStatistics {
 
     public CustomerSyncStatistics() {
@@ -19,9 +17,6 @@ public class CustomerSyncStatistics extends BaseSyncStatistics {
      */
     @Override
     public String getReportMessage() {
-        reportMessage = format(
-            "Summary: %s customers were processed in total (%s created, %s updated and %s failed to sync).",
-            getProcessed(), getCreated(), getUpdated(), getFailed());
-        return reportMessage;
+        return getDefaultReportMessageForResource("customers");
     }
 }
--- a/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
+++ b/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
@@ -1,0 +1,27 @@
+package com.commercetools.sync.customers.helpers;
+
+import com.commercetools.sync.commons.helpers.BaseSyncStatistics;
+
+import static java.lang.String.format;
+
+public class CustomerSyncStatistics extends BaseSyncStatistics {
+
+    public CustomerSyncStatistics() {
+        super();
+    }
+
+    /**
+     * Builds a summary of the customer sync statistics instance that looks like the following example:
+     *
+     * <p>"Summary: 2 customers have been processed in total (0 created, 0 updated and 0 failed to sync)."
+     *
+     * @return a summary message of the customer sync statistics instance.
+     */
+    @Override
+    public String getReportMessage() {
+        reportMessage = format(
+            "Summary: %s customers have been processed in total (%s created, %s updated and %s failed to sync).",
+            getProcessed(), getCreated(), getUpdated(), getFailed());
+        return reportMessage;
+    }
+}
--- a/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
+++ b/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
@@ -6,6 +6,10 @@ import static java.lang.String.format;
 
 public class CustomerSyncStatistics extends BaseSyncStatistics {
 
+    public CustomerSyncStatistics() {
+        super();
+    }
+
     /**
      * Builds a summary of the customer sync statistics instance that looks like the following example:
      *
--- a/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
+++ b/src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
@@ -20,7 +20,7 @@ public class CustomerSyncStatistics extends BaseSyncStatistics {
     @Override
     public String getReportMessage() {
         reportMessage = format(
-            "Summary: %s customers have been processed in total (%s created, %s updated and %s failed to sync).",
+            "Summary: %s customers were processed in total (%s created, %s updated and %s failed to sync).",
             getProcessed(), getCreated(), getUpdated(), getFailed());
         return reportMessage;
     }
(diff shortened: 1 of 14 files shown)


src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
```diff
@@ -0,0 +1,27 @@
+package com.commercetools.sync.customers.helpers;
+
+import com.commercetools.sync.commons.helpers.BaseSyncStatistics;
+
+import static java.lang.String.format;
+
+public class CustomerSyncStatistics extends BaseSyncStatistics {
+
+    public CustomerSyncStatistics() {
```
ahmetoz, line 9: IMO, this constructor is not necessary as it's the same with java's implicit constructor.

lojzatran:
@salander85 CustomerSyncOptionsBuilder is not in this PR, just the test.

src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Builds a summary of the customer sync statistics instance that looks like the following example:
+     *
+     * <p>"Summary: 2 customers have been processed in total (0 created, 0 updated and 0 failed to sync)."
+     *
+     * @return a summary message of the customer sync statistics instance.
+     */
+    @Override
+    public String getReportMessage() {
```
lojzatran, line 21: I see this method is the same in many other sync statistics classes. What about putting it in the parent class?

salander85:
@lojzatran You're right, was already implemented. Changed the description.

src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
```diff
@@ -0,0 +1,27 @@
...
+        reportMessage = format(
+            "Summary: %s customers have been processed in total (%s created, %s updated and %s failed to sync).",
```
ahmetoz, line 23: The same meaning, here to make it consistent with other sync implementations.

```suggestion
            "Summary: %s customers were processed in total (%s created, %s updated and %s failed to sync).",
```

lojzatran (approved):
LGTM 🎉

src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
ahmetoz: Every sync instance could have a different message specific to the implementation.

lojzatran: I checked and most of them don't have different implementation. It's 4 out of 10 sync instances that have different implementations :) So 6 of them have the same.

src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
salander85: a1a146907d929e9639f580ebe47f894e7cdce157

src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
ahmetoz: I don't know where are you seeing value for improvement maybe explain if you see a different case than me. for me, this method just returns a string that is valuable inside its sub classes, maybe you could pass different parts with the constructor but it will look weird  😄

ahmetoz (approved):

src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
lojzatran: So it's like this now. 6 children have this method implemented

```
@Override
    public String getReportMessage() {
        reportMessage = format(
            "Summary: %s types were processed in total (%s created, %s updated and %s failed to sync).",
            getProcessed(), getCreated(), getUpdated(), getFailed());

        return reportMessage;
    }
```

The only difference is the resource name in the message, in this case it is `types`. This could be implemented in the child, but other than that, we can have a common method for all of those 6 children (and more children to come). So then you don't have issues like https://github.com/commercetools/commercetools-sync-java/pull/585#discussion_r497335686.

Also `reportMessage` is a field, which I have no idea why? So I'm for the change to make a mother implementation in `BaseSyncStatistics` and also to remove `reportMessage` field.

salander85: c61198d

src/main/java/com/commercetools/sync/customers/helpers/CustomerSyncStatistics.java
```diff
@@ -6,6 +6,10 @@
 
 public class CustomerSyncStatistics extends BaseSyncStatistics {
 
+    public CustomerSyncStatistics() {
```
ahmetoz, line 7: Why you reverted it ?

salander85: IT were broken.

ahmetoz: Looks like a flaky situation related to API consistency, nothing related to the constructor of customer object:

````
com.commercetools.sync.integration.externalsource.producttypes.ProductTypeWithNestedAttributeSyncIT.setup(ProductTypeWithNestedAttributeSyncIT.java:74)
      Caused by: io.sphere.sdk.client.ErrorResponseException: detailMessage: A duplicate value '"key_3"' exists for field 'key'.
      summary: POST https://api.europe-west1.gcp.commercetools.com/java-sync-target/product-types failed  with response code 400 with X-Correlation-ID `java-sync-target/bfd545fe-c93f-4596-b045-4aebbaa83cf8` on 2020-09-30T09:02:21.653Z
        "statusCode" : 400,
        "message" : "A duplicate value '"key_3"' exists for field 'key'.",
        "errors" : [ {
          "code" : "DuplicateField",
          "message" : "A duplicate value '"key_3"' exists for field 'key'.",
          "duplicateValue" : "key_3",
          "field" : "key"
        } ]
      }
      http response: io.sphere.sdk.http.HttpResponseImpl@630361b7[statusCode=400,headers={date=[Wed, 30 Sep 2020 09:02:21 GMT], content-length=[226], server=[istio-envoy], x-envoy-upstream-service-time=[10], Alt-Svc=[clear], access-control-allow-headers=[Accept, Authorization, Content-Type, Origin, User-Agent, X-Correlation-ID], x-correlation-id=[java-sync-target/bfd545fe-c93f-4596-b045-4aebbaa83cf8], access-control-allow-methods=[GET, POST, DELETE, OPTIONS], Via=[1.1 google], access-control-allow-origin=[*], access-control-max-age=[299], content-type=[application/json; charset=utf-8], server-timing=[projects;dur=9]},associatedRequest=<null>,textInterpretedBody={"statusCode":400,"message":"A duplicate value '"key_3"' exists for field 'key'.","errors":[{"code":"DuplicateField","message":"A duplicate value '"key_3"' exists for field 'key'.","duplicateValue":"key_3","field":"key"}]}]
      SDK: 1.53.0
      project: java-sync-target
      endpoint: POST /product-types
      Java: 1.8.0_252
````

salander85: 86b533e

PR #587, opened 2020-10-01
Add helpers of CustomerSync: CustomerBatchValidator, CustomerReferenceResolver, CustomerReferenceResolverUtils.

ahmetoz:
#### Summary
The PR contains the helpers/utils that will be used by CustomerSync class.

#### Relevant Issue
#579 

#### Todo:
- [x] Full Coverage 
- [x] Check store reference case: when it's provided as id but no key, in this case, Customer#getStore will not return a resource identifier with id, so we need to force the library used to provide a key-value otherwise an update comparison can not be done.  **Update: instead it will be added as a new store.**
- [x] Add validation for the password if it's a blank return draft with error callback.

#### Hints for Review

- Customer addresses are not unique by key in the commercetools API, so we are doing validations for the addresses to avoid edge cases in syncing addresses. 

- CustomerDraft is storing indexes as integers, but the Customer type is storing addresses (like billing and shippingAddresses) by its ids. So in CustomerReferenceResolverUtils, the mapping is handled for the update cases.

- Customer includes customer group, custom type, and stores as references, so it needs to be resolved before calculating the update actions, the helper for this is CustomerReferenceResolver.

- Store is returning as a key reference so the reference resolver is not resolving the id and directly returning the resource identifier.


lojzatran:
@ahmetoz can you stick to the contributing rules next time? https://github.com/commercetools/commercetools-sync-java/blob/master/docs/CONTRIBUTING.md#features-or-bug-fixes

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+        @Nonnull final CustomerBatchValidator.ReferencedKeys referencedKeys,
+        @Nonnull final CustomerDraft customerDraft) {
+
+        collectReferencedKeyFromResourceIdentifier(customerDraft.getCustomerGroup(),
+            referencedKeys.customerGroupKeys::add);
+        collectReferencedKeyFromCustomFieldsDraft(customerDraft.getCustom(),
+            referencedKeys.typeKeys::add);
+    }
+
+    public static class ReferencedKeys {
```
leungkinghin, line 211: It seems not necessary to make it `static`

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+    @Nonnull
+    private List<Integer> getIndexes(@Nonnull final List<Address> list, @Nonnull final Predicate<Address> predicate) {
+
+        final List<Integer> indexes = new ArrayList<>();
+        for (int i = 0; i < list.size(); i++) {
+            if (predicate.test(list.get(i))) {
+                indexes.add(i);
+            }
+        }
+        return indexes;
```
leungkinghin, line 146:
```suggestion
return list.stream().filter(predicate).collect(Collectors.toList());
```

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+            handleError(format(ADDRESSES_THAT_KEYS_NOT_UNIQUE, customerDraft.getKey(), indexesOfDuplicateAddressKeys));
+            return false;
+        }
+
+        return true;
+    }
+
+    @Nonnull
+    private List<Integer> getIndexes(@Nonnull final List<Address> list, @Nonnull final Predicate<Address> predicate) {
+
```
leungkinghin, line 139:
```suggestion
     return list.stream().filter(predicate).collect(Collectors.toList());
```

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
leungkinghin, line 146:
```suggestion
        return list.stream().filter(predicate).collect(Collectors.toList());
```

src/test/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolverTest.java
```diff
@@ -0,0 +1,352 @@
...
+
+        when(customerGroupService.fetchCachedCustomerGroupId(anyString()))
+            .thenReturn(completedFuture(Optional.empty()));
+
+        final String expectedMessageWithCause = format(FAILED_TO_RESOLVE_CUSTOMER_GROUP_REFERENCE,
+            "dummyKey", format(CUSTOMER_GROUP_DOES_NOT_EXIST, "anyKey"));
+
+        referenceResolver.resolveCustomerGroupReference(customerDraftBuilder)
+                         .exceptionally(exception -> {
+                             assertThat(exception).hasCauseExactlyInstanceOf(ReferenceResolutionException.class);
```
lojzatran, line 187: If the test doesn't go to `.exceptionally` block, this test will also pass?

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+    private List<Integer> getIndexValues(@Nonnull final List<Integer> list,
+        @Nonnull final Predicate<Integer> predicate) {
+
+        final List<Integer> indexes = new ArrayList<>();
+        for (Integer integer : list) {
+            if (predicate.test(integer)) {
+                indexes.add(integer);
+            }
+        }
+        return indexes;
```
leungkinghin, line 198:
```suggestion
return list.stream().filter(predicate).collect(Collectors.toList());
```

src/test/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolverTest.java
```diff
@@ -0,0 +1,352 @@
...
+    @Test
+    void resolveStoreReferences_WithValidStores_ShouldResolveReferences() {
+        final CustomerDraftBuilder customerDraftBuilder = CustomerDraftBuilder
+            .of("email@example.com", "secret123")
+            .stores(asList(ResourceIdentifier.ofKey("store-key1"),
+                ResourceIdentifier.ofKey("store-key2"),
+                ResourceIdentifier.ofId("store-id-3")))
+            .key("dummyKey");
+
+        // test
```
lojzatran, line 295: I see sometimes you have `//test` and `// assertion`, sometimes you don't. I would have 1 style - either have it everywhere or nowhere. In this file, the tests are pretty easy, so it's not even necessary.

lojzatran (changes requested):

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Predicate;
+
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toSet;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+public class CustomerBatchValidator
+    extends BaseBatchValidator<CustomerDraft, CustomerSyncOptions, BaseSyncStatistics> {
```
leungkinghin, line 24: Or we should use `CustomerSyncStatistics` here? since we don't expect to parse `SyncStatistics` class of other resources as argument.

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+    static final String BILLING_ADDRESSES_ARE_NOT_VALID = "CustomerDraft with key: '%s' does not contain indices: '%s' "
+        + "of the 'billingAddresses' in the addresses list. "
+        + "Please make sure all customer drafts have valid index values for the 'billingAddresses'.";
+
+    static final String SHIPPING_ADDRESSES_ARE_NOT_VALID = "CustomerDraft with key: '%s' does not contain indices: '%s'"
+        + " of the 'shippingAddresses' in the addresses list. "
+        + "Please make sure all customer drafts have valid index values for the 'shippingAddresses'.";
+
+    public CustomerBatchValidator(@Nonnull final CustomerSyncOptions syncOptions,
+                                  @Nonnull final BaseSyncStatistics syncStatistics) {
```
leungkinghin, line 48: Same as class definition above

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: @leungkinghin do you think one-liner will do the same? because method returns integer values (indexes), not addresses?

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: same here, don't think your one-liner will do the same.

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
+package com.commercetools.sync.customers.helpers;
+
+import com.commercetools.sync.commons.helpers.BaseBatchValidator;
+import com.commercetools.sync.commons.helpers.BaseSyncStatistics;
+import com.commercetools.sync.customers.CustomerSyncOptions;
+import io.sphere.sdk.customers.CustomerDraft;
+import io.sphere.sdk.models.Address;
+import org.apache.commons.lang3.tuple.ImmutablePair;
```
leungkinghin, line 8: Better to follow alphabetical order for package import?

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+     * <li>It has a email which is not blank (null/empty)</li>
+     * <li>It has not null addresses</li>
+     * <li>It has valid addresses with keys is not blank (null/empty)</li>
+     * <li>It has unique address keys</li>
+     * <li>It has valid 'billingAddresses' and 'shippingAddresses'</li>
+     * </ol>
+     *
+     * <p>Note that {@link io.sphere.sdk.customers.CustomerDraftDsl} is validating the 'defaultBillingAddress' and
+     * 'defaultShippingAddress' with '#isValidAddressIndex' method in it's constructor, so default values are not
+     * possible to be invalid.
```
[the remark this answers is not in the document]
ahmetoz: Yes, is 'invalid' word a negation? 

I will try to rephrase.

src/test/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolverTest.java
ahmetoz: Need to check, probably a hasFailed check is required.

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: IMO in java inner classes are better with static vs non-static, otherwise every instance will create a different reference of it, also static inner class can not access it's enclosing scope, which is better in terms of secure, I guess there was a section about this in effective java book, maybe you might check and let me know if I am wrong.

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: Good catch, I was developing that before the CustomerSyncStatistic object is added, and then forgot to change it.

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: I did not get this comment, what do you mean ?

ahmetoz:
> @ahmetoz can you stick to the contributing rules next time? https://github.com/commercetools/commercetools-sync-java/blob/master/docs/CONTRIBUTING.md#features-or-bug-fixes

Why sticking something unnecessary for now? Maybe some rules in here need to be updated/changed/removed.

IMO a contribution is valuable without git conventions, putting the commit id in the commit is not required, as we need to consider using squash (as I push the team for it) with cleaning all unnecessary commit history and commit messages, because in the issue maybe you have 100 commits to address comments separately and you are losing focus when you check the issue in GitHub page. 

Importing thing is referencing a PR with a GitHub issue, for this PR before I've opened a GitHub issue and linked with labels, which most of the time team does not do. 

In general, this is a PR for the chucks of work, and we will commit all together in master after we combine them in one PR.
For the branch name, we could add the issue number but this is not a must and branch name is just a reference because we are removing branches when we merge them, so searching is not an issue.

If you think I am missing something please let me know specifically.

src/main/java/com/commercetools/sync/customers/utils/CustomerReferenceResolutionUtils.java
```diff
@@ -0,0 +1,180 @@
...
+        @Nullable final String addressId) {
+
+        if (allAddresses == null) {
+            return null;
+        }
+        if (isBlank(addressId)) {
+            return null;
+        }
+        for (int i = 0; i < allAddresses.size(); i++) {
+            String id = allAddresses.get(i).getId();
```
leungkinghin, line 123: Address element may have potential to be null. I know we did the check in validator. As a utility class we cannot assume it has been validated before it is manipulated. wdyt?

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
leungkinghin: I get your point now. It's ok :)

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Predicate;
```
leungkinghin, line 17: Better to follow alphabetical order for package import for easy search
```suggestion
import javax.annotation.Nonnull;
import javax.annotation.Nullable;
import java.util.ArrayList;
import java.util.HashSet;
import java.util.List;
import java.util.Objects;
import java.util.Set;
import java.util.function.Predicate;
import org.apache.commons.lang3.tuple.ImmutablePair;
```

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
leungkinghin: I mean placing the org.apache.* after java.* for easier search

src/main/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolver.java
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Given a {@link CustomerDraftBuilder} this method attempts to resolve the stores and return
+     * a {@link CompletionStage} which contains a new instance of the builder with the resolved references.
+     *
+     * @param draftBuilder the customer draft to resolve its store references.
+     * @return a {@link CompletionStage} that contains as a result a new builder instance with resolved references or,
+     *         in case an error occurs during reference resolution, a {@link ReferenceResolutionException}.
+     */
+    @Nonnull
+    public CompletionStage<CustomerDraftBuilder> resolveStoreReferences(
```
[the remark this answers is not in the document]
ahmetoz: I don't want to handle the checked exception here, so the stage is not complete for the exceptional case.

src/main/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolver.java
```diff
@@ -0,0 +1,167 @@
...
+        @Nonnull final CustomerDraftBuilder draftBuilder) {
+
+        final List<ResourceIdentifier<Store>> storeResourceIdentifiers = draftBuilder.getStores();
+        if (storeResourceIdentifiers == null || storeResourceIdentifiers.isEmpty()) {
+            return completedFuture(draftBuilder);
+        }
+        final List<ResourceIdentifier<Store>> resolvedReferences = new ArrayList<>();
+        for (ResourceIdentifier<Store> storeResourceIdentifier : storeResourceIdentifiers) {
+            if (storeResourceIdentifier != null) {
+                if (storeResourceIdentifier.getId() == null) {
```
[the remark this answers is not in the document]
ahmetoz: The draft could have a resource identifier with id set. 

Resolving the id means fetching the id from the target project and putting into a reference, here API will return with the key (normally a reference is not returning it's key if you don't expand) so update action could check the key.

src/test/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolverTest.java
ahmetoz: 328f92e

src/main/java/com/commercetools/sync/customers/utils/CustomerReferenceResolutionUtils.java
```diff
@@ -0,0 +1,180 @@
...
+import static org.apache.http.util.TextUtils.isBlank;
+
+/**
+ * Util class which provides utilities that can be used when syncing resources from a source commercetools project
+ * to a target one.
+ */
+public final class CustomerReferenceResolutionUtils {
+
+    /**
+     * Returns an {@link List}&lt;{@link CustomerDraft}&gt; consisting of the results of applying the
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/587/commits/1b1a917489f3586eda81452a92cd6596abdffb1c

src/test/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolverTest.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/587/commits/ebf3d30834d883ab6fb714a8bcc2756d65b19ff3

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/587/commits/3a05f4366b7bfec8d95b5737d8acd8f4b4f2701d

src/main/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolver.java
```diff
@@ -0,0 +1,167 @@
...
+    public static final String FAILED_TO_RESOLVE_STORE_REFERENCE =
+        "Failed to resolve store resource identifier on CustomerDraft with key:'%s'. Reason: %s";
+    public static final String FAILED_TO_RESOLVE_CUSTOM_TYPE = "Failed to resolve custom type reference on "
+        + "CustomerDraft with key:'%s'.";
+    public static final String CUSTOMER_GROUP_DOES_NOT_EXIST = "CustomerGroup with key '%s' doesn't exist.";
+
+    private final CustomerGroupService customerGroupService;
+
+    /**
+     * Takes a {@link CustomerSyncOptions} instance, a  {@link TypeService} to instantiate a
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/587/commits/379e24cf84de70a9c370fef64717719fa1ef9f43

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
```diff
@@ -0,0 +1,223 @@
...
+     * drafts and return an {@link ImmutablePair}&lt;{@link Set}&lt;{@link CustomerDraft}&gt;,
+     * {@link CustomerBatchValidator.ReferencedKeys}&gt; which contains the {@link Set} of valid drafts and
+     * referenced keys within a wrapper.
+     *
+     * <p>A valid customer draft is one which satisfies the following conditions:
+     * <ol>
+     * <li>It is not null</li>
+     * <li>It has a key which is not blank (null/empty)</li>
+     * <li>It has a email which is not blank (null/empty)</li>
+     * <li>It has not null addresses</li>
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/587/commits/125078eef75e61887dc99dee7f5a28fd8b195bfb

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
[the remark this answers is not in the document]
ahmetoz: I removed this:

https://github.com/commercetools/commercetools-sync-java/pull/587/commits/88b66a385348e87d0966e07adcdba11d75707d2e

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/587/commits/721edc11b784fe4db9efaa484405ce15b0115bf5

src/main/java/com/commercetools/sync/customers/helpers/CustomerBatchValidator.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/587/commits/721edc11b784fe4db9efaa484405ce15b0115bf5

src/main/java/com/commercetools/sync/customers/helpers/CustomerReferenceResolver.java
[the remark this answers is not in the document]
lojzatran: I see, but if you throw here instead of doing `exceptionallyCompletedFuture`, will it be caught anyway in a `.exceptionally` clause or you need to explicitly do try/catch? I'm not sure how it works in Java, but I thought it's the same way as in JS - if you have a an exception thrown in a Promise, it will be caught also in a Promise. 

```
    public CustomerDraftBuilder resolveStoreReferences(
        @Nonnull final CustomerDraftBuilder draftBuilder) throws ReferenceResolutionException {

        final List<ResourceIdentifier<Store>> storeResourceIdentifiers = draftBuilder.getStores();
        if (storeResourceIdentifiers == null || storeResourceIdentifiers.isEmpty()) {
            return draftBuilder;
        }
        final List<ResourceIdentifier<Store>> resolvedReferences = new ArrayList<>();
        for (ResourceIdentifier<Store> storeResourceIdentifier : storeResourceIdentifiers) {
            if (storeResourceIdentifier != null) {
                if (storeResourceIdentifier.getId() == null) {
                    try {
                        final String storeKey = getKeyFromResourceIdentifier(storeResourceIdentifier);
                        resolvedReferences.add(ResourceIdentifier.ofKey(storeKey));
                    } catch (ReferenceResolutionException referenceResolutionException) {
                            throw new ReferenceResolutionException(
                                format(FAILED_TO_RESOLVE_STORE_REFERENCE,
                                    draftBuilder.getKey(), referenceResolutionException.getMessage()));
                    }
                } else {
                    resolvedReferences.add(ResourceIdentifier.ofId(storeResourceIdentifier.getId()));
                }
            }
        }
        return draftBuilder.stores(resolvedReferences);
    }
```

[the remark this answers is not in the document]
ahmetoz: You are not wrong with saying if it's not an async operation why you are using it, here it's my preference to make less effort, if I throw a checked exception here other methods need to implement it, which will go till the sync implementation so here exception handling is using the futures, If it's ok for you, I want to keep it as is.

Issue #271, opened 2018-03-08
Always return more specific CompletebleFuture not generic CompletetionStage at method returns


heshamMassoud:
### Description
Always return more specific CompletebleFuture not generic CompletetionStage at method returns

Check more: https://github.com/commercetools/commercetools-sync-java/pull/269#discussion_r173141642

ahmetoz:
IMO, the past title here is not ***always*** correct, not seeing this as an issue.

```
 i think we did design mistake from the beginning: it is generally recommended to return as narrow (low) class as possible, and for most of our functions we have to return CompletableFuture!
```

No. it's not generally recommended, for example, we are not returning an ArrayList but returning List instead, also when you have an interface you are returning interface instead of implementation of your service. 

The CompletionStage is meant to use in an async manner, to completable future is only needed when someone is trying to use the sync in a blocker way. 

So it depends on how you want to use it, if you have an internal method to be used with get or join maybe it could return in the future to avoid calling the .toCompletableFuture().

PR #588, opened 2020-10-04
#271 Return CompletableFuture instead of CompletionStage

b0n541:
#### Summary
Migration of all methods that returned CompletionStage to return CompletableFuture

#### Description
I did change all occurances of CompleationStage to CompletableFuture and adjusted the tests where needed.

#### Relevant Issues
#271

#### Todo

- Tests
    - [ ] Unit 
    - [ ] Integration
- [ ] Documentation

#### Hints for Review
Too many files that were changes. Can't really give a hint here.


ahmetoz:
Thanks for the contribution but we've decided to not implement this change because it is not always right o use CompletableFuture instead of CompletionStage.

PR #590, opened 2020-10-05
Add CustomerSync class implementation

ahmetoz:
The implementation part of the CustomerSync, PR contains only the unit tests.
For other sync implementations we were writing unnecessary unit test scenarios into integration tests, this PR also covers the test scenarios like concurrent modification failures and other exceptional case scenarios. 

For example https://github.com/commercetools/commercetools-sync-java/blob/impl-sync-class/src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java#L443-L476

Relevant issue: #579


src/main/java/com/commercetools/sync/customers/CustomerSync.java
```diff
@@ -0,0 +1,235 @@
...
+                        statistics.incrementCreated();
+                    } else {
+                        statistics.incrementFailed();
+                    }
+                }))
+            .orElseGet(() -> CompletableFuture.completedFuture(null));
+    }
+
+    /**
+     * Given a {@link String} {@code errorMessage} and a {@link Throwable} {@code exception}, this method calls the
```
lojzatran, line 223: I'm not sure if this comment is valid. Can you check?

lojzatran (commented):
Are integration tests also part of this PR?

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
```diff
@@ -0,0 +1,322 @@
...
+import io.sphere.sdk.models.SphereException;
+import io.sphere.sdk.types.CustomFieldsDraft;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletionException;
```
leungkinghin, line 25:
```suggestion
import java.util.ArrayList;
import java.util.HashSet;
import java.util.List;
import java.util.Optional;
import java.util.concurrent.CompletionException;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Test;
```

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
```diff
@@ -0,0 +1,322 @@
...
+public class CustomerSyncTest {
+
+    private CustomerSyncOptions syncOptions;
+    private List<String> errorMessages;
+    private List<Throwable> exceptions;
+
+    @BeforeEach
+    void setup() {
+        errorMessages = new ArrayList<>();
+        exceptions = new ArrayList<>();
```
leungkinghin, line 62: Looks better if we can avoid creating new `List` object before every test case.
```suggestion
    private List<String> errorMessages = new ArrayList<>();
    private List<Throwable> exceptions = new ArrayList<>();

    @BeforeEach
    void setup() {
        errorMessages.clear();
        exceptions.clear();
```

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
```diff
@@ -0,0 +1,322 @@
...
+        final SphereClient ctpClient = mock(SphereClient.class);
+
+        syncOptions = CustomerSyncOptionsBuilder
+            .of(ctpClient)
+            .errorCallback((exception, oldResource, newResource, updateActions) -> {
+                errorMessages.add(exception.getMessage());
+                exceptions.add(exception);
+            })
+            .build();
```
leungkinghin, line 71: Can we initialize the `syncOptions` object globally rather than initialize it again and again in `setup()?`

leungkinghin, line 71: Can we initialize the `syncOptions` object globally rather than initialize it again and again in `setup()`?

src/main/java/com/commercetools/sync/customers/CustomerSync.java
```diff
@@ -0,0 +1,235 @@
...
+        final Set<CustomerDraft> validCustomerDrafts = result.getLeft();
+        if (validCustomerDrafts.isEmpty()) {
+            statistics.incrementProcessed(batch.size());
+            return completedFuture(statistics);
+        }
+
+        return referenceResolver
+            .populateKeyToIdCachesForReferencedKeys(result.getRight())
+            .handle(ImmutablePair::new)
+            .thenCompose(cachingResponse -> {
```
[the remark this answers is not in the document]
ahmetoz: Looks like you are right because it's not needed in here, most structure of this code coming from the existing code base, for other implementations maybe we could check and refactor all similar unnecessary mappings in a different PR.

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
ahmetoz: not a big improvement for a test scenario but doable.

src/main/java/com/commercetools/sync/customers/CustomerSync.java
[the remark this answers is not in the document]
ahmetoz: it is not easy to return stages from the handle, if I change it, I need to change in many places, this practice does not need an update for now.

src/main/java/com/commercetools/sync/customers/CustomerSync.java
ahmetoz: I removed some javadocs in here, for all private methods: https://github.com/commercetools/commercetools-sync-java/pull/590/commits/a94cacc484b1e7578de4d86291f6b44723e39876

src/test/java/com/commercetools/sync/commons/asserts/statistics/AssertionsForStatistics.java
```diff
@@ -115,4 +116,15 @@ public static TaxCategorySyncStatisticsAssert assertThat(@Nullable final TaxCate
     public static CustomObjectSyncStatisticsAssert assertThat(@Nullable final CustomObjectSyncStatistics statistics) {
         return new CustomObjectSyncStatisticsAssert(statistics);
     }
+
+    /**
+     * Create assertion for {@link CustomerSyncStatistics}.
+     *
+     * @param statistics the actual value.
```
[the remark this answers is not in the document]
ahmetoz: not seeing any improvement.

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
ahmetoz: will not change this, as the import is an automated process by the intellij idea.

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
ahmetoz: will not change this, in the end, the same thing, it's not production code so optimization not needed.

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
ahmetoz: I would rather use JUnit hooks to create fixtures. So I would keep it as is.

src/test/java/com/commercetools/sync/commons/asserts/statistics/AssertionsForStatistics.java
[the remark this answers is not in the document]
lojzatran: Just the grammar :)

lojzatran (approved):

src/test/java/com/commercetools/sync/customers/CustomerSyncTest.java
leungkinghin: It depends. Intellij sometimes generates wildcard import for developers. I would rather fix it in that case. Here it is about import ordering. it is just "nice-to-have" and still acceptable if we keep it unchanged.

leungkinghin (approved):

leungkinghin:
Nice. We begin not to mock ctp client in SyncTest :)

PR #601, opened 2020-10-15
Handles edge cases for the update actions: `setStores` and `setCustomerNumber`.

ahmetoz:
## Solutions:
1- We will return a warning message with a callback. For example: `"Customer with key: "key1" has already a customer number: "1234", once it's set it cannot be changed. Hereby, the update action is not created."`

2- Instead of using `addStore` and `removeStore` actions together for this we will use `setStores`.

# Problems:
It is possible that the library will create some update actions that are not valid or un-necessary. We figured out those actions while working with integration tests.

1 - Customer number uniqueness: https://docs.commercetools.com/api/projects/customers#set-customer-number

Currently, the library will create an update action for customer numbers and it will be failed by API because the customer number can not be changed after it set.

```
"statusCode" : 400,
"message" : "A Customer number already exists and cannot be set again.",
"errors" : [ {
"code" : "InvalidOperation",
"message" : "A Customer number already exists and cannot be set again.",
"action" : {
"action" : "setCustomerNumber",
"customerNumber" : "gold-2"
},
"actionIndex" : 7
} ]
```

2 - Email uniqueness: https://docs.commercetools.com/api/projects/customers#email-uniqueness

The case only happening when email changes and some existing stores are removed, 
````
{
"action" : "changeEmail",
"email" : "jane@example.com"
}
{
"action" : "removeStore",
"store" : {
"key" : "store-hamburg"
}
````

````
"statusCode" : 400,
"message" : "There is already an existing customer with the email '\"john@example.com\"'.",
"errors" : [ {
"code" : "DuplicateField",
"message" : "There is already an existing customer with the email '\"john@example.com\"'.",
"duplicateValue" : "john@example.com",
"field" : "email"
} ]
}
````

### DIFF
--- a/src/main/java/com/commercetools/sync/customers/utils/CustomerUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/customers/utils/CustomerUpdateActionUtils.java
@@ -52,6 +52,7 @@ import static com.commercetools.sync.commons.utils.CommonTypeUpdateActionUtils.b
 import static com.commercetools.sync.commons.utils.CommonTypeUpdateActionUtils.buildUpdateActionForReferences;
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
@@ -319,6 +320,7 @@ public final class CustomerUpdateActionUtils {
         }
 
         // TODO (JVM-SDK), see: SUPPORT-10261 SetCustomerGroup could be created with a ResourceIdentifier
+        // https://github.com/commercetools/commercetools-jvm-sdk/issues/2072
         return new ResourceImpl<CustomerGroup>(null, null, null, null) {
             @Override
             public Reference<CustomerGroup> toReference() {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/main/java/com/commercetools/sync/customers/utils/CustomerUpdateActionUtils.java [hunk 1]
```diff
@@ -349,17 +375,30 @@ private CustomerUpdateActionUtils() {
...
         return buildSetStoreUpdateAction(oldStores, newStores)
             .map(Collections::singletonList)
-            .orElseGet(() -> {
-                if (oldStores != null && newStores != null) {
-                    final List<UpdateAction<Customer>> updateActions =
-                        buildRemoveStoreUpdateActions(oldStores, newStores);
+            .orElseGet(() -> prepareStoreActions(oldStores, newStores));
+    }
+
+    private static List<UpdateAction<Customer>> prepareStoreActions(
```
lojzatran, line 381: Why do you have both setStores and add/remove stores? Isn't it enough to have setStores?

ahmetoz: It's enough, `setStores` can do add and remove together, but here I want to expose both and also assuming it's better for API's performance. I was thinking those are public methods but realized it's not :) https://github.com/commercetools/commercetools-sync-java/pull/601/commits/49848f378825d7331a484f5f96e7e753d38d0af2

lojzatran: Then I guess you can also update this part
```
        return buildSetStoreUpdateAction(oldStores, newStores)
            .map(Collections::singletonList)	
            .orElseGet(() -> prepareStoreActions(oldStores, newStores));
```

ahmetoz: Let me check what could be done in here.

lojzatran: I think you can remove this part:
```
            .orElseGet(() -> prepareStoreActions(oldStores, newStores));
```

ahmetoz: Anyways I will keep this as is, setStores will not have a separate public method, 
so `buildStoreUpdateActions` will prepare all actions, which is acceptable for the purpose of the `setStores`. 
`buildRemoveStoreUpdateActions` will prepare only  `removeStore` and same for 
`buildAddStoreUpdateActions` will prepare only `addStore` actions.

src/main/java/com/commercetools/sync/customers/utils/CustomerUpdateActionUtils.java
```diff
@@ -349,17 +375,30 @@ private CustomerUpdateActionUtils() {
 
         return buildSetStoreUpdateAction(oldStores, newStores)
             .map(Collections::singletonList)
-            .orElseGet(() -> {
-                if (oldStores != null && newStores != null) {
-                    final List<UpdateAction<Customer>> updateActions =
-                        buildRemoveStoreUpdateActions(oldStores, newStores);
+            .orElseGet(() -> prepareStoreActions(oldStores, newStores));
```
lojzatran, line 378: This part here is a bit weird. You already call `buildSetStoreUpdateAction` which should return you correct update actions to have a synced stores. And if this update action is empty, then I guess `prepareStoreActions` would also return empty as it generates `addStore` or `removeStore`.

src/main/java/com/commercetools/sync/customers/utils/CustomerUpdateActionUtils.java
lojzatran: I added a comment, maybe it's more clear now: https://github.com/commercetools/commercetools-sync-java/pull/601/files#r505397040

src/main/java/com/commercetools/sync/customers/utils/CustomerUpdateActionUtils.java
ahmetoz: The trick in the `buildSetStoreUpdateAction` private method, it's creating only `setStore` action when old customer does not have stores, acting like adding stores. The function flows a little bit weird but results are expected.

lojzatran (approved):

PR #603, opened 2020-10-16
Improve documentation for default client creation

ninalindl:
#### Summary
Users should see more easily how to create a sphere client.
#### Description

To ensure users know how to create the default sphere client with ClientConfigurationUtils, it was added to every sync user guide.
#### Todo

- Tests
    - [ ] Unit 
    - [ ] Integration
- [x] Documentation

### DIFF
--- a/docs/usage/CART_DISCOUNT_SYNC.md
+++ b/docs/usage/CART_DISCOUNT_SYNC.md
@@ -141,17 +141,17 @@ final CartDiscountSyncOptions cartDiscountSyncOptions =
 If you want to customize the sync process, consider the following:
 The sync library is not meant to be executed in a parallel fashion. For example:
 ````java
-final ProductSync productSync = new ProductSync(syncOptions);
-final CompletableFuture<ProductSyncStatistics> syncFuture1 = productSync.sync(batch1).toCompletableFuture();
-final CompletableFuture<ProductSyncStatistics> syncFuture2 = productSync.sync(batch2).toCompletableFuture();
+final CartDiscountSync cartDiscountSync = new CartDiscountSync(syncOptions);
+final CompletableFuture<CartDiscountSyncStatistics> syncFuture1 = cartDiscountSync.sync(batch1).toCompletableFuture();
+final CompletableFuture<CartDiscountSyncStatistics> syncFuture2 = cartDiscountSync.sync(batch2).toCompletableFuture();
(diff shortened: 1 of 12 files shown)


leungkinghin (changes requested):
The change looks good. Besides we need to include this section as well, since it is not linked anywhere in new version.
https://github.com/commercetools/commercetools-sync-java/blob/master/docs/usage/IMPORTANT_USAGE_TIPS.md#tuning-the-sync-process

I would suggest we can list this section as **Additional Info** after the `sphereClient` creation, so that users can follow it if they really want to do customisation.

docs/usage/CART_DISCOUNT_SYNC.md
```diff
@@ -25,11 +25,19 @@ against a [CartDiscountDraft](https://docs.commercetools.com/http-api-projects-c
...
 #### Prerequisites
-1. The sync expects a list of `CartDiscountDraft`s that have their `key` fields set to be matched with
+1. Create a `sphereClient`:
+Use the `ClientConfigurationUtils#createClient` util which applies the best practices for `SphereClient` creation.
+
+    If you have special requirements on the sphere client creation, then you can use the following:  
+    - Limit the number of concurrent requests done to CTP. This can be done by decorating the `sphereClient` with [QueueSphereClientDecorator](http://commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/client/QueueSphereClientDecorator.html)
+ 
+    - Retry decorator on 5xx errors with a retry strategy. This can be achieved by decorating the `sphereClient` with the [RetrySphereClientDecorator](http://commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/client/RetrySphereClientDecorator.html)
+
```
butenkor, line 35: Do you think it would be better to link to a separate page which describes both decorators and maybe more in the future instead of repeating it everywhere?

ninalindl: Until now this information was being linked. It was included in the "Important Usage Tips" page. The goal of this change is to improve the visibility of this information, that is why it is now included in every user guide. However, since I am not a user of the library I am not sure how users are using the docs and where they would search for this information.

leungkinghin (approved):

docs/usage/CART_DISCOUNT_SYNC.md
ninalindl: 2f5557f4

Issue #208, opened 2017-12-06
Change ctp client configuration


heshamMassoud:
These configuration on the ctp client should:
- [ ] timeout 30s -> 3 minutes
-> It's important to know, that the timeout value passed to the `SphereBlockingClient` here is only used when the method `executeBlocking()` is called. Therefore, its not actually used from within the sync. : 
https://github.com/commercetools/commercetools-sync-java/blob/482356facd204b6261120b5d022ee6ef4a8a5402/src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java#L45-L57

Therefore, to change the timeout value, this has to be changed on the default values of the underlying HTTP client here:
https://github.com/commercetools/commercetools-sync-java/blob/482356facd204b6261120b5d022ee6ef4a8a5402/src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java#L76-L86
using the following setters:
````java
                        .setRequestTimeout(30000)
                        .setConnectTimeout(1000)
                        .setReadTimeout(100)
````
The meaning and significance of each, need to be researched though.

- [ ] retryInterval start value 30s -> 200ms
- [ ] max retry count 50 -> 100
- [ ] The retry should also retry on all 5xx errors (including 500): 

https://github.com/commercetools/commercetools-sync-java/blob/b61de4b097fa87cf487f17cecf4260f8045fcc81/src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java#L92-L93

#### Update (14/02/2018) [Breaking change]
- Consider enforcing client creation in the sync library, rather than providing the option to inject the client to the library. 
##### Advantages
 1. This is to set the most optimal timeout defaults, max parallel requests, correct client decoration order
2. Encapsulate the complexity of the client configuration away from the user.
3. Configurable client configuration can then be exposed upon requests.
##### Disadvantages
1. If the client is used in another context other than the sync library, then it wouldn't be possible with the same client instance. Another client would have to be created.

heshamMassoud:
More context:

https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/repeater-task-queue.coffee
https://github.com/commercetools/zeg-cli/blob/master/lib/utils/common.js#L72

ahmetoz:
We have updated our client configuration in the previous release v3.0.0, all notes in this issue are handled. I am closing this out of date issue for now.

PR #609, opened 2020-10-28
610 - Fix duration calculation of ClientConfigurationUtils

ahmetoz:
Issue: https://github.com/commercetools/commercetools-sync-java/issues/610


src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
```diff
@@ -45,76 +56,74 @@
...
-        final int maxAttempts = 5;
-        final RetryAction scheduledRetry = RetryAction
-            .ofScheduledRetry(maxAttempts, context -> calculateVariableDelay(context.getAttempt()));
+    private static SphereClient withRetry(
+        @Nonnull final SphereClient delegate,
+        long maxRetryAttempt,
+        @Nonnull final Function<RetryContext, Duration> durationFunction) {
+
+        final RetryAction scheduledRetry = RetryAction.ofScheduledRetry(maxRetryAttempt, durationFunction);
         final RetryPredicate http5xxMatcher = RetryPredicate
```
lojzatran, line 103: Should it retry for all 5xx errors? Like we have here: https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/repeater-task-queue.coffee#L78. Additionally, we also have retry on some particular error statuses, maybe you can check them as well if they fit here.

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
```diff
@@ -18,21 +19,31 @@
...
 public final class ClientConfigurationUtils {
-    private static HttpClient httpClient;
-    private static final long DEFAULT_TIMEOUT = 30000;
-    private static final TimeUnit DEFAULT_TIMEOUT_TIME_UNIT = TimeUnit.MILLISECONDS;
-    private static Map<SphereClientConfig, SphereClient> delegatesCache = new HashMap<>();
+
+    protected static final long DEFAULT_TIMEOUT = 30000;
+    protected static final long INITIAL_RETRY_DELAY = 200;
+    protected static final TimeUnit DEFAULT_TIMEOUT_TIME_UNIT = TimeUnit.MILLISECONDS;
+    protected static final int MAX_RETRIES = 5;
```
lojzatran, line 35: Why only 5 times? We have in node-sdk 50 times: https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/repeater-task-queue.coffee#L50

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
```diff
@@ -45,76 +56,74 @@
...
-                    new DefaultAsyncHttpClientConfig.Builder()
-                                                    .setHandshakeTimeout((int) DEFAULT_TIMEOUT)
-                                                    .build());
-            httpClient = AsyncHttpClientAdapter.of(asyncHttpClient);
-        }
-        return httpClient;
+        final AsyncHttpClient asyncHttpClient =
+            new DefaultAsyncHttpClient(
+                new DefaultAsyncHttpClientConfig.Builder()
+                                                .setHandshakeTimeout((int) DEFAULT_TIMEOUT)
```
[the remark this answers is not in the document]
ahmetoz: That is a good question, also I am not sure about the timeout time, but the util is not focusing on a configurable one. Maybe we might discuss configurable client utils later.

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
ahmetoz: Actually, in commercetools docs, retry is suggested only for 503 and 409 (after resolving the version), so I am not sure when to retry. It is also interesting that the new node SDK is retrying only for 503. In this util class, in addition to 503, we are retrying for 502 and 504. Let me know if you know more details about this topic.

ahmetoz: I found an issue for the topic: https://github.com/commercetools/commercetools-sync-java/issues/208, I need to check the findings that we have before.

ahmetoz: I have checked the HTTP client library-`asynchttpclient`, and those values are is the defaults, comes from the file called, `ahc-defaults.properties`, I guess using 30 seconds handshake is not a big improvement here, not sure why it's needed, also in 208 its suggested less timeouts than defaults, which does not look right to me.

```` java
        final DefaultAsyncHttpClientConfig build = new DefaultAsyncHttpClientConfig.Builder()
            .build();
        System.out.println(build.getReadTimeout()); // 60000
        System.out.println(build.getConnectTimeout()); // 5000
        System.out.println(build.getHandshakeTimeout()); // 10000
        System.out.println(build.getRequestTimeout());; // 60000
````

File: src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java (hunk 2 above)
Comment (ahmetoz) [reply src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java:35]: 50 🤔 this does not look right to me, with exponential backoff it will not be easy. 

Here the delays for the 10 retries.
  ````
            Retry 1: 318 millisecond
            Retry 2: 740 millisecond
            Retry 3: 1284 millisecond
            Retry 4: 2002 millisecond
            Retry 5: 6054 millisecond
            Retry 6: 8690 millisecond
            Retry 7: 15567 millisecond
            Retry 8: 30000 millisecond
            Retry 9: 30000 millisecond
            Retry 10: 30000 millisecond
````

File: src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java (hunk 1 above)
Comment (lojzatran) [reply src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java:103]: > I have checked the HTTP client library-`asynchttpclient`, and those values are is the defaults, comes from the file called, `ahc-defaults.properties`, I guess using 30 seconds handshake is not a big improvement here, not sure why it's needed, also in 208 its suggested less timeouts than defaults, which does not look right to me.
> 
> ```java
>         final DefaultAsyncHttpClientConfig build = new DefaultAsyncHttpClientConfig.Builder()
>             .build();
>         System.out.println(build.getReadTimeout()); // 60000
>         System.out.println(build.getConnectTimeout()); // 5000
>         System.out.println(build.getHandshakeTimeout()); // 10000
>         System.out.println(build.getRequestTimeout());; // 60000
> ```

Are you talking about the retry for 5xx? Anyway, seems like you're right with the documentation saying you should retry only on 503 😕 So you can choose - either follow the official docs and retry only on 503 or retry on all 5xx errors 🙂  I don't know much about this topic, I guess you can create a support ticket to ask for sure.

File: src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java (hunk 2 above)
Comment (lojzatran) [reply src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java:35]: OK it's quite slow :| What if you keep 10 retries and don't do `final double exponentialFactor = Math.pow(2, retryAttempt - 1);` and do `INITIAL_RETRY_DELAY * retryAttempt * jitter`?

File: src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java (hunk 1 above)
Comment (butenkor) [reply src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java:103]: I would go with a retry on any 5xx errors as it might be just a temporal issue on one of the API nodes only. Defaults for read/connect/handshake timeouts seem fine.

File: src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java (hunk 2 above)
Comment (butenkor) [reply src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java:35]:
```
         Retry 8: 30000 millisecond
          Retry 9: 30000 millisecond
          Retry 10: 30000 millisecond
```
This does not seem right and might cause snowball effect. It should be always randomly growing. Maybe just compare if generated delay is bigger or equal default timeout and then stop retrying instead?

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
ahmetoz: I am not sure about the topic, and not sure why docs are suggesting only for 503. 

What about the 500? We don't know what is the problem that returns 500? If it's a node related problem so not a problem, it could be also the opposite, retrying all 5xx could also harm the platform if it's a bug, right ?

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
ahmetoz: Not sure how I could stop the retry, 
In our defaults, it's 5, so those cases are not possible. when we plan a retry with a configurable max retry count like 10 in this case, that might be an issue, in this case, I will increase the maximum timeout or might not put a max constant for the delay. Let me check.

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
ahmetoz: Maybe this case should be retried but not more than once, if it's a bug it will not be different, if not it might be resolved. WDYT?

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
JudeNiroshan: We have our upper bound set as `30000`. Once we reach this limit, we will no longer increase our delay.

ahmetoz:
The next parts will be handled by #623

PR #614, opened 2020-11-05
Concept of the shopping lists' line item update actions.

ahmetoz:
- Introduced ADR to document decisions.

Review Hint: 
It would be better to check the file preview otherwise hard to understand because of the HTML source code in markdown.

PR #618, opened 2020-11-11
Implement LineItem actions

ahmetoz:
Implementation of the line item actions. The PR should follow the decisions on the decided concept: #614 

### Todos:

- [X] Implement actions of the line item (`changeLineItemQuantity`, ` setLineItemCustomType`, `setLineItemCustomField`)
- [x] Implement matching algorithm of the line item lists.
- [x] Implement actions of the line item (`addLineItem `, `removeLineItem`)
- [x] Cover (`changeLineItemsOrder `)  with removing and adding back.
- [ ] What if no new line items and no order change -> case 3 in concept doc.
- [x] Update action for updating `addedAt` value. 

Note: Text line items will be planned as a different PR, to avoid confusions.


src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
```diff
@@ -0,0 +1,257 @@
...
+     *         identical. Otherwise, if the line items are identical, an empty list is returned.
+     */
+    @Nonnull
+    public static List<UpdateAction<ShoppingList>> buildLineItemsUpdateActions(
+        @Nonnull final ShoppingList oldShoppingList,
+        @Nonnull final ShoppingListDraft newShoppingList,
+        @Nonnull final ShoppingListSyncOptions syncOptions) {
+
+        boolean hasOldLineItems = oldShoppingList.getLineItems() != null && !oldShoppingList.getLineItems().isEmpty();
+        boolean hasNewLineItems = newShoppingList.getLineItems() != null && !newShoppingList.getLineItems().isEmpty()
```
leungkinghin, line 52: Add `final` keywords

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
```diff
@@ -0,0 +1,257 @@
...
+                                  .filter(Objects::nonNull)
+                                  .map(AddLineItemWithSku::of)
+                                  .collect(toList());
+        }
+
+        final List<LineItem> oldLineItems = oldShoppingList.getLineItems();
+        final List<LineItemDraft> newlineItems = newShoppingList.getLineItems()
+                                                                .stream()
+                                                                .filter(Objects::nonNull)
+                                                                .collect(toList());
```
leungkinghin, line 79: Can move to line 50, so that it is not required to check nonNull element in line 53 and 70. wdyt?

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
```diff
@@ -0,0 +1,257 @@
...
+        @Nonnull final LineItemDraft newLineItem,
+        @Nonnull final ShoppingListSyncOptions syncOptions) {
+
+        final List<UpdateAction<ShoppingList>> updateActions = filterEmptyOptionals(
+            buildChangeLineItemQuantityUpdateAction(oldLineItem, newLineItem)
+        );
+
+        updateActions.addAll(
+            buildLineItemCustomUpdateActions(oldShoppingList, newShoppingList, oldLineItem, newLineItem, syncOptions));
+
```
leungkinghin, line 189: Do we need not to cater **SetLineItemCustomField**?

src/test/java/com/commercetools/sync/shoppinglists/utils/LineItemListUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,559 @@
...
+        RES_ROOT + "shoppinglist-with-lineitems-sku-123-without-addedAt.json";
+
+    private static final ShoppingListSyncOptions SYNC_OPTIONS =
+        ShoppingListSyncOptionsBuilder.of(mock(SphereClient.class)).build();
+
+    private static final LineItemReferenceResolver lineItemReferenceResolver =
+        new LineItemReferenceResolver(SYNC_OPTIONS, getMockTypeService());
+
+    @Test
+    void buildLineItemsUpdateActions_WithEmptyNewAndNullOldLineItems_ShouldNotBuildActions() {
```
leungkinghin, line 92: Is it the same as this test case?
`void buildLineItemsUpdateActions_WithNullOldAndEmptyNewLineItems_ShouldNotBuildActions()`

leungkinghin:
The algorithm implemented in `buildLineItemsUpdateActions` looks good.

leungkinghin (approved):

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: It's part of it.

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: It's required, I will keep null checks and array creation separate.

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/618/commits/a54ab333f7253d144160690befebf76ba508a6aa

src/test/java/com/commercetools/sync/shoppinglists/utils/LineItemListUpdateActionUtilsTest.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/618/commits/a7a7e0a30b50ef509dd70e39f0b9a6e9ca9e79d8

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            if (oldLineItem.getVariant().getSku().equals(newLineItem.getSku())
+                && hasIdenticalAddedAtValues(oldLineItem, newLineItem)) {
+                // same sku, calculate actions.
+                updateActions.addAll(buildLineItemUpdateActions(
+                    oldShoppingList, newShoppingList, oldLineItem, newLineItem, syncOptions));
+            } else {
+                // different sku means the order is different.
```
lojzatran, line 128: Or order is the same but addedAt value is different?

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
```diff
@@ -0,0 +1,257 @@
...
+        return updateActions;
+    }
+
+    /**
+     * Compares the {@code quantity} values of a {@link LineItem} and a {@link LineItemDraft}
+     * and returns an {@link Optional} of update action, which would contain the {@code "changeLineItemQuantity"}
+     * {@link UpdateAction}. If both {@link LineItem} and {@link LineItemDraft} have the same
+     * {@code quantity} values, then no update action is needed and empty optional will be returned.
+     *
+     * <p>Note: If {@code quantity} from the {@code newLineItem} is {@code null} or {@code 0}, the new {@code quantity}
```
lojzatran, line 201: Why not set it to zero? Then it will be removed from the list.

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
```diff
@@ -0,0 +1,257 @@
...
+        @Nonnull final LineItemDraft newLineItem,
+        @Nonnull final ShoppingListSyncOptions syncOptions) {
+
+        return CustomUpdateActionUtils.buildCustomUpdateActions(
+            oldShoppingList,
+            newShoppingList,
+            oldLineItem::getCustom,
+            newLineItem::getCustom,
+            new LineItemCustomActionBuilder(),
+            -1, // not used by util.
```
lojzatran, line 250: Can you pass null instead? Or overload the method?

lojzatran (changes requested):

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
lojzatran: But why do we have to do the update actions because of `addedAt`? You cannot update it according to the platform, so we should not care here IF we are not going to preserve it (which is very complex).

src/main/java/com/commercetools/sync/shoppinglists/helpers/ShoppingListBatchValidator.java
```diff
@@ -96,7 +95,7 @@ private boolean isValidShoppingListDraft(
         } else {
             final List<String> draftErrors = getErrorsInAllLineItemsAndTextLineItems(shoppingListDraft);
             if (!draftErrors.isEmpty()) {
-                final String concatenatedErrors = draftErrors.stream().collect(joining(","));
+                final String concatenatedErrors = String.join(",", draftErrors);
```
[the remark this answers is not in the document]
ahmetoz: refactoring the warning.

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: 0 means removing the line item. https://docs.commercetools.com/api/projects/shoppingLists#change-lineitem-quantity

ahmetoz: Default is 1, so if not set it should be the default in sync context, if you want to remove don't put it to the draft.

lojzatran: True that it should not be there. But if it's there anyway with zero, then just update with zero. Don't put default there. It's not expected and additionally we have unnecessary data.

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: I could update this comment..

In sync, it's different projects (or external resources) so not the same purpose. Here maybe you have a front end and you want to order them correctly.  It's not complex but hard to do and can only be done with removing and adding back which requires order change too.

lojzatran: But it's not possible to do it with the platform, right? Why should we support it then? 
And even if you want to support it like you have it here, `addedAt` will not work if you change the order of line items by removing and adding. It will be overwritten anyway.

ahmetoz: I am not sure about this, maybe you are right we should not support because it's not supported as an action in API, @butenkor can you also check that?

It only makes sense for the front end which is ordering with `addedAt` values.

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: fefcd0d

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: yes makes more sense, https://github.com/commercetools/commercetools-sync-java/pull/618/commits/de059195b2ac171c4ed99e84f0e8624ca83e968d

src/main/java/com/commercetools/sync/shoppinglists/utils/LineItemUpdateActionUtils.java
ahmetoz: TextLineItem is also not supporting addedAt changes, looks like forcing a change in here would not be necessary until we have a valid requirement from users of the library. I haven't implemented it for the text line items, and planning to remove it from line item actions too.

lojzatran (approved):

Issue #582, opened 2020-09-29
Improve the performance of the cacheKeytoIds methods using grapqhl and configurable cache implementation.


ahmetoz:
**Situation:**
We are using rest calls for fetching ids' of the resources with the given keys in the library's service classes called cacheKeysToIds for every type in the library, for example, ProductService has cacheKeysToIds, it takes a set of keys to fetch ids from the commercetools project.

In the service classes of the Sync library, we have implemented an in-memory cache to store a hash map for commercetools reference key to reference ids. This is used for the reference resolution of the library, this is a simple cache to reduce reference resolution based calls to the commercetools API, so the library will go only one time to fetch an id of resource, so the other references that are referencing the same resource might use the id in the cache instead of going to the commercetools API, which will improve the overall performance of the sync and commercetools API.

**Complication:**

Using rest calls for fetching just an id is not a good approach, so it could decrease the performance and bad memory consumption of the library when the customer has a huge number of resources because, with rest call, the library will fetch all fields from commercetools but not just ids. One of the problems in rest is resolved by graphql that is fetching only the needed fields with a query. Also, we might use one query to fetch from multiple resources to improve performance.

The second problem to the existing code base is that we are using a hashmap for caching purposes for id to the key, of cacheToKeysToIds result, this is implemented by using a concurrent hash map and does not support an evict strategy. 
Here an evict strategy is needed otherwise we could store millions of keys to ids entry in a simple hashmap, which is not good for performance and memory consumption. 
The cache size should be configurable by the customer (with sync options and a good default), This means we could have a good memory footprint instead of storing millions of keys to ids maps in our sync implementations, which could cause memory problems if the process does not have enough memory and CPU resource (for example in a container environment)

As fetching and storing the result handled by cacheKeysToIds methods in service implementations, so this 2 improvement should be handled in one task because it's part of the same context.

Related context needs to be fixed together:
#84, #403 

Please consider the comments for the method structures of cacheKeyToIds:
https://github.com/commercetools/commercetools-sync-java/pull/581#discussion_r496607931
https://github.com/commercetools/commercetools-sync-java/pull/581#discussion_r496608941

PR #628, opened 2020-11-18
Improve current caching solution

salander85:
#### Summary
Add a cache implementation which can be evictable by size

#### Description
Use Caffeine with a maximum size which can remove not recently used entries from cache beyond that given size using the
`Window TinyLfu` policy due to its high hit rate and low memory footprint.

#### Relevant Issues
https://github.com/commercetools/commercetools-sync-java/issues/582

#### Todo

- Tests
    - [x] Unit 

#### Hints for Review
There's only one UT for the caching solution - do you think we need to test more cases?


src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java
```diff
@@ -68,6 +69,24 @@ public T batchSize(final int batchSize) {
         return getThis();
     }
 
+    /**
+     * Set option that indicates cache size for sync process. To increase performance during the sync some resource keys
```
ahmetoz, line 73:
```suggestion
     * Sets the cache size that indicates the key to id cache size of the sync process.To increase performance during the sync some resource keys
```

src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java
```diff
@@ -68,6 +69,24 @@ public T batchSize(final int batchSize) {
...
+     * -> ids are cached which are required for resolving references. To keep the cache performant outdated entries are
+     * evicted when a certain size is reached.
+     *
+     * <p>This cache size is set to 10.000 by default.
```
ahmetoz, line 77:
```suggestion
     * <p>Note: This cache size is set to 10.000 by default.
```

src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java
```diff
@@ -68,6 +69,24 @@ public T batchSize(final int batchSize) {
...
+     *
+     * @param cacheSize long that indicates cache size of the key to id cache used for reference resolution. Has to be
```
ahmetoz, line 79:
```suggestion
     * @param cacheSize a long number value that indicates cache size of the key to id cache used for reference resolution. Has to be
```

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -223,7 +224,10 @@
 
         return CtpQueryUtils
             .queryAll(syncOptions.getCtpClient(), keysQueryMapper.apply(keysNotCached), pageConsumer)
-            .thenApply(result -> keyToIdCache.asMap());
+            .thenApply(result -> {
+                keyToIdCache.cleanUp();
```
ahmetoz, line 228: Why we need a cleanup, and what is the benefit? 

I would assume the cleanup will be run when the cache reached some limit? Did you try it out?

ahmetoz: OK, I see what you did here, I tested quickly like below and understand eviction is not automatic.

```` java
     final Cache<String, String> testCache = Caffeine.newBuilder()
                                                    .maximumSize(10)
                                                    .build();
        for (int i = 0; i < 100; i++) {
            testCache.put("key" + i, "val" + i);
        }
````

Comment (salander85) [reply src/main/java/com/commercetools/sync/services/impl/BaseService.java:228]: Yes, cleanup is running when cache reached the size limit asynchronously. Actually i added to be able to test the eviction. Cannot access the `keyToIdCache` field from tests... - i could make it accessible for the test. WDYT?

Comment (ahmetoz) [reply src/main/java/com/commercetools/sync/services/impl/BaseService.java:228]: But the problem is different because the code is using locks/threads and not really performant to run for every call 🤔

File: src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java (hunk 3 above)
Comment (salander85) [reply src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java]: 07f051e

File: src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java (hunk 2 above)
Comment (salander85) [reply src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java]: 07f051e

File: src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java (hunk 1 above)
Comment (salander85) [reply src/main/java/com/commercetools/sync/commons/BaseSyncOptionsBuilder.java]: 07f051e

File: src/main/java/com/commercetools/sync/services/impl/BaseService.java (hunk 4 above)
Comment (leungkinghin) [reply src/main/java/com/commercetools/sync/services/impl/BaseService.java:228]: I wrote a simple method like this 
```
        Cache<String, String> cache = Caffeine.newBuilder().maximumSize(10).build();

        for (int i = 0; i<100 ; ++i) {
            cache.put(String.valueOf(i), String.valueOf(i));

        }
        System.out.println(cache.estimatedSize()); // printout 100
        Thread.sleep(1000);
        System.out.println(cache.estimatedSize()); // printout 10
```
As mentioned, cleanup is running when cache reached the size limit asynchronously. It automatically triggers and takes a second to do eviction. So I guess we don't need to `cleanUp()` manually.

Comment (leungkinghin) [reply src/main/java/com/commercetools/sync/services/impl/BaseService.java:228]: Another small test
 ```
       Cache<String, String> cache = Caffeine.newBuilder()
                                                .maximumSize(10)
                                                .build();

        for (int i = 0; i<100 ; ++i) {
            Thread.sleep(10);
            cache.put(String.valueOf(i), String.valueOf(i));

        }
        System.out.println(cache.estimatedSize()); // printout 11
        Thread.sleep(1000);
        System.out.println(cache.estimatedSize()); // printout 10
```
It proves that high cache size produced before because cache input is quicker than eviction.

leungkinghin: @salander85 I don't suggest to change method access modifier easily for test case purpose. Your added unit test case is good enough to verify eviction. I can't find any other scenario we can verify the eviction.

leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/628/commits/120f43bb2d59c9bbc3f856c86d3a85bba651f21c

leungkinghin: However, without explicitly cleanup the cache in every call, i found the test class fails all the time. I have to insert few milliseconds delay in eviction test case to make it run properly. I guess cache handles the operation asynchronously when running test class and cause the problem.

ben-manes: fyi, you can use `Caffeine.executor(Runnable::run)` to perform the eviction on the calling thread.

salander85: 67aad19

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -223,7 +225,7 @@
 
         return CtpQueryUtils
             .queryAll(syncOptions.getCtpClient(), keysQueryMapper.apply(keysNotCached), pageConsumer)
-            .thenApply(result -> keyToIdCache.asMap());
+            .thenApply(result ->  keyToIdCache.asMap());
```
ahmetoz, line 228:
```suggestion
            .thenApply(result -> keyToIdCache.asMap());
```

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -268,7 +270,7 @@
             keyToIdCache.putAll(keyToIdMapper.apply(resource)));
 
         return CtpQueryUtils.queryAll(syncOptions.getCtpClient(), keysRequestMapper.apply(keysNotCached),
-            resultConsumer).thenApply(result -> keyToIdCache.asMap());
+            resultConsumer).thenApply(result ->  keyToIdCache.asMap());
```
ahmetoz, line 273:
```suggestion
            resultConsumer).thenApply(result -> keyToIdCache.asMap());
```

leungkinghin (approved):

src/test/java/com/commercetools/sync/services/impl/BaseServiceImplTest.java
```diff
@@ -292,6 +296,37 @@ void cacheKeysToIdsUsingGraphQl_WithAllCachedKeys_ShouldMakeNoRequestAndReturnCa
         verify(client, times(1)).execute(any(ProductQuery.class));
     }
 
+    @Test
+    void cacheKeysToIds_WithCachedKeysExceedingCacheSize_ShouldEvictOldEntriesAndReturnLatestUsed() {
```
ahmetoz, line 300:
```suggestion
    void cacheKeysToIds_WithCachedKeysExceedingCacheSize_ShouldNotReturnLeastUsedKeys() {
```

ahmetoz (changes requested):
You have to adapt to changes with the `master` branch. so shopping lists service.

src/main/java/com/commercetools/sync/services/impl/BaseService.java
ahmetoz: @ben-manes thanks for the suggestion 👍🏽

src/test/java/com/commercetools/sync/services/impl/BaseServiceImplTest.java
salander85: f8e4454

src/main/java/com/commercetools/sync/services/impl/BaseService.java
salander85: f8e4454

src/main/java/com/commercetools/sync/services/impl/BaseService.java
salander85: f8e4454

salander85:
> You have to adapt to changes with the `master` branch. so shopping lists service.

Done in base branch https://github.com/commercetools/commercetools-sync-java/commit/2b632b76ef8a7159cabbc44b412939d4f8608495

ahmetoz:
> > You have to adapt to changes with the `master` branch. so shopping lists service.
> 
> Done in base branch [2b632b7](https://github.com/commercetools/commercetools-sync-java/commit/2b632b76ef8a7159cabbc44b412939d4f8608495)

PR should have everything in the base branch, this PR does not, base branch has shopping lists only for grapql changes but caching includes the cache size change, I am not sure about that. Correct me if I am wrong.

salander85:
> > > You have to adapt to changes with the `master` branch. so shopping lists service.
> > 
> > 
> > Done in base branch [2b632b7](https://github.com/commercetools/commercetools-sync-java/commit/2b632b76ef8a7159cabbc44b412939d4f8608495)
> 
> PR should have everything in the base branch, this PR does not, base branch has shopping lists only for grapql changes but caching includes the cache size change, I am not sure about that. Correct me if I am wrong.

you're right, i didn't remember the cache size change

ahmetoz:
> > > > You have to adapt to changes with the `master` branch. so shopping lists service.
> > > 
> > > 
> > > Done in base branch [2b632b7](https://github.com/commercetools/commercetools-sync-java/commit/2b632b76ef8a7159cabbc44b412939d4f8608495)
> > 
> > 
> > PR should have everything in the base branch, this PR does not, base branch has shopping lists only for grapql changes but caching includes the cache size change, I am not sure about that. Correct me if I am wrong.
> 
> you're right, i didn't remember the cache size change. [50c91a8](https://github.com/commercetools/commercetools-sync-java/commit/50c91a8eba6e564c366c1f4907c73880dfabef8b)

@salander85  Please prepare a `commented` release note for your change and sync with the master branch, then we could merge it.

docs/RELEASE_NOTES.md
```diff
@@ -27,7 +27,18 @@
 7. Add Migration guide section which specifies explicitly if there are breaking changes and how to tackle them.
 
 -->
+<!--
+### 3.0.1 - Sep 29, 2020
```
ahmetoz, line 31:
```suggestion
### 3.0.1 - MM DD, 2020
```

ahmetoz (approved):
🎉 good work @salander85

PR #622, opened 2020-11-12
251 custom decorator fix product test

leungkinghin:
#### Summary
Resolve test case for Product Resources

#### Description
After client decoration in next featured branch https://github.com/commercetools/commercetools-sync-java/commit/e57c36407487c0813bb3761b170df244c8628e83, we have to resolve test cases failure in IT and UT. The first step is to implement different resolutions for product test cases.

#### Relevant Issues
#251 

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [ ] Documentation
- [ ] Add Release Notes entry.

Issue #251, opened 2018-02-14
Decorate client with Header client


heshamMassoud:
### Description
In order to ensure that the client always has the version of the sync library in it's headers, it could be that the SolutionInfo is not enough since, It could be overridden by another solutionInfo from the user application. Related: https://github.com/commercetools/commercetools-jvm-sdk/issues/1691

### Current Behaviour
The version is injected from the release name into SyncSolutionInfo which adds the version in the headers of the client.

### Expected Behaviour
The version should be appended to the HTTP requests headers flexibly using a custom decorator. Check more here: https://github.com/commercetools/commercetools-jvm-sdk/issues/1691#issuecomment-365202896

Issue #578, opened 2020-09-28
Fix all compile time warnings in project


ahmetoz:
1- hasOnlyOneElementSatisfying is deprecated, we need to use singleElement()

```
warning: [deprecation] hasOnlyOneElementSatisfying(Consumer<? super ELEMENT>) in AbstractIterableAssert has been deprecated
            .hasOnlyOneElementSatisfying(throwable -> {
            ^
```

For example:
````java
assertThat(errorMessages)
            .hasSize(1)
            .hasOnlyOneElementSatisfying(message -> {
                assertThat(message).contains("Failed to create draft with key: 'productKey'.");
                assertThat(message).contains("BadRequestException");
            });
````
to
````java
assertThat(errorMessages)
            .hasSize(1)
            .singleElement(as(STRING))
            .contains("Failed to create draft with key: 'productKey'.")
            .contains("BadRequestException");
````

second example:
````java
 assertThat(errorCallBackExceptions).hasOnlyOneElementSatisfying(throwable -> {
            assertThat(throwable).isInstanceOf(SyncException.class);
            assertThat(throwable).hasMessage(PRODUCT_TYPE_DRAFT_IS_NULL);
        });
````

to 

````java

        assertThat(errorCallBackExceptions).singleElement()
                                           .withFailMessage(PRODUCT_TYPE_DRAFT_IS_NULL)
                                           .isInstanceOf(SyncException.class);
````

Issue #142, opened 2017-10-11
Add lib version in additional CTP client headers


heshamMassoud:
- [x] Add lib version in additional CTP client headers

PR #636, opened 2020-11-27
Fix test cases for custom decorator 

leungkinghin:
#### Summary
Resolve test cases failure after client decoration in `BaseSyncOptions`

#### Description
Resolve test cases failure in different ways
- For Service test classes, implement a fake client and inject desired result into it.
- For Sync test classes, find out appropriate return result from service class and mock it, instead of mocking `SphereClient` from `SyncOptions`.
- For SyncOptionsBuilder test classes, do the assertion on decorated client.

#### Relevant Issues
#251 

#### Todo

- Tests
    - [x] Unit 
    - [x] Integration
- [ ] Documentation

#### Hints for Review
For general, please refer to this sample PR
https://github.com/commercetools/commercetools-sync-java/pull/622

Following test cases in `StateSyncIT` cannot move to unit test and execute successfully.
sync_WithUpdatedTransitionAndClientThrowsError_ShouldAddErrorMessage
sync_WithExceptionWhenFetchingUnresolvedTransition_ShouldPrintErrorMessage
sync_withCreateStateException_shouldPrintMessage
sync_withNewStateWihCreationException_shouldPrintErrorMessage
sync_withNewCustomObjectAndBadRequest_shouldNotCreateButHandleError

We adapt the implementation as shown below
```
stateSyncOptions = spy(StateSyncOptionsBuilder
                .of(sphereClient)
                .errorCallback((exception, oldResource, newResource, updateActions) -> {
                    errorMessages.add(exception.getMessage());
                    errorExceptions.add(exception.getCause());
                })
                .build());
when(stateSyncOptions.getCtpClient()).thenReturn(sphereClient);
```

Following test cases in `UnresolvedTransitionServiceImplTest` and `UnresolvedReferencesServiceImplTest` cannot adapt fake client because ArgumentCaptor is used to verify the hashed URL


src/integration-test/java/com/commercetools/sync/integration/externalsource/customobjects/CustomObjectSyncIT.java
```diff
@@ -115,133 +108,6 @@ void sync_withExistingCustomObjectThatHasSameValue_shouldNotUpdateCustomObject()
...
-        Assertions.assertThat(errorCallBackMessages).hasSize(1);
-        Assertions.assertThat(errorCallBackExceptions).hasSize(1);
-
-        Assertions.assertThat(errorCallBackMessages.get(0)).contains(
-            format("Failed to update custom object with key: '%s'. Reason: Not found when attempting to fetch while"
-                + " retrying after concurrency modification.", CustomObjectCompositeIdentifier.of(customObjectDraft)));
-    }
-
     @Test
     void sync_withNewCustomObjectAndBadRequest_shouldNotCreateButHandleError() {
```
ahmetoz, line 112: Do you think that we need to move this into unit tests? 🤔

src/integration-test/java/com/commercetools/sync/integration/externalsource/states/StateSyncIT.java
```diff
@@ -943,7 +744,10 @@ void sync_WithUpdatedTransitionAndClientThrowsError_ShouldAddErrorMessage() {
             })
             .warningCallback((exception, newResource, oldResource) ->
                 warningCallBackMessages.add(exception.getMessage()))
-            .build();
+            .build());
+
+        when(stateSyncOptions.getCtpClient()).thenReturn(spyClient);
```
ahmetoz, line 749: I assume some integration tests in the StateSyncIT could be moved to unit tests, any reason to no move? 

`sync_withCreateStateException_shouldPrintMessage`
`sync_withNewStateWihCreationException_shouldPrintErrorMessage`
`sync_WithExceptionWhenFetchingUnresolvedTransition_ShouldPrintErrorMessage`
`sync_WithUpdatedTransitionAndClientThrowsError_ShouldAddErrorMessage`

Actually those looks like a unit test becasue of mocking execute of client .

src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
```diff
@@ -3,7 +3,7 @@
 import io.sphere.sdk.client.SolutionInfo;
 
 public final class SyncSolutionInfo extends SolutionInfo {
-    private static final String LIB_NAME = "commercetools-sync-java";
+    public static final String LIB_NAME = "commercetools-sync-java";
```
ahmetoz, line 6: As we are targeting the add version header within the CustomHeaderSphereClientDecorator, do we need to remove the solution info metafile from the resources? WDYT? 

https://github.com/commercetools/commercetools-sync-java/blob/master/src/main/resources/META-INF/services/io.sphere.sdk.client.SolutionInfo

src/main/java/com/commercetools/sync/products/templates/beforeupdatecallback/KeepOtherVariantsSync.java
```diff
@@ -9,7 +9,7 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
-public class KeepOtherVariantsSync {
+public final class KeepOtherVariantsSync {
```
ahmetoz, line 12: Why do you need this breaking change?

src/test/java/com/commercetools/sync/internals/helpers/CustomHeaderSphereClientDecoratorTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.jupiter.api.Test;
+
+import static com.commercetools.sync.commons.utils.SyncSolutionInfo.LIB_VERSION;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+
+public class CustomHeaderSphereClientDecoratorTest {
+
+    private static final String LIB_NAME = "commercetools-sync-java";
```
ahmetoz, line 22: you might use the SyncSolutionInfo.LIB_NAME

src/test/java/com/commercetools/sync/internals/helpers/CustomHeaderSphereClientDecoratorTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import io.sphere.sdk.producttypes.commands.updateactions.ChangeName;
+import org.junit.jupiter.api.Test;
+
+import static com.commercetools.sync.commons.utils.SyncSolutionInfo.LIB_VERSION;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+
+public class CustomHeaderSphereClientDecoratorTest {
```
ahmetoz, line 20:
```suggestion 
class CustomHeaderSphereClientDecoratorTest {
```

ahmetoz (changes requested):
Good job! with refactoring the tests better than before, it will benefit a lot. 👍🏽 

Thanks for fixing some warnings, before I opened an issue for the assertj warnings: https://github.com/commercetools/commercetools-sync-java/issues/578 maybe check and resolve the issue if it's complete. 🙏🏽

src/integration-test/java/com/commercetools/sync/integration/externalsource/states/StateSyncIT.java
leungkinghin: `sync_withCreateStateException_shouldPrintMessage`
`sync_withNewStateWihCreationException_shouldPrintErrorMessage`
- For creation resources, the exception has been handled in `BaseService` class.  
https://github.com/commercetools/commercetools-sync-java/blob/471f863328aafbcb9326be171ae97ae64d60e30b/src/main/java/com/commercetools/sync/services/impl/BaseService.java#L306
If we move them to unit test class, empty object is mocked instead of a throwable object when calling method of `StateServiceImpl`. My concern is that it is a bit confusing to developer, because no exception class is needed in test case, so i keep it in integration test.

`sync_WithExceptionWhenFetchingUnresolvedTransition_ShouldPrintErrorMessage`
- No way to inject `UnresolvedTransitionsServiceImpl` into `StateSync`, which is required to do mocking in unit test.

`sync_WithUpdatedTransitionAndClientThrowsError_ShouldAddErrorMessage`
- I tried to move it but failed. When inputting three drafts together into sync process, the behaviour between IT and UT is different. In IT, the mocked exception can interrupt the entire sync process and quit. But In UT, the mocked exception does not affect sync process of other two drafts, and they continue to sync. Probably we can create follow-up ticket in future to investigate this question.

src/main/java/com/commercetools/sync/products/templates/beforeupdatecallback/KeepOtherVariantsSync.java
leungkinghin: Added a private constructor for coverage rate.
Let me roll it back and see whether i can implement else classes to raise up the coverage.

src/test/java/com/commercetools/sync/internals/helpers/CustomHeaderSphereClientDecoratorTest.java
leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/636/commits/843a09c90ff708ec0ad4c6a6ecf3de92a8e62cbb

src/test/java/com/commercetools/sync/internals/helpers/CustomHeaderSphereClientDecoratorTest.java
leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/636/commits/843a09c90ff708ec0ad4c6a6ecf3de92a8e62cbb

praveenkumarct (approved):
Good work on moving tests and fixing them :)

src/integration-test/java/com/commercetools/sync/integration/externalsource/customobjects/CustomObjectSyncIT.java
leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/636/commits/c1e492b1dfa42b4f1ed55e9adb5c5bb88db4e9d7

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncWithReferencedProductsInAnyOrderIT.java
```diff
@@ -58,16 +55,11 @@
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singleton;
-import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.when;
+
```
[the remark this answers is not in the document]
leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/636/commits/131c7dd6732d9ec8c6776b54a38d7ebad001e295

src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
leungkinghin: I have seen Hesham's PR and the metafile is created at that moment
https://github.com/commercetools/commercetools-sync-java/issues/142

On the other hand, I found now the library version injection is done by this, which is added by Hesham later.
https://github.com/commercetools/commercetools-sync-java/blob/251-custom-decorator-fix-test/gradle-scripts/set-library-version.gradle

@butenkor Can you confirm that the solution info metafile is deprecated already?

src/integration-test/java/com/commercetools/sync/integration/externalsource/states/StateSyncIT.java
ahmetoz: Good to know the challenges, it looks like the effort needed for these changes does not needed at the moment, maybe create an issue for the descriptions here and put our issue list, what do you think?

src/main/java/com/commercetools/sync/products/templates/beforeupdatecallback/KeepOtherVariantsSync.java
leungkinghin: I just added test case elsewhere to raise up the coverage.
Here I revert the commit.
https://github.com/commercetools/commercetools-sync-java/pull/636/commits/762a065e970d1ae75d0392ffdc0fcbc66bdcc619

src/integration-test/java/com/commercetools/sync/integration/externalsource/states/StateSyncIT.java
leungkinghin: Agree. I would create issue later on .

src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
ahmetoz: First of all, our SyncSolutionInfo might stay as you will need the version in Gradle, also the Gradle script is make sense to automate getting the latest version of the library.

The work here is started to create an alternative of SolutionInfo for adding the version headers using decorators. Now it looks that we have it.

Here, I would remove the solution info metafile, and also remove the extending SolutionInfo of the SyncSolutionInfo, and run an integration test and ensure the new headers in the API.

The patch: 
````
Index: src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java	(revision 471f863328aafbcb9326be171ae97ae64d60e30b)
+++ src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java	(date 1606755014333)
@@ -1,25 +1,9 @@
 package com.commercetools.sync.commons.utils;
 
-import io.sphere.sdk.client.SolutionInfo;
-
-public final class SyncSolutionInfo extends SolutionInfo {
+public final class SyncSolutionInfo {
     public static final String LIB_NAME = "commercetools-sync-java";
     /**
      * This value is injected by the script at gradle-scripts/set-library-version.gradle.
      */
     public static final String LIB_VERSION = "#{LIB_VERSION}";
-
-    /**
-     * Extends {@link SolutionInfo} class of the JVM SDK to append to the User-Agent header with information of the
-     * commercetools-sync-java library
-     *
-     * <p>A User-Agent header with a solution information looks like this:
-     * {@code commercetools-jvm-sdk/1.46.1 (AHC/2.0) Java/1.8.0_92-b14 (Mac OS X; x86_64)
-     * {@value LIB_NAME}/{@value LIB_VERSION}}</p>
-     *
-     */
-    public SyncSolutionInfo() {
-        setName(LIB_NAME);
-        setVersion(LIB_VERSION);
-    }
-}
\ No newline at end of file
+}
diff --git src/main/resources/META-INF/services/io.sphere.sdk.client.SolutionInfo src/main/resources/META-INF/services/io.sphere.sdk.client.SolutionInfo
deleted file mode 100644
index 7ade83f6f85b957eff000084d8b2940c727c4820..0000000000000000000000000000000000000000
GIT binary patch
literal 0
Hc$@<O00001
`````

Comment (leungkinghin) [reply src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java:6]: I just run IT locally and check the logs in Kibana.
Inheritance from SolutionInfo is a must, because it provides setter method to SyncSolutionInfo to set library name and version.

I tried to remove the metafile and run the test. The library name and version immediately disappear in User-Agent in request header log. For this reason i would keep the metafile.

Comment (ahmetoz) [reply src/main/java/com/commercetools/sync/commons/utils/SyncSolutionInfo.java:6]: >Inheritance from SolutionInfo is a must, because it provides setter method to SyncSolutionInfo to set library name and version.

So the solution that you have found as a custom decorator not working ? 🤔 Please check the old discussions of this work, I would expect without a metafile and solution info we should be adding headers 🤔

File: src/main/java/com/commercetools/sync/internals/helpers/CustomHeaderSphereClientDecorator.java [hunk 8]
```diff
@@ -43,7 +43,7 @@ protected HttpRequestIntent getHttpRequestIntent() {
         @Override
         public HttpRequestIntent httpRequestIntent() {
             final HttpHeaders headers =
-                    HttpHeaders.of(HttpHeaders.USER_AGENT, LIB_NAME + "( ver : " + LIB_VERSION + ")");
+                    HttpHeaders.of(HttpHeaders.USER_AGENT, LIB_NAME + " (ver : " + LIB_VERSION + ")");
```
ahmetoz, line 46: why not using `lib_name/lib_version`

leungkinghin: @ahmetoz 
To distinguish the change between old approach (adapt SolutionInfo) and new approach (decorator) after release.

ahmetoz (approved):
Note: Should not merge to master yet.

PR #642, opened 2020-12-14
Document cacheSize option in usage docs

ahmetoz:
Almost all of them is the same, so PR could be based on only one file, then I could update the other files :)

### DIFF
--- a/docs/usage/PRODUCT_SYNC.md
+++ b/docs/usage/PRODUCT_SYNC.md
@@ -253,8 +253,8 @@ final ProductSyncOptions productSyncOptions =
 ````
 
 ##### cacheSize
-In the service classes of the commercetools-sync-java library, we have implemented an in-memory [LRU cache](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) to store a map for commercetools reference key to reference ids.  This is used for the reference resolution of the library to reduce reference resolution based calls to the 
-commercetools API, so the library will go only one time to fetch an id of resource, so the other references that are referencing the same resource might use the id in the cache instead of going to the commercetools API, 
+In the service classes of the commercetools-sync-java library, we have implemented an in-memory [LRU cache](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) to store a map used for the reference resolution of the library.
+The cache reduces the reference resolution based calls to the commercetools API as the required fields of a resource will be fetched only one time. This cached fields then might be used by another resource referencing the already resolved resource instead of fetching from commercetools API. It turns out, having the in-memory LRU cache will improve overall performance of the sync library and commercetools API.
 which will improve the overall performance of the sync and commercetools API.
 
 Playing with this option can change the memory usage of the library. If it is not set, the default cache size is `10.000` for product sync.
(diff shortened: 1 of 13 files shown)


docs/usage/PRODUCT_SYNC.md
```diff
@@ -251,6 +252,18 @@ final ProductSyncOptions productSyncOptions =
          ProductSyncOptionsBuilder.of(sphereClient).batchSize(50).build();
 ````
 
+##### cacheSize
+In the service classes of the commercetools-sync-java library, we have implemented an in-memory [LRU cache](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) to store a map for commercetools reference key to reference ids.  This is used for the reference resolution of the library to reduce reference resolution based calls to the
```
salander85, line 256: Do we need to tell whats stored in the map?
```suggestion
In the service classes of the commercetools-sync-java library, we have implemented an in-memory [LRU cache](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) to store a map used for the reference resolution of the library.
```

docs/usage/PRODUCT_SYNC.md
```diff
@@ -251,6 +252,18 @@ final ProductSyncOptions productSyncOptions =
          ProductSyncOptionsBuilder.of(sphereClient).batchSize(50).build();
 ````
 
+##### cacheSize
+In the service classes of the commercetools-sync-java library, we have implemented an in-memory [LRU cache](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) to store a map for commercetools reference key to reference ids.  This is used for the reference resolution of the library to reduce reference resolution based calls to the 
+commercetools API, so the library will go only one time to fetch an id of resource, so the other references that are referencing the same resource might use the id in the cache instead of going to the commercetools API,
```
salander85, line 257: Maybe shorten this sentence to make it easier to read. 
```suggestion
The cache reduces the reference resolution based calls to the commercetools API as the required fields of a resource will be fetched only one time. This cached fields then might be used by another resource referencing the already resolved resource instead of fetching from commercetools API. It turns out, having the in-memory LRU cache will improve overall performance of the sync library and commercetools API.
```

salander85 (changes requested):

salander85 (approved):

docs/usage/PRODUCT_SYNC.md
```diff
@@ -251,6 +252,18 @@ final ProductSyncOptions productSyncOptions =
          ProductSyncOptionsBuilder.of(sphereClient).batchSize(50).build();
 ````
 
+##### cacheSize
+In the service classes of the commercetools-sync-java library, we have implemented an in-memory [LRU cache](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) to store a map used for the reference resolution of the library.
+The cache reduces the reference resolution based calls to the commercetools API as the required fields of a resource will be fetched only one time. This cached fields then might be used by another resource referencing the already resolved resource instead of fetching from commercetools API. It turns out, having the in-memory LRU cache will improve overall performance of the sync library and commercetools API.
```
praveenkumarct, line 257:
```suggestion
The cache reduces the reference resolution based calls to the commercetools API as the required fields of a resource will be fetched only one time. This cached fields then might be used by another resource referencing the already resolved resource instead of fetching from commercetools API. It turns out, having the in-memory LRU cache will improve overall performance of the sync library and commercetools API,
```

praveenkumarct (approved):
Looks good, Thanks for updating documents.

PR #644, opened 2020-12-15
Adapt changes of JVM-SDK 1.56.0

ahmetoz:
### Changes:

- Use the RetryableSphereClientBuilder in jvm-sdk-convenience.
- Adapt the changes of JVM-SDK, Address selection by key is supported so use the actions from SDK instead of creating custom update action implementations.


butenkor (approved):
👍

praveenkumarct (approved):

PR #677, opened 2021-02-18
Signing mvn deployments

ahalberkamp:
In this pr the following was done:

1)Signing the artefacts before deploying to sonatype
2)Improve documentation.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,6 +3,8 @@ distributionPath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+
```
ahalberkamp, line 6:
```suggestion
```

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,6 +3,8 @@ distributionPath=wrapper/dists
...
+
```
ahalberkamp:
```suggestion
```

docs/BUILD.md
```diff
@@ -5,6 +5,10 @@
 - [Release workflow](#release-workflow)
   - [Step 1: Pull request](#step-1-pull-request)
   - [Step 2: Release library in Github](#step-2-release-library-in-github)
+  - [Step 3: Publish to Maven central](#step-3-publish-to-maven-central)
```
ahmetoz, line 8: I wonder why we are not automating the release from "OSSRH" to "maven central" ? In that way, we don't need to document or do extra steps.
 
Check step 5: https://blog.sonatype.com/what-publishers-need-to-know-about-migrating-from-jcenter-/-bintray-to-the-central-repository

gradle-scripts/maven-publish.gradle
```diff
@@ -53,3 +50,17 @@ publishing {
         }
     }
 }
+
+signing {
+    def signingKey = System.getenv("PGP_KEY")
```
ahmetoz, line 55: Can we recreate those when we lose the credentials? And what is the best way to share it in team ?

ahmetoz (changes requested):

gradle-scripts/maven-publish.gradle
ahalberkamp: Yes, the GPG-Key can easily be recreated using GPG and uploaded to github.

ahalberkamp: I think if we lost the key, we can just create a new one...-

docs/BUILD.md
ahalberkamp: 217cbaa31

PR #688, opened 2021-03-08
use Chunk utils

praveenkumarct:
Use Chunk utils for running Query and caching.

CHUNK_SIZE - has to be decided. By default I have considered it same as for Id value of 41 characters.
But I see while creating ProductType with Key - Placeholder says as below - 
  "May only contain between 2 and 256 alphanumeric characters, underscores, or hyphens (no white space or special characters like ñ, ü, #, %) " 
  Let me know your suggestion.

### DIFF
--- a/src/main/java/com/commercetools/sync/services/impl/BaseService.java
+++ b/src/main/java/com/commercetools/sync/services/impl/BaseService.java
@@ -2,12 +2,14 @@ package com.commercetools.sync.services.impl;
 
 import static com.commercetools.sync.commons.utils.SyncUtils.batchElements;
 import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 import com.commercetools.sync.commons.BaseSyncOptions;
 import com.commercetools.sync.commons.exceptions.SyncException;
 import com.commercetools.sync.commons.helpers.ResourceKeyIdGraphQlRequest;
 import com.commercetools.sync.commons.models.ResourceKeyId;
+import com.commercetools.sync.commons.utils.ChunkUtils;
 import com.commercetools.sync.commons.utils.CtpQueryUtils;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
@@ -18,6 +20,7 @@ import io.sphere.sdk.commands.UpdateCommand;
 import io.sphere.sdk.models.ResourceView;
 import io.sphere.sdk.queries.MetaModelQueryDsl;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -59,6 +62,12 @@ abstract class BaseService<
 
   private static final int MAXIMUM_ALLOWED_UPDATE_ACTIONS = 500;
   static final String CREATE_FAILED = "Failed to create draft with key: '%s'. Reason: %s";
+  /*
+   * A key is a 41 characters long hashed string. (i.e: c25a67e262265fbc36edb33ed0375263586be278) We
+   * chunk them in 250 keys we will have around a query around 11.000 characters. Above this size it
+   * could return - Error 413 (Request Entity Too Large)
+   */
+  static final int CHUNK_SIZE = 250;
 
   BaseService(@Nonnull final S syncOptions) {
     this.syncOptions = syncOptions;
@@ -222,7 +231,8 @@ abstract class BaseService<
 
   /**
    * Given a set of keys this method caches a mapping of the keys to ids of such keys only for the
-   * keys which are not already in the cache.
+   * keys which are not already in the cache. This method uses {@link ChunkUtils} to chunk the keys
+   * and execute query to avoid 413 error.
    *
    * @param keys keys to cache.
    * @param keyMapper a function to get the key from the resource.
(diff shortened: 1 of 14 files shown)


src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -59,6 +62,12 @@
 
   private static final int MAXIMUM_ALLOWED_UPDATE_ACTIONS = 500;
   static final String CREATE_FAILED = "Failed to create draft with key: '%s'. Reason: %s";
+  /*
+   * A key is a 41 characters long hashed string. (i.e: c25a67e262265fbc36edb33ed0375263586be278) We
+   * chunk them in 250 keys we will have around a query around 11.000 characters. Above this size it
+   * could return - Error 413 (Request Entity Too Large)
+   */
+  static final int CHUNK_SIZE = 250;
```
butenkor, line 70: Maybe we should be more conservative here as AFAIK the request length is calculated not just from the 250 keys but also all of the provided headers, encoded parameters and request URL.

praveenkumarct: Yes @butenkor , But as mentioned in the PR comment already, I wasn't sure how to calculate the key size here. (I see this methods in BaseService uses Key and Sku values. When I was creating ProductType, It says the below message - 
 "May only contain between 2 and 256 alphanumeric characters, underscores, or hyphens (no white space or special characters like ñ, ü, #, %) ". 
 But I don't see any restrictions for "key" values while creating Products and Categories.

lojzatran (changes requested):

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -59,6 +62,12 @@
 
   private static final int MAXIMUM_ALLOWED_UPDATE_ACTIONS = 500;
   static final String CREATE_FAILED = "Failed to create draft with key: '%s'. Reason: %s";
+  /*
+   * A key is a 41 characters long hashed string. (i.e: c25a67e262265fbc36edb33ed0375263586be278) We
```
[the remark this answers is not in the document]
praveenkumarct: I am really not sure, What is the key size here. So I am asking your suggestions in the PR comment.

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -268,7 +286,8 @@
 
   /**
    * Given a set of keys this method caches a mapping of the keys to ids of such keys only for the
-   * keys which are not already in the cache.
+   * keys which are not already in the cache. This method uses {@link ChunkUtils} to chunk the keys
```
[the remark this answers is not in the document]
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/688/commits/c5b7afdeae89cf7ec9781a0d9fda6718c400d009

src/main/java/com/commercetools/sync/services/impl/BaseService.java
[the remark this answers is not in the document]
lojzatran: Since it's not clear in size, you can either count it as max length 256 chars. This will be safe.

As to how to calculate the exact length, check this: https://github.com/sphereio/sphere-product-import/blob/841957349815683d8825855fccaca07a680b894f/lib/common-utils.coffee#L25. As Roman suggested, it might not include headers, so you need to leave space for headers too.

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -288,12 +305,20 @@
       return CompletableFuture.completedFuture(keyToIdCache.asMap());
     }
 
-    final Consumer<Set<ResourceKeyId>> resultConsumer =
-        results -> results.forEach(resource -> keyToIdCache.putAll(keyToIdMapper.apply(resource)));
+    final List<List<String>> chunkedKeys = ChunkUtils.chunk(keysNotCached, CHUNK_SIZE);
```
lojzatran, line 308: Do we need to chunk for graphql? It's sent in body as POST, so I don't think it's a problem here.

praveenkumarct: As discussed, Yes we will need to do it for graphql.

src/main/java/com/commercetools/sync/services/impl/BaseService.java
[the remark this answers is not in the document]
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/688/commits/6c5a1d96da9ff116e189543315e418cfcca67813

src/main/java/com/commercetools/sync/services/impl/BaseService.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/688/commits/6c5a1d96da9ff116e189543315e418cfcca67813.

src/main/java/com/commercetools/sync/services/impl/BaseService.java
```diff
@@ -59,6 +62,12 @@
 
   private static final int MAXIMUM_ALLOWED_UPDATE_ACTIONS = 500;
   static final String CREATE_FAILED = "Failed to create draft with key: '%s'. Reason: %s";
+  /*
+   * Considering maximum of 256 characters for key and sku fields (key and sku field doesn't have
+   * limit except for ProductType(256)) We chunk them in 60 (keys or sku) we will have around a query
+   * around 15.000 characters. Above this size it could return - Error 414 (Request-URI Too Large)
```
lojzatran, line 68: Where did you get 414? I also think I saw 413, but I'm not sure. Can you clarify?

lojzatran: @butenkor Maybe you know if it was HTTP status 413 or 414? In the docs, I could not find anything.

praveenkumarct: http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers. You can check here for both 413 and 414. We are getting 414, Since it is a get call.

lojzatran: I tested it, and it returns 414, so you're right.

lojzatran (approved):

praveenkumarct:
It doesn't seem reasonable to use 256 characters for key and sku fields. So, we will be considering 41 characters and chunk them accordingly(also considering some conservative space for headers).

PR #690, opened 2021-03-16
Fix benchmark gradle script.

ahmetoz:
The build is failing with the error message below on benchmark step, See: https://github.com/commercetools/commercetools-sync-java/actions/runs/656992470

````
Script '/home/runner/work/commercetools-sync-java/commercetools-sync-java/gradle-scripts/benchmark.gradle' line: 19
* What went wrong:
Execution failed for task ':benchmark'.
> Creating directory /home/runner/.gradle/daemon/5.6.2/tmp_git_dir/.git/refs failed
````

The changes is trying to attempt fix issues on the benchmark script, additionally the gradle is upgraded to never version.

### CODE REVIEW

**Comment by ahmetoz** [comment]
Benchmark issue is resolved with the commit: https://github.com/commercetools/commercetools-sync-java/pull/690/commits/3c723e6274b03636d32bf4eec595ec8a46cf24c9
Build example: https://github.com/commercetools/commercetools-sync-java/runs/2148819942

File: src/main/java/com/commercetools/sync/services/impl/BaseService.java [hunk 1]
```diff
@@ -348,7 +348,12 @@
 
     List<Q> keysQueryMapperList =
         chunkedKeys.stream()
-            .map(_keys -> keysQueryMapper.apply(new HashSet<>(_keys)))
+            .map(
+                _keys ->
+                    keysQueryMapper
+                        .apply(new HashSet<>(_keys))
+                        .withLimit(_keys.size())
+                        .withFetchTotal(false))
```
praveenkumarct, line 356: Do you see any advantage adding this 2 -> .withLimit and withFetchTotal ?
If yes, Can you add the same to cacheKeysToIds and cacheKeysToIdsUsingGraphQl methods as well? Or may be refactor to a method and use them. WDYT?

ahmetoz: I will check other parts, withFetchTotal is for performance. withLimit is necessary otherwise it would return only with the default limit and some of them would be missed to sync :) like in benchmarks

ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/690/commits/3346ad0658a56af69876501ca48c9f606919f7aa

praveenkumarct (approved):

PR #697, opened 2021-03-18
Resolve References using cache for Product resource.

praveenkumarct:
- Create a service which accepts client and map(cache) and products to transform and map.
- update utils to perform reference resolution using cache.
- Unit and Integration tests.
- Proper Javadocs.


src/main/java/com/commercetools/sync/products/service/ProductReferenceTransformService.java
```diff
@@ -0,0 +1,31 @@
...
+public interface ProductReferenceTransformService {
+
+  /**
+   * Transforms products by resolving the references and map them to ProductDrafts.
+   *
+   * <p>This method replaces the ids on attribute references with keys and resolves(fetch key values
+   * for the reference id's) non null and unexpanded references of the product{@link Product} by
+   * using cache.
+   *
+   * <p>If the reference id's are already cached, key values are fetched from the cache else it
```
ahmetoz, line 18:
```suggestion
   * <p>If the reference ids are already cached, key values are pulled from the cache, otherwise it
```

src/main/java/com/commercetools/sync/products/service/ProductReferenceTransformService.java
```diff
@@ -0,0 +1,31 @@
...
+   * executes the query to fetch the key value for the reference id's and store the idToKey value
+   * pair in the cache for reuse.
+   *
+   * <p>Also maps the Product to ProductDraft by performing reference resolution considering idToKey
```
ahmetoz, line 22:
```suggestion
   * <p>Then maps the Product to ProductDraft by performing reference resolution considering idToKey
```

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
```diff
@@ -0,0 +1,496 @@
...
+import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProductReferenceTransformServiceImpl extends BaseTransformServiceImpl
+    implements ProductReferenceTransformService {
+
+  private static final Logger LOGGER =
```
ahmetoz, line 52: We don't have a logger in the java-sync 🤔

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
```diff
@@ -0,0 +1,496 @@
...
+  @Nonnull
+  @Override
+  public CompletableFuture<List<ProductDraft>> transformProductReferences(
+      @Nonnull final List<Product> products) {
+
+    return replaceAttributeReferenceIdsWithKeys(products)
+        .handle(
+            (productsResolved, throwable) -> {
+              if (throwable != null && LOGGER.isWarnEnabled()) {
+                LOGGER.warn(
```
ahmetoz, line 75: Looks like this error handling part is more related to the project-sync code, not related to java-sync . In this case, we might throw the error from the service as we could not resolve this right? The library user should decide what to do.

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
```diff
@@ -0,0 +1,496 @@
+package com.commercetools.sync.products.service.impl;
```
ahmetoz, line 1: What do you think about moving the services to the "com.commercetools.sync.services.transform" package?

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
```diff
@@ -0,0 +1,496 @@
...
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(ProductReferenceTransformServiceImpl.class);
+  private static final String WITH_IRRESOLVABLE_REFS_ERROR_MSG =
+      "The product with id '%s' on the source project ('%s') will "
+          + "not be synced because it has the following reference attribute(s): %n"
+          + "%s.%nThese references are either pointing to a non-existent resource or to an existing one but with a blank key. "
+          + "Please make sure these referenced resources are existing and have non-blank (i.e. non-null and non-empty) keys.";
+
+  public ProductReferenceTransformServiceImpl(
```
ahmetoz, line 60: What about providing a default map backed by caffeeine as a reference? So the library users would use it without investing in their best practices?

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
```diff
@@ -0,0 +1,496 @@
...
+
+    if (nonCachedCustomObjectIds.isEmpty()) {
+      return CompletableFuture.completedFuture(referenceIdToKeyCache);
+    }
+
+    final List<List<String>> chunkedIds = ChunkUtils.chunk(nonCachedCustomObjectIds, CHUNK_SIZE);
+
+    final List<CustomObjectQuery<JsonNode>> chunkedRequests =
+        chunkedIds.stream()
+            .map(ids -> CustomObjectQuery.ofJsonNode().plusPredicates(p -> p.id().isIn(ids)))
```
ahmetoz, line 480: can you put withTotal=false and a limit to the query, otherwise it's limited to 20 custom objects, (we figured that with benchmark test in the previous release if you remember)

src/main/java/com/commercetools/sync/products/utils/ProductReferenceResolutionUtils.java
```diff
@@ -221,56 +225,13 @@ static CategoryReferencePair mapToCategoryReferencePair(@Nonnull final Product p
...
-            expansionModel ->
-                expansionModel.masterData().staged().allVariants().attributes().value())
-        .plusExpansionPaths(
-            expansionModel ->
-                expansionModel.masterData().staged().allVariants().attributes().valueSet())
-        .plusExpansionPaths(
-            ExpansionPath.of("masterData.staged.masterVariant.assets[*].custom.type"))
-        .plusExpansionPaths(
-            ExpansionPath.of("masterData.staged.variants[*].assets[*].custom.type"));
+    return ProductQuery.of();
```
ahmetoz, line 234: That would conflict with Andi's work, please consider not merging those changes into master until the whole development is done.

src/main/java/com/commercetools/sync/services/impl/BaseTransformServiceImpl.java
```diff
@@ -0,0 +1,95 @@
...
+      @Nonnull final Map<String, String> referenceIdToKeyCache) {
+    this.ctpClient = ctpClient;
+    this.referenceIdToKeyCache = referenceIdToKeyCache;
+  }
+
+  protected SphereClient getCtpClient() {
+    return ctpClient;
+  }
+
+  //  @SuppressFBWarnings(
```
ahmetoz, line 43: Is this comment required?

ahmetoz (changes requested):

src/main/java/com/commercetools/sync/products/utils/ProductReferenceResolutionUtils.java
praveenkumarct: Sure, I will wait for Andi to merge his changes into master and then i will fix the conflicts in my branch.

src/main/java/com/commercetools/sync/services/impl/BaseTransformServiceImpl.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/db6bca046d69046ac3d92b2b3bac89e65ac5f1f4

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/c5e5c9b9c226f383f5d8ba70db7357c7bfdd0d8d

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
praveenkumarct: Since the class name already contains "Transform" and also we don't have any other service classes as of now. I think it is good to have in com.commercetools.sync.service package itself. Let me know if you think still it is better to have transform package!

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
praveenkumarct: Do you mean to provide the default map when the map is null ?  Or just update the javadocs about the default cache Caffeine for library users to implement it with example?

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/c15173a3fe3479816c7abb1d34ecc09a498b8f0e

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/c15173a3fe3479816c7abb1d34ecc09a498b8f0e

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
ahmetoz: I'm asking giving a reference class for to use by default on our snippets, also document for custom use cases, and invalidation scenarios for our clients (in the doc snippets)

For instance we would have a class called `DefaultTransformServiceCache` (or you name it) which is documented as an example in the docs and used by integration tests.

Class is backed by caffeine cache with LRU cache eviction strategy, which is taking the size and documents the best practices for the caching needs. It could also have additional utility methods like invalidateAll method, or others.

ahmetoz (changes requested):

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/8a458e56c6e4a63b346644e26f27b6cf44057a5d
I have added Default cache using Caffeine and this will be documented as part of documentation task.

praveenkumarct:
@ahmetoz 
When I used a Default cache(Caffeine), observed that it doesn't allow null values into the map.
Because of this, Integration tests failed for the case(where ProductType key value is null). To handle this case, I have introduced a new static variable (BLANK_KEY_VALUE = "KEY_VALUE_EITHER_NULL_OR_BLANK") in BaseTransformServiceImpl.
Let me know your view on this? Do you have any other clean way of handling this scenario?

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
ahmetoz: First it's not something related to product package, maybe this should be part of the utils, also instead of explaining it, I pushed a commit to another branch, which I meant in my comment: https://github.com/commercetools/commercetools-sync-java/commit/c258f2cbe171378379d6fbc0ef44659a84f64d54

praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/840a2dea62f16b192af58dbc402883551f0a39f5 merged into this branch using cherry-pick

src/main/java/com/commercetools/sync/products/utils/ProductReferenceResolutionUtils.java
ahmetoz: you might base to `release_5` branch, that Andi created already, so we could review this again, as there would be lots of conflict changes

src/main/java/com/commercetools/sync/services/impl/BaseTransformServiceImpl.java
```diff
@@ -0,0 +1,102 @@
...
+import javax.annotation.Nonnull;
+
+public class BaseTransformServiceImpl {
+  /*
+   * An id is a 36 characters long string. (i.e: 53c4a8b4-754f-4b95-b6f2-3e1e70e3d0c3) We
+   * chunk them in 300 ids, we will have a query around 11.000 characters. Above this size it
+   * could return - Error 413 (Request Entity Too Large)
+   */
+  public static final int CHUNK_SIZE = 300;
+  public static final String BLANK_KEY_VALUE = "KEY_VALUE_EITHER_NULL_OR_BLANK";
```
ahmetoz, line 31:
```suggestion
  public static final String KEY_IS_NOT_SET_PLACE_HOLDER = "KEY_IS_NOT_SET";
```

src/main/java/com/commercetools/sync/services/impl/BaseTransformServiceImpl.java
```diff
@@ -0,0 +1,102 @@
...
+      @Nonnull final List<List<String>> chunkedIds,
+      @Nonnull final GraphQlQueryResources resourceType) {
+    return chunkedIds.stream()
+        .map(chunk -> new ResourceIdsGraphQlRequest(new HashSet<>(chunk), resourceType))
+        .collect(toList());
+  }
+
+  @Nonnull
+  protected Set<String> getNonCachedReferenceIds(@Nonnull final Set<String> referenceIds) {
+    return referenceIds.stream()
```
ahmetoz, line 84: Here we need to filter out the keys not exist but have a placeholder, KEY_IS_NOT_SET_PLACE_HOLDER. 
Otherwise, when the key is added to the project it would not be synced and the exceptions will not be resolved.

Could you check for other code if `getNonCachedReferenceIds` is overridden in subclasses? Also please try to check document reference for custom objects, as it's not stored as a simple key but the combination of container and key (("id", "co-container|co-key"))

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
```diff
@@ -0,0 +1,473 @@
...
+
+    final Set<String> nonCachedProductIds = getNonCachedReferenceIds(productIds);
+    final Set<String> nonCachedCategoryIds = getNonCachedReferenceIds(categoryIds);
+    final Set<String> nonCachedProductTypeIds = getNonCachedReferenceIds(productTypeIds);
+    final Set<String> nonCachedCustomObjectIds = getNonCachedReferenceIds(customObjectIds);
+
+    if (nonCachedProductIds.isEmpty()
+        && nonCachedCategoryIds.isEmpty()
+        && nonCachedProductTypeIds.isEmpty()) {
+      return fetchCustomObjectKeys(customObjectIds);
```
ahmetoz, line 414: Looks like, it's fetching the keys that is already in the cache: 

```suggestion
      return fetchCustomObjectKeys(nonCachedCustomObjectIds);
```

src/main/java/com/commercetools/sync/products/service/impl/ProductReferenceTransformServiceImpl.java
```diff
@@ -0,0 +1,473 @@
...
+
+    return ChunkUtils.executeChunks(getCtpClient(), chunkedRequests)
+        .thenApply(ChunkUtils::flattenPagedQueryResults)
+        .thenApply(
+            customObjects -> {
+              customObjects.forEach(
+                  customObject -> {
+                    referenceIdToKeyCache.put(
+                        customObject.getId(),
+                        CustomObjectCompositeIdentifier.of(customObject).toString());
```
ahmetoz, line 468: I checked if this would be needed a placeholder here, but in commercetools, both container and key are required fields, looks like we are safe in this case, so looks like no additional check is needed.

src/main/java/com/commercetools/sync/services/impl/BaseTransformServiceImpl.java
```diff
@@ -0,0 +1,102 @@
...
+        .filter(id -> !referenceIdToKeyCache.containsKey(id))
+        .collect(toSet());
+  }
+
+  protected void cacheResourceReferenceKeys(final Set<ResourceKeyId> results) {
+    Optional.ofNullable(results)
+        .orElseGet(Collections::emptySet)
+        .forEach(
+            resourceKeyId -> {
+              final String key = resourceKeyId.getKey();
```
ahmetoz, line 94: Here as you were doing before, you could keep the placeholder as a key if it could be fetched from CTP, then during reference resolution, it could use the same error handling, which is also failing fast to error callback before executing actions.

ahmetoz (commented):
As requested, I added my comments considering not set keys on the target project.

src/main/java/com/commercetools/sync/products/utils/ProductReferenceResolutionUtils.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/4890668f7c7cf19bfab7e7ed41860fdbc8a67b3f

src/main/java/com/commercetools/sync/services/impl/BaseTransformServiceImpl.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/4c3892041cb3afa8f9a6167c31647b69aebf75cf

src/main/java/com/commercetools/sync/services/impl/BaseTransformServiceImpl.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/697/commits/ffe6f19db27aa206410807c5e56943283370b0c3
I am checking for custom Objects.

praveenkumarct:
> it will be always not blank, so this check might be unnecessary.

https://github.com/commercetools/commercetools-sync-java/pull/697/commits/ffe6f19db27aa206410807c5e56943283370b0c3

ahmetoz (dismissed):

PR #702, opened 2021-03-26
Cache references on shoppinglists

praveenkumarct:
- Create a service which accepts client and map(cache) and shoppingLists to transform and map.
- update utils to perform reference resolution using cache.
- Unit and Integration tests.
- Proper Javadocs.


src/integration-test/java/com/commercetools/sync/integration/services/impl/StateServiceImplIT.java
```diff
@@ -161,6 +161,7 @@ void fetchMatchingStatesByKeys_WithAnyExistingKeys_ShouldReturnASetOfStates() {
     assertThat(transition.getId()).isEqualTo(transitionState.getId());
     assertThat(transition.getObj()).isNull();
 
+    clearTransitions(CTP_TARGET_CLIENT, fetchState);
```
salander85, line 164: Please check if this line is necessary, came with the Merge.

salander85 (commented):
LGTM

src/integration-test/java/com/commercetools/sync/integration/services/impl/StateServiceImplIT.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/702/commits/40a1359c55894abc96c98ec084a7b2f9b4268c9b
Thanks for noticing this. I had to do workaround to fix the tests. But after Andi's changes, this is resolved. So, reverted back to the old code.

src/main/java/com/commercetools/sync/shoppinglists/utils/ShoppingListReferenceResolutionUtils.java
```diff
@@ -177,56 +188,49 @@ private static LineItemDraft mapToLineItemDraft(@Nonnull final LineItem lineItem
...
+   *     expanded except for Variants of the LineItems.
    */
   public static ShoppingListQuery buildShoppingListQuery() {
-    return ShoppingListQuery.of()
-        .withExpansionPaths(ShoppingListExpansionModel::customer)
-        .plusExpansionPaths(ExpansionPath.of("custom.type"))
-        .plusExpansionPaths(ExpansionPath.of("lineItems[*].variant"))
-        .plusExpansionPaths(ExpansionPath.of("lineItems[*].custom.type"))
-        .plusExpansionPaths(ExpansionPath.of("textLineItems[*].custom.type"));
+    return ShoppingListQuery.of().plusExpansionPaths(ExpansionPath.of("lineItems[*].variant"));
```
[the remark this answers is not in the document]
praveenkumarct: Yes, Because for variant's we are fetching SKU and quantity values but not the key values. And In service class also we are using Variant expansion in this case. So, discussed this already with @ahmetoz and I left variant to expand in the query for now.

leungkinghin (approved):

PR #705, opened 2021-03-26
Explizit delete transactions before deleting states.

ahalberkamp:
In this pr the "deleteStates"  method was refactored so that all transitions are explicitly deleted before the states were deleted. Otherwise, it could be that some states cannot be deleted, because some transitions are still cached.

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/commons/utils/StateITUtils.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/commons/utils/StateITUtils.java
@@ -74,8 +74,8 @@ public final class StateITUtils {
     // delete transitions
     CtpQueryUtils.queryAll(ctpClient, stateQueryBuilder.build(), Function.identity())
         .thenApply(
-            fetchedCategories ->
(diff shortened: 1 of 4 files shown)


src/integration-test/java/com/commercetools/sync/integration/commons/utils/StateITUtils.java
```diff
@@ -52,24 +52,30 @@ private StateITUtils() {}
...
-    CtpQueryUtils.queryAll(
-            ctpClient,
-            StateQueryBuilder.of().plusPredicates(QueryPredicate.of("builtIn = false")).build(),
-            Function.identity())
+    CtpQueryUtils.queryAll(ctpClient, stateQueryBuilder.build(), Function.identity())
         .thenApply(
-            fetchedCategories ->
-                fetchedCategories.stream().flatMap(List::stream).collect(Collectors.toList()))
+            fetchedStates ->
+                fetchedStates.stream().flatMap(List::stream).collect(Collectors.toList()))
```
ahalberkamp, line 78: 91bb5ff62

ahmetoz (approved):
Thanks for the boy-scout update @ahalberkamp

PR #720, opened 2021-04-15
Support syncing Set of (Set of Set of..) of NestedType

ahmetoz:
**Situation:**
Syncing product types with an attribute of type Set of NestedType attribute is supported. However, Set of (Set of Set of..) of NestedType is not yet supported.

**Complication:**
As it was an expensive query for the commercetools platform to expand that type of set chains before we decided to limit it. Now with using cache service instead of expansion, we could resolve this limitation/caveat on the product type sync, so we could provide users fully functional sync for the SetAttributeType chains. 

## TODOs:
- [x] Add release notes for the changes, remove caveat and increment the versions.

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/commons/utils/ProductTypeITUtils.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/commons/utils/ProductTypeITUtils.java
@@ -29,6 +29,7 @@ import io.sphere.sdk.products.attributes.EnumAttributeType;
 import io.sphere.sdk.products.attributes.LocalizedEnumAttributeType;
 import io.sphere.sdk.products.attributes.LocalizedStringAttributeType;
 import io.sphere.sdk.products.attributes.NestedAttributeType;
+import io.sphere.sdk.products.attributes.SetAttributeType;
 import io.sphere.sdk.products.attributes.StringAttributeType;
 import io.sphere.sdk.producttypes.ProductType;
 import io.sphere.sdk.producttypes.ProductTypeDraft;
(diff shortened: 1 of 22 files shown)


src/main/java/com/commercetools/sync/producttypes/helpers/AttributeDefinitionReferenceResolver.java
```diff
@@ -83,16 +85,54 @@ public AttributeDefinitionReferenceResolver(
...
+      if (nestedAttributeType instanceof NestedAttributeType) {
+        return resolveNestedAttributeTypeReferences(
+            attributeDefinitionDraftBuilder, maxDepth, (NestedAttributeType) nestedAttributeType);
       }
     }
     return completedFuture(attributeDefinitionDraftBuilder);
   }
 
+  @Nonnull
+  private CompletionStage<AttributeDefinitionDraftBuilder> resolveNestedAttributeTypeReferences(
```
salander85, line 104: A little comment would help as this is pretty complex.

src/integration-test/java/com/commercetools/sync/integration/commons/utils/ProductTypeITUtils.java
```diff
@@ -61,16 +62,19 @@
   public static final String PRODUCT_TYPE_KEY_2 = "key_2";
   public static final String PRODUCT_TYPE_KEY_3 = "key_3";
   public static final String PRODUCT_TYPE_KEY_4 = "key_4";
+  public static final String PRODUCT_TYPE_KEY_5 = "key_5";
```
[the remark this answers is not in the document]
ahmetoz: Maybe not in ProductTypeITUtils but used in other test classes 😄 If not used, mostly it would fail also in the code checks.

src/main/java/com/commercetools/sync/producttypes/helpers/AttributeDefinitionReferenceResolver.java
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/720/commits/4eaf98b9f722d6c2b4614192dadd9a29b6037ee3

salander85 (approved):

PR #726, opened 2021-05-04
Special character fix product sync

praveenkumarct:
# Description:

The PR is to fix the potential error while syncing resources with key value containing special characters, This is observed by the [issue](https://github.com/commercetools/commercetools-project-sync/issues/269). 

#Review notes:
- For `Product` we are building where query manually using string format. check [productserviceImpl](https://github.com/commercetools/commercetools-sync-java/pull/726/files#diff-2f984d1bd1857a09ee354220da81e60e2fc9c40af5f08d44aa5334a6f8f990c9)
- And for GraphQlquery, also We are manually building the where query for search based on Key. So have fixed it.


src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
```diff
@@ -104,5 +106,17 @@ public static boolean isUuid(@Nonnull final String id) {
     return regexPattern.matcher(id).matches();
   }
 
+  /**
+   * @param key to check if it contains special character(") and format it accordingly.
+   * @return formatted key value.
+   */
+  public static String formatKeyWithEscapeCharacter(final String key) {
```
JudeNiroshan, line 113: do we really need a separate function for this? Shouldn't it be part of the service class?

JudeNiroshan (changes requested):

src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
```diff
@@ -104,5 +106,17 @@ public static boolean isUuid(@Nonnull final String id) {
     return regexPattern.matcher(id).matches();
   }
 
+  /**
+   * @param key to check if it contains special character(") and format it accordingly.
```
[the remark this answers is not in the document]
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/726/commits/0605901c73598397a4a011b24dfab34cdc594783

src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
praveenkumarct: No more required. I have refactored as part your previous review comment. https://github.com/commercetools/commercetools-sync-java/pull/726/commits/0605901c73598397a4a011b24dfab34cdc594783

src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
```diff
@@ -11,6 +11,8 @@
 import javax.annotation.Nullable;
 
 public final class SyncUtils {
+  private static final String DOUBLE_QUOTE = "\"";
+  private static final String DOUBLE_QUOTE_WITH_ESCAPE_CHARACTERS = "\\\\\"";
```
butenkor, line 15: Why do we need this and the one above?

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountReferenceResolutionUtils.java
```diff
@@ -73,7 +74,7 @@ private static CartDiscountDraft mapToCartDiscountDraft(
             cartDiscount.getSortOrder(),
             cartDiscount.getTarget(),
             cartDiscount.getValue())
-        .key(cartDiscount.getKey())
+        .key(StringEscapeUtils.escapeJava(cartDiscount.getKey()))
```
butenkor, line 77: Why do you escape in the draft? Draft should be created with all the special characters as provided and we want to escape on queries only like [here](https://github.com/commercetools/commercetools-sync-java/blob/c0cc6c41768bcde703616a1dbde4e2330d3df3c0/src/main/java/com/commercetools/sync/types/TypeSync.java#L109), right? Same applies for other resources.

src/main/java/com/commercetools/sync/commons/utils/SyncUtils.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/726/commits/672a324a4c2e4044fd1c098d08238ec10860c5e3

src/main/java/com/commercetools/sync/cartdiscounts/utils/CartDiscountReferenceResolutionUtils.java
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/726/commits/cd1968be7dccb8c7d8c34f49a21866d61b80cdb5

JudeNiroshan (approved):

PR #727, opened 2021-05-07
release preparation v5.1.1

praveenkumarct:
v5.1.1 release preparation.

Hint: Sorry for the confusion on branch name(v5.2.0).

### DIFF
--- a/docs/RELEASE_NOTES.md
+++ b/docs/RELEASE_NOTES.md
@@ -27,6 +27,15 @@
 7. Add Migration guide section which specifies explicitly if there are breaking changes and how to tackle them.
 -->
 
+### 5.1.1 - May 7, 2021
+[Commits](https://github.com/commercetools/commercetools-sync-java/compare/5.1.0...5.1.1) |
+[Javadoc](https://commercetools.github.io/commercetools-sync-java/v/5.1.1/) |
+[Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/5.1.1)
+
+- 🐞 **Bug Fixes** (1)
+    - **Product Sync** - Fixed a bug in the `ProductSync` related to querying of productProjections based on key values, when key value contains special characters(") [#269](https://github.com/commercetools/commercetools-project-sync/issues/269)
+    - Same fix also done for GraphqlQuery search based on keys.
+
 ### 5.1.0 - Apr 20, 2021
 [Commits](https://github.com/commercetools/commercetools-sync-java/compare/5.0.0...5.1.0) |
 [Javadoc](https://commercetools.github.io/commercetools-sync-java/v/5.1.0/) |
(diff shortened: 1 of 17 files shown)


docs/RELEASE_NOTES.md
```diff
@@ -27,6 +27,15 @@
 7. Add Migration guide section which specifies explicitly if there are breaking changes and how to tackle them.
 -->
 
+### 5.1.1 - May 7, 2021
+[Commits](https://github.com/commercetools/commercetools-sync-java/compare/5.1.0...5.1.1) |
+[Javadoc](https://commercetools.github.io/commercetools-sync-java/v/5.1.1/) |
+[Jar](https://bintray.com/commercetools/maven/commercetools-sync-java/5.1.1)
+
+- 🐞 **Bug Fixes** (1)
+    - **Product Sync** - Fixed a bug in the `ProductSync` related to querying of productProjections based on key values, when key value contains special characters(") [#269](https://github.com/commercetools/commercetools-project-sync/issues/269)
```
JudeNiroshan, line 36:
```suggestion
    - **Product Sync** - Special characters can be defined for ProductDraft key. [#269](https://github.com/commercetools/commercetools-project-sync/issues/269)
```

docs/RELEASE_NOTES.md
```diff
@@ -27,6 +27,15 @@
...
+    - Same fix also done for GraphqlQuery search based on keys.
```
JudeNiroshan, line 37: This is not a separate issue we fixed in this release. So, it would be better to include this point to the previous sentence or remove this point as it is obvious.

JudeNiroshan (changes requested):

docs/RELEASE_NOTES.md
praveenkumarct: It was always allowed to define special characters in the ProductDraft key. We didn't change this. We fixed the code where we are building query for the search based on Key's.
> Special characters can be defined for ProductDraft key.
May be this will mis guide?

JudeNiroshan: ok, in that case, please make sure the mention the end result for the library user. If you mention the internal bug fixes, in the release note, you should always describe those fixes who the end-user can be benefited.

praveenkumarct: > please make sure to mention the end result for the library user
Then your statement makes sense for me. It explains users to have special characters in the productDraft key.

docs/RELEASE_NOTES.md
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/727/commits/f9dc9631f8a1221bad15ea5f2b4586c2ce621df2 This is also related to ProductDraft key value, So removed it.

JudeNiroshan (approved):

PR #738, opened 2021-05-31
Remove Irresolvable references check in Product transformer service.

praveenkumarct:
Remove Irresolvable references check from transformer service.

Hint for review:
Irresolvable references should not be skipped in the transformer service. Because they are handled in the later part of the code and informing user about the irresolvable references. Check [here](https://github.com/commercetools/commercetools-sync-java/blob/700e3c03d7b26eb544846be494abf3fa6f9b3446/src/main/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImpl.java#L56).


leungkinghin (approved):
It looks good to me

PR #750, opened 2021-07-05
ProductProjection - Use the new predicate to filter and search based on key.

praveenkumarct:
- Remove the manual construction of query for fetching product projections based on key and use key predicate provided by the new JVM-SDK version.

#### Hints for Review
- Tests are already covered, Check this PR - https://github.com/commercetools/commercetools-sync-java/pull/726/

### DIFF
--- a/src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
+++ b/src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
@@ -2,8 +2,6 @@ package com.commercetools.sync.services.impl;
 
 import static com.commercetools.sync.commons.utils.SyncUtils.batchElements;
 import static io.sphere.sdk.products.ProductProjectionType.STAGED;
-import static java.lang.String.format;
-import static java.util.Collections.singleton;
 
 import com.commercetools.sync.commons.helpers.ResourceKeyIdGraphQlRequest;
 import com.commercetools.sync.commons.models.GraphQlQueryResources;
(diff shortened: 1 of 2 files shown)


src/test/java/com/commercetools/sync/services/impl/ProductServiceTest.java
```diff
@@ -138,33 +136,4 @@ void updateProduct_WithMockCtpResponse_ShouldReturnMock() {
     verify(productSyncOptions.getCtpClient())
         .execute(eq(ProductUpdateCommand.of(productProjectionMock, updateActions)));
   }
-
-  @Test
-  void buildProductKeysQueryPredicate_WithEmptyProductKeysSet_ShouldBuildCorrectQuery() {
```
ahmetoz, line 143: Not sure but, when we refactor the code, is not it better to refactor test instead of dropping them, is there an equivalent of that those queries are correct now ?

src/main/java/com/commercetools/sync/services/impl/ProductServiceImpl.java
```diff
@@ -70,23 +64,6 @@ public ProductServiceImpl(@Nonnull final ProductSyncOptions syncOptions) {
             new ResourceKeyIdGraphQlRequest(keysNotCached, GraphQlQueryResources.PRODUCTS));
   }
 
-  // TODO This code should be removed and we should use StringQuerySortingModel(It already handles
```
ahmetoz, line 73: goos to see how garbage is removed from the code :)

src/test/java/com/commercetools/sync/services/impl/ProductServiceTest.java
praveenkumarct: This tests are completely for the deleted code and yes, for the method I have already added tests in my previous PR.
Check the description.

ahmetoz (approved):

PR #753, opened 2021-07-07
Change the logic for TaxRates update actions.

praveenkumarct:
- **TaxCategory Sync**: This PR is to fix the Sync issue, when we have TaxRates with and without state.
- Linked Issue - https://github.com/commercetools/commercetools-project-sync/issues/298

#### Hints for Review
- To cover all the cases, now we are removing and adding taxRates, if the `state` value is changed.

### DIFF
--- a/src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
@@ -111,9 +111,10 @@ final class TaxRatesUpdateActionUtils {
             oldTaxRate ->
                 newTaxRatesDrafts.stream()
                     .filter(
-                        taxRateDraft ->
-                            Objects.equals(oldTaxRate.getCountry(), taxRateDraft.getCountry())
-                                && Objects.equals(oldTaxRate.getState(), taxRateDraft.getState()))
+                        newTaxRateDraft ->
+                            Objects.equals(oldTaxRate.getCountry(), newTaxRateDraft.getCountry())
+                                && Objects.equals(
+                                    oldTaxRate.getState(), newTaxRateDraft.getState()))
                     .findFirst()
                     .map(
                         matchedTaxRateDraft -> {
--- a/src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
@@ -112,13 +112,8 @@ final class TaxRatesUpdateActionUtils {
                 newTaxRatesDrafts.stream()
                     .filter(
                         taxRateDraft ->
-                            Objects.equals(oldTaxRate.getCountry(), taxRateDraft.getCountry()))
-                    .filter(
-                        taxRateDraft ->
-                            oldTaxRate.getState() == null
-                                || (oldTaxRate.getState() != null
-                                    && taxRateDraft.getState() == null)
-                                || Objects.equals(oldTaxRate.getState(), taxRateDraft.getState()))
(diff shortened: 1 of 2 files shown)


src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
```diff
@@ -112,13 +112,8 @@ private TaxRatesUpdateActionUtils() {}
...
-                        taxRateDraft ->
-                            oldTaxRate.getState() == null
-                                || (oldTaxRate.getState() != null
-                                    && taxRateDraft.getState() == null)
-                                || Objects.equals(oldTaxRate.getState(), taxRateDraft.getState()))
+                            Objects.equals(oldTaxRate.getCountry(), taxRateDraft.getCountry())
+                                && Objects.equals(oldTaxRate.getState(), taxRateDraft.getState()))
                     .findFirst()
                     .map(
                         matchedTaxRateDraft -> {
```
butenkor, line 119:
```
newTaxRatesDrafts.remove(matchedTaxRateDraft);
```
Why do we need to remove here? Also, if we have to it seem not the best practice to modify the list we iterate over since you might have unexpected results or ConcurrentModificationException. 
https://blog.jooq.org/2014/06/13/java-8-friday-10-subtle-mistakes-when-using-the-streams-api/ (check point 8)

src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
```diff
@@ -112,13 +112,8 @@ private TaxRatesUpdateActionUtils() {}
                 newTaxRatesDrafts.stream()
                     .filter(
                         taxRateDraft ->
```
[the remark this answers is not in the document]
praveenkumarct: https://github.com/commercetools/commercetools-sync-java/pull/753/commits/9b796dce2d3136aabf9b7713cbc289d514a5964e

src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
praveenkumarct: @butenkor Yes, you are right we should not remove elements from collection, when we iterate through it. But in this case I don't see any problem because we are no more continuing the loop after this line(remove element from collection). We are returning back from the current loop and we will iterate with a new collection(without removed element).
> newTaxRatesDrafts.remove(matchedTaxRateDraft);
                            return singletonList(
                                ReplaceTaxRate.of(oldTaxRate.getId(), matchedTaxRateDraft));
 

And I see this is old code, we are removing elements to reduce the number of iterations every time on filter.

leungkinghin (approved):
Looks good to me. Please address Roman's feedback later on.

src/main/java/com/commercetools/sync/taxcategories/utils/TaxRatesUpdateActionUtils.java
leungkinghin: Perhaps we can clone `newTaxRatesDrafts` in line 112 and iterate it, and keep the matched draft removal. It makes the code snippet look less complicated.

butenkor: > But in this case I don't see any problem because we are no more continuing the loop after this line(remove element from collection)

Since it is a bad practice you do not know which surprises it can hold for you. For example our [callbacks](https://github.com/commercetools/commercetools-sync-java/blob/1c03af337d807fce00187715705da4074cad4299/src/main/java/com/commercetools/sync/inventories/InventorySyncOptions.java#L24-L43) provide on different events a draft representation for logging/debugging purposes. Let's assume we remove all the taxRateDrafts from the original draft above, hit API error and call errorCallback - what will the user see in the taxRateDrafts array?

PR #757, opened 2021-07-19
Bug fix for multi-channel inventory sync.

ahmetoz:
#### Situation:
`sync-java` library uses chunking and we are limiting to fetch results upto 250 (No pagination used on the results).
Since Inventory entries can be more than 1 for each sku -

Example: 
Lets say we have 3 InventoryEntries with sku = 5299 and with
1. channel = null.
2. channel = 123
3. channel = 234

Then the results for sku = 5299 will be 3 entries.

Similarly, If we are executing query for 250 sku's, if every InventoryEntry has 2 channels and without a channel. Then we should retrieve 250*3 = 750 InventoryEntries. 

Which is not possible in our case and we are missing out few inventory entries, which results in the error while syncing.

#### Complication:
A user reported that they are getting the below error,

"errors\":[{\"code\":\"DuplicateField\",\"message\":\"A duplicate value '\\\"5299\\\"' exists for field 'sku'.\",\"duplicateValue\":\"5299\",\"field\":\"sku\"}]}]

Issue link: commercetools/commercetools-project-sync#301

#### Resolution

This PR fixes the problems described above.

- Tests
    - [x] Unit 
    - [x] Integration
- [x] Documentation (javadocs)
- [X] Add Release Notes entry.

#### Hints for Review
- Core of the change is changing the query mechanism on the inventory fetching, this part is separated to a class already: InventoryEntryQueryBuilder
- Query requires to know the id of a supply channel before the fetch, so InventorySync is refactored to make that possible, so before fetching the matched resources from target project, the sync first do reference resolution. (note it's not a performance issue as the sync populatingthe key to id map with populateKeyToIdCachesForReferencedKeys before the reference resolution.


src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+    inventoryEntryQueries.add(getInventoryEntryQuery(queryBuilder, limit));
+    return inventoryEntryQueries;
+  }
+
+  private static InventoryEntryQuery getInventoryEntryQuery(
+      @Nonnull final StringBuilder queryBuilder, final int limit) {
+    // drop " or " keyword in the end of the predicate.
+    String queryString = queryBuilder.substring(0, queryBuilder.length() - 4);
+    // clear/reset the queryBuilder
+    queryBuilder.setLength(0);
```
ahalberkamp, line 70: I would reset the query-builder in line 51

src/main/java/com/commercetools/sync/inventories/InventorySync.java
```diff
@@ -164,16 +149,63 @@ public InventorySync(@Nonnull final InventorySyncOptions syncOptions) {
...
-   *
-   * @param oldInventories inventory entries from CTP
-   * @param inventoryEntryDrafts drafts that need to be synced
-   * @return a future which contains an empty result after execution of the update
-   */
+  private CompletableFuture<Set<InventoryEntryDraft>> resolveReferences(
+      @Nonnull final Set<InventoryEntryDraft> inventoryEntryDrafts) {
+
+    final List<CompletionStage<InventoryEntryDraft>> collect =
+        inventoryEntryDrafts.stream().map(this::resolveReferences).collect(toList());
```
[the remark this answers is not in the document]
ahmetoz: This is not possible anymore as I added a line to https://github.com/commercetools/commercetools-sync-java/blob/4c4a3b7dc5a7da2b4995ae30a566284e91bcbaa8/src/main/java/com/commercetools/sync/commons/utils/CompletableFutureUtils.java#L124

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
ahmetoz: good point: https://github.com/commercetools/commercetools-sync-java/pull/757/commits/fe73e1f37fc19d58f3f4bc91c2cc46d60773063c

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryIdentifier.java
```diff
@@ -13,24 +13,24 @@
 
 /**
  * This class provides as a container of the unique identifier of an {@link InventoryEntry} for the
- * sync which is a combination of both the SKU of the inventory entry and the supply channel key of
+ * sync which is a combination of both the SKU of the inventory entry and the supply channel id of
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-sync-java/pull/757/commits/ba66846173b6ef092255dea708c8a19d94d586cd

ahalberkamp (approved):

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+      }
+
+      queryBuilder.append("(");
+      queryBuilder.append(predicate);
+      queryBuilder.append(") or ");
+      limit++;
+    }
+
+    // add last one:
+    inventoryEntryQueries.add(getInventoryEntryQuery(queryBuilder, limit));
```
butenkor, line 64: How do you know that you always need it? Why not check if your queryBuilder is not empty instead?

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        inventoryEntryQueries.add(getInventoryEntryQuery(queryBuilder, limit));
+
+        // clear/reset the queryBuilder and limit
+        queryBuilder.setLength(0);
+        limit = 0;
+      }
+
+      queryBuilder.append("(");
+      queryBuilder.append(predicate);
+      queryBuilder.append(") or ");
```
butenkor, line 59: I guess instead of always adding it and then always doing a substring removal below you could add "or" as the first line after the `for` statement, i.e:

```
for (InventoryEntryIdentifier identifier : identifiers) {
 if (inventoryEntryQueries.size() > 0) {
  queryBuilder.append(") or ");
}
```

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+
+    final List<InventoryEntryQuery> inventoryEntryQueries = new ArrayList<>();
+    StringBuilder queryBuilder = new StringBuilder();
+    int limit = 0;
+    for (InventoryEntryIdentifier identifier : identifiers) {
+
+      String predicate = buildQueryPredicate(identifier).toSphereQuery();
+
+      if (queryBuilder.length() + predicate.length() > MAX_QUERY_LENGTH) {
+        inventoryEntryQueries.add(getInventoryEntryQuery(queryBuilder, limit));
```
butenkor, line 50: It seem that the limit should be counted up after `buildQueryPredicate` statement? For example if `identifiers` array has 500 SKUs and we would fit in into a single query max 250 items then it would query for 250 skus with limit of 249. Also, how do we make sure that the limit of 500 items per requests is guaranteed since otherwise we run into 409 error?

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
ahmetoz: I think the code flow ensures it, that queryBuilder is not empty, isn't it ?  https://github.com/commercetools/commercetools-sync-java/blob/fe73e1f37fc19d58f3f4bc91c2cc46d60773063c/src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java#L57-L60

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
ahmetoz: I could not get the relation with  `inventoryEntryQueries.size() > 0`
if I add an or for the first identifier, then the remaining ones would not have it ? 

In the end this comment might be only related to coding style and don't think it will effect much on readability or performance of the code.

src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java
ahmetoz: > It seem that the limit should be counted up after buildQueryPredicate statement?

it's before because you don't want to exceed the limit when the new query is added. limit case does not seem reasonable as it's only increases when the query is added to query builder

>Also, how do we make sure that the limit of 500 items per requests is guaranteed since otherwise we run into 409 error?

size of the MAX_QUERY_LIMIT ensures a query could have max 238 sub query, the min is depend on the size of the buildQueryPredicate.
In this case the limits are based on the query and can be calculated dynamically with sizes.

https://github.com/commercetools/commercetools-sync-java/blob/fe73e1f37fc19d58f3f4bc91c2cc46d60773063c/src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java#L17-L22

maybe the 10000/42 calculation in comment might be improved, because I figured out I'd not say in comment why it's 42, 
min size of a subquery such as: `(sku="a" and supplyChannel is not defined)` (string length = 42) but I think this could be guessed from the code comment from below as the queries will be sth like below:

https://github.com/commercetools/commercetools-sync-java/blob/fe73e1f37fc19d58f3f4bc91c2cc46d60773063c/src/main/java/com/commercetools/sync/inventories/helpers/InventoryEntryQueryBuilder.java#L28-L30

PR #817, opened 2022-03-21
Fix ProductSync: resource identifier equal checks on AddToCategory and RemoveFromCategory actions && Release.

### LINKED ISSUE #816: Product sync always produces AddToCategory and RemoveFromCategory actions, which lead to errors
**Describe the bug**
Product sync always produces AddToCategory and RemoveFromCategory actions. This causes errors, as the AddToCategory action tries to add the product to a category that it is already in.

This is caused by the following code:

https://github.com/commercetools/commercetools-sync-java/blob/1625e2a5d94bb3865234fd9ae1d5ffabf204b18e/src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java#L179-L188

Calling `toResourceIdentifier` is not enough, as it returns a self reference. The following `equals` call will therefore call `Reference.equals`, which is not compatible to a `ResourceIdentifer`. 

**To Reproduce**
Just try to sync a product (which has a category).

**Expected behavior**
No UpdateActions regarding categories unless any changed.

ahmetoz:
Resolves #816 

Additionally PR fixes the dependency issues and adds some changes for the github action steps for the dependency updates + next release preparation.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -12,6 +12,14 @@
     {
       "matchPackageNames": ["com.github.ben-manes.caffeine:"],
       "allowedVersions": "< 3.0.0"
+    },
+    {
+      "matchPackageNames": ["org.ajoberstar.git-publish:"],
+      "allowedVersions": "< 4.0.0"
+    },
+    {
+      "matchPackageNames": ["org.ajoberstar.grgit:"],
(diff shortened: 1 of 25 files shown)


praveenkumarct (approved):
Good work here !!

PR #840, opened 2022-08-07
Migrate the minimum support to Java 11.

praveenkumarct:
#### Summary
Upgrade Java version minimum support from 8 to 11.

#### Description
We are missing updates from different libraries If we are still using Java 8 as minimum version, Hence upgraded to next stable version 11 as minimum version.

#### Relevant Issues
- https://github.com/commercetools/commercetools-sync-java/pull/834


leungkinghin (approved):
It looks good to me

Issue #841, opened 2022-08-10
ShoppingListSync callbacks (error, warning) does not pass draft objects


vasanth-mahendran:
**Describe the bug**
com.commercetools.sync.shoppinglists.ShoppingListSync handleError method does not have  parameter for draft object, hence callback method does not receive the draft object reference. 

**To Reproduce**
1. Sync draft objects with ShoppingListSync
2. callback does not receive draft objects on error

**Expected behavior**
Callback should receive draft objects on error

**Additional context**
As part of our process, we are tracking error records which requires draft objects to be passed in error callback

lojzatran:
Hi @vasanth-mahendran 

Thank you for reporting this issue. It is indeed missing in the callbacks. We will try to address it as the next task.

leungkinghin:
Has been fixed and released in https://github.com/commercetools/commercetools-sync-java/blob/master/docs/RELEASE_NOTES.md#902---sep-21-2022

PR #850, opened 2022-09-15
Add error callback arguments

leungkinghin:
#### Summary
Missing old resource, new draft and updateActions as arguments in error callback.

#### Description
When there is an exception during sync, a method called handleError is invoked. Users can supply an error callback and their callback will be called inside that method once error happens. However, error callback does not receive old resource, new draft and updateActions in ShoppingListSync, CustomerSync and CartDiscountSync. 

#### Relevant Issues
https://github.com/commercetools/commercetools-sync-java/issues/841

#### Todo
- [x] Add new arguments in error callback and `handleError` methods in `ShoppingListSync`, `CustomerSync` and `CartDiscountSync`.
- [x] Create new unit test case in `ShoppingListSyncTest`, `CustomerSyncTest` and `CartDiscountSyncTest`

### DIFF
--- a/src/main/java/com/commercetools/sync/commons/BaseSync.java
+++ b/src/main/java/com/commercetools/sync/commons/BaseSync.java
@@ -146,9 +146,9 @@ public abstract class BaseSync<
       @Nullable final Throwable exception,
       final S oldResource,
       final T newResourceDraft,
-      List<UpdateAction<S>> updateActions,
+      final List<UpdateAction<S>> updateActions,
       final int failedTimes) {
-    SyncException syncException =
+    final SyncException syncException =
         exception != null
             ? new SyncException(errorMessage, exception)
             : new SyncException(errorMessage);
--- a/src/main/java/com/commercetools/sync/commons/BaseSync.java
+++ b/src/main/java/com/commercetools/sync/commons/BaseSync.java
@@ -140,6 +140,7 @@ public abstract class BaseSync<
    * @param failedTimes The number of times that the failed cart discount statistic counter is
    *     incremented.
    */
+  @SuppressWarnings("unchecked")
   protected void handleError(
       @Nonnull final String errorMessage,
       @Nullable final Throwable exception,
--- a/src/main/java/com/commercetools/sync/commons/BaseSync.java
+++ b/src/main/java/com/commercetools/sync/commons/BaseSync.java
@@ -1,13 +1,18 @@
 package com.commercetools.sync.commons;
 
+import com.commercetools.sync.commons.exceptions.SyncException;
 import com.commercetools.sync.commons.helpers.BaseSyncStatistics;
 import io.sphere.sdk.client.ConcurrentModificationException;
+import io.sphere.sdk.commands.UpdateAction;
+import io.sphere.sdk.models.Versioned;
 import java.util.List;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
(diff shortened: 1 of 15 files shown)


src/main/java/com/commercetools/sync/shoppinglists/ShoppingListSync.java
```diff
@@ -329,9 +356,14 @@ private CompletionStage<Void> applyCallbackAndCreate(
         .orElseGet(() -> CompletableFuture.completedFuture(null));
   }
 
-  private void handleError(@Nonnull final SyncException syncException, final int failedTimes) {
+  private void handleError(
```
lojzatran, line 359: Please make this method general.

lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/850/commits/1f42e8a88ab662a10548d9e5d31c7554ad7fa9c8

lojzatran:
I checked the coverage, but I'm not sure where exactly is the coverage difference. Everything seems to be green.

src/main/java/com/commercetools/sync/commons/BaseSync.java
```diff
@@ -122,4 +127,32 @@ protected static <S> S executeSupplierIfConcurrentModificationException(
...
+   */
+  @SuppressWarnings("unchecked")
+  protected void handleError(
+      @Nonnull final String errorMessage,
+      @Nullable final Throwable exception,
+      final S oldResource,
+      final T newResourceDraft,
+      List<UpdateAction<S>> updateActions,
+      final int failedTimes) {
+    SyncException syncException =
```
praveenkumarct, line 151:
```suggestion
    final SyncException syncException =
```

src/main/java/com/commercetools/sync/products/ProductSync.java
```diff
@@ -547,24 +552,14 @@ private CompletionStage<Void> applyCallbackAndCreate(@Nonnull final ProductDraft
...
-    handleError(errorMessage, null, oldResource, newResource, null);
-  }
-
   /**
    * Given a {@link String} {@code errorMessage} and a {@link Throwable} {@code exception}, this
    * method calls the optional error callback specified in the {@code syncOptions} and updates the
-   * {@code statistics} instance by incrementing the total number of failed products to sync.
+   * {@code statistics} instance by incrementing the total number of failed products to sync. <br>
+   * <br>
+   * NOTE: This method is similar to {@link com.commercetools.sync.commons.BaseSync#handleError}. It
```
praveenkumarct, line 560: Can we change the name of this method?
I don't think it is overriding from baseSync. So instead of using super.handleError() everywhere in the code. We can rename this method here and use it when required.
WDYT?

lojzatran: Can you suggest the name too? 🙏

src/main/java/com/commercetools/sync/commons/BaseSync.java
lojzatran: Should be effectively final 😅 https://github.com/commercetools/commercetools-sync-java/pull/850/commits/879ca67ed21444c1152bed3dac4dc37ab7326662

src/main/java/com/commercetools/sync/products/ProductSync.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/850/commits/6e5977279f117dfd9356016bc7cd0de613c371b2

praveenkumarct (approved):
Good work !!

PR #859, opened 2022-10-26
Adapt SDK-v2 compatible layer

leungkinghin:
**Description**
Adaption of SDK-v2 compatible layer to create sphere client.

**Hints for Review**
- The new Sphere Client instantiation is implemented in `ClientV2ConfigurationUtils`
- `ProductSyncWithSdkV2IT` is cloned from `ProductSyncIT`, with sphere client created by SDK-v2 compatible layer
- `ProductSyncWithSdkV2Benchmark` is cloned from `ProductSyncBenchmark`, with sphere client created by SDK-v2 compatible layer.

- Build with Benchmark : https://github.com/commercetools/commercetools-sync-java/pull/859/commits/cef3f46694c27ea4b758d339715b05eb6186e594


src/integration-test/java/com/commercetools/sync/integration/commons/utils/SphereClientUtils.java
```diff
@@ -19,6 +21,18 @@ public class SphereClientUtils {
   public static final SphereClient CTP_SOURCE_CLIENT = createClient(CTP_SOURCE_CLIENT_CONFIG);
   public static final SphereClient CTP_TARGET_CLIENT = createClient(CTP_TARGET_CLIENT_CONFIG);
```
lojzatran, line 22: So these clients won't be replaced by the v2?

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
```diff
@@ -14,6 +16,49 @@
 
 public final class ClientConfigurationUtils {
 
+  public static final String CLIENT_VERSION_V1 = "CLIENT_VERSION_V1";
+  public static final String CLIENT_VERSION_V2 = "CLIENT_VERSION_V2";
```
lojzatran, line 20: Why do you need these?

lojzatran (changes requested):
I see the resolution like this:

```
Until full migration is completed use SDK v2 compatibility layer/wrapper which reduces the migration efforts drastically and release it.
```

Does it mean we completely replace the client v1 with v2 compat layer? What is the idea here?

gradle-scripts/spotless.gradle
```diff
@@ -1,3 +1,7 @@
+repositories {
```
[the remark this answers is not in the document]
leungkinghin: Because of the gradle issue in these days
https://github.com/diffplug/spotless/issues/1390

It seems resolved in gradle side. I would try to remove the `mavenCentral` and see how the build would be.

src/integration-test/java/com/commercetools/sync/integration/commons/utils/SphereClientUtils.java
```diff
@@ -19,6 +21,18 @@ public class SphereClientUtils {
...
 
+  public static final String CTP_TARGET_PROJECT_KEY = CTP_TARGET_CLIENT_CONFIG.getProjectKey();
```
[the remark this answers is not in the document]
leungkinghin: No. At this moment we still keep old `SphereClient` to avoid breaking change. That's why our integration test still make use of old `SphereClient`.  Regarding the test of new `SphereClient`, I have discussed with @butenkor, we can simply add one integration test and benchmark test in ProductSync to test it.

src/integration-test/java/com/commercetools/sync/integration/commons/utils/SphereClientUtils.java
leungkinghin: It won't be. At this moment we still keep old `SphereClient` to avoid breaking change. That's why our integration test still make use of old `SphereClient`.  Regarding the test of new `SphereClient`, I have discussed with @butenkor, we can simply add one integration test and benchmark test in ProductSync to test it.

src/main/java/com/commercetools/sync/commons/utils/ClientConfigurationUtils.java
leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/859/commits/40ad6f992c7a67629613fee9de935190532e65f2

lojzatran:
LGTM 🥇 Can you add to JIRA what you wrote in the comments since it's not clear what is the resolution of the task.

gradle-scripts/spotless.gradle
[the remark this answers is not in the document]
leungkinghin: It has been fixed in gradle side
I removed the mavenCentral assignment and it builds successfully now.
https://github.com/commercetools/commercetools-sync-java/pull/859/commits/ecba302c0acb747eadc013ab8f40a60ff983361b

lojzatran (approved):

PR #868, opened 2022-11-25
Run benchmark test on pull request

leungkinghin:
- Copy the benchmark test execution from CD to CI, so that it will be executed in every commit.
- Add boolean flag in env-var to indicate whether benchmarks.json should be modified.
- Update the new benchmark threshold for every resources

### DIFF
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -10,7 +10,8 @@ pre-defined threshold, so that we know if the corresponding commit influences th
 
 3. Benchmark results are written as JSON to 
 [benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json) only when a 
-release is made or the build has a git tag. The JSON file is saved in the gh-pages branch of the repo.
+new release is made. The JSON file is saved in the [gh-pages branch](https://github.com/commercetools/commercetools-sync-java/tree/gh-pages) 
+of the repo.
 
 4. The average results in [benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json)
 are used to display the results in the form of this [graph](https://commercetools.github.io/commercetools-sync-java/benchmarks/). 
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -5,12 +5,12 @@
 1. Benchmarks are run by JUnit as a separate source set just like main, test and integration-test. The benchmarks can be 
 found [here](/src/benchmark/java/com/commercetools/sync/benchmark).
 
-2. Every benchmark writes it's results as JSON to 
-[benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json) which is saved in
-the gh-pages branch of the repo.
+2. Every time a commit is made in any branch, the benchmarks are run and the run shows whether the benchmarks exceed the
+pre-defined threshold, so that we know if the corresponding commit influences the performance.
 
-3. Every time a release is made or the build has a git tag, the benchmarks are run and the results are added to 
-[benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json).
+3. Benchmark results are written as JSON to 
+[benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json) only when a 
+release is made or the build has a git tag. The JSON file is saved in the gh-pages branch of the repo.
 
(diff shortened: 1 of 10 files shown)


src/benchmark/java/com/commercetools/sync/benchmark/BenchmarkUtils.java
```diff
@@ -44,10 +47,11 @@ final class BenchmarkUtils {
   static void saveNewResult(
       @Nonnull final String sync, @Nonnull final String benchmark, final double newResult)
       throws IOException {
-
-    final JsonNode rootNode = new ObjectMapper().readTree(getFileContent());
-    final JsonNode withNewResult = addNewResult(rootNode, sync, benchmark, newResult);
-    writeToFile(withNewResult.toString());
+    if (SUBMIT_BENCHMARK_RESULT) {
```
lojzatran, line 50: More logical would be to put the if condition in the caller. When I call this method, I would expect it to save new results every time.

src/benchmark/java/com/commercetools/sync/benchmark/ProductSyncBenchmark.java
```diff
@@ -60,6 +60,10 @@ class ProductSyncBenchmark {
   private List<String> errorCallBackMessages;
   private List<String> warningCallBackMessages;
   private List<Throwable> errorCallBackExceptions;
+  private static final int PRODUCT_BENCHMARKS_CREATE_ACTION_THRESHOLD =
+      30_000; // 150% of the maximum value (20K)
```
lojzatran, line 64: Why is it sometimes 200%, sometimes 150% of the maximum value? Please explain or unify this.

lojzatran (changes requested):
What about this point `Discuss the performance effect of this change and decide with considering tradeoffs.`? What is the build time before and now?

What about `Update docs/BENCHMARK.md according to changes.`?

src/benchmark/java/com/commercetools/sync/benchmark/ProductSyncBenchmark.java
leungkinghin: Some resources have a threshold with 200% of highest value because the current benchmark already exceeded 150%. That's why I keep the threshold unchanged. Meanwhile some resources keep the benchmark at low level, that's why I adjusted the threshold a bit (as discussed) to 150%.

Anyway, let me unify them to 200% as used to be.

src/benchmark/java/com/commercetools/sync/benchmark/BenchmarkUtils.java
leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/868/commits/4426a2b4eaf895cbed477c288bcea1e884f55ed1

src/benchmark/java/com/commercetools/sync/benchmark/ProductSyncBenchmark.java
leungkinghin: > What about this point `Discuss the performance effect of this change and decide with considering tradeoffs.`? What is the build time before and now?
> 
> What about `Update docs/BENCHMARK.md according to changes.`?

Currently benchmarks test elapse about 7-8 minutes. That means the entire build time would be 7-8 minutes longer than it used to be. I think it is acceptable and it helps developers to trace the performance change for every commit.

The documentation has been modified here
https://github.com/commercetools/commercetools-sync-java/pull/868/commits/398f735c6db05c425165a4453721f37309776fe7

docs/BENCHMARKS.md
```diff
@@ -5,12 +5,12 @@
...
-[benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json) which is saved in
-the gh-pages branch of the repo.
+2. Every time a commit is made in any branch, the benchmarks are run and the run shows whether the benchmarks exceed the
+pre-defined threshold, so that we know if the corresponding commit influences the performance.
 
-3. Every time a release is made or the build has a git tag, the benchmarks are run and the results are added to 
-[benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json).
+3. Benchmark results are written as JSON to 
+[benchmarks.json](https://commercetools.github.io/commercetools-sync-java/benchmarks/benchmarks.json) only when a 
+release is made or the build has a git tag. The JSON file is saved in the gh-pages branch of the repo.
```
lojzatran, line 13: What's the difference between release is made and the build has a git tag? How do you make a release?

You can also link to the branch `gh-pages`.

leungkinghin: https://github.com/commercetools/commercetools-sync-java/pull/868/commits/74c54b584bcef747a68db33bd78dd41ac975b5f0

PR #918, opened 2023-03-15
Issue 447 master variant switch

lojzatran:
#### Summary

This PR fixes the issue that was reported in project-sync: https://github.com/commercetools/commercetools-project-sync/issues/447

However, after fixing the reported issue, there is a new issue that comes up and the test case is still failing. The next issue is that to add a new variant with the same SKU as the master variant, the following has to be done:
1. temporarily rename the existing master SKU (I choose to add a `-temp` suffix to the SKU)
2. add the new variant
3. change master to the new variant
4. remove the old master variant

Q: Why does it add new variant with the same SKU instead of updating the existing variant?
A: The matching of variant is NOT done by SKU, but by key. In this case, the keys of the variants are different. However, their SKUs are the same. Therefore we have this issue.

Q: Will the temporary SKU be saved in the product?
A: No. It will be applied only in memory when processing the requests.

Q: Do we need to rename the temporary SKU back once the master is switched?
A: No. It is replaced by another variant that has the same SKU, but different key. As we always remove the temporary variant afterwards, there is no need to rename back.

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncMultipleVariantsIT.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncMultipleVariantsIT.java
@@ -8,6 +8,7 @@ import static com.commercetools.sync.integration.commons.utils.SphereClientUtils
 import static com.commercetools.sync.products.ProductSyncMockUtils.*;
 import static com.commercetools.tests.utils.CompletionStageUtil.executeBlocking;
 import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThat;
 
 import com.commercetools.sync.commons.asserts.statistics.AssertionsForStatistics;
 import com.commercetools.sync.commons.exceptions.SyncException;
@@ -16,9 +17,13 @@ import com.commercetools.sync.products.ProductSync;
 import com.commercetools.sync.products.ProductSyncOptions;
 import com.commercetools.sync.products.ProductSyncOptionsBuilder;
 import com.commercetools.sync.products.helpers.ProductSyncStatistics;
+import io.sphere.sdk.products.Product;
 import io.sphere.sdk.products.ProductDraft;
 import io.sphere.sdk.products.ProductProjection;
+import io.sphere.sdk.products.ProductProjectionType;
+import io.sphere.sdk.products.ProductVariant;
 import io.sphere.sdk.products.commands.ProductCreateCommand;
+import io.sphere.sdk.products.queries.ProductProjectionByIdGet;
 import io.sphere.sdk.producttypes.ProductType;
 import java.util.ArrayList;
 import java.util.List;
@@ -39,6 +44,8 @@ public class ProductSyncMultipleVariantsIT {
   private List<Throwable> errorCallBackExceptions;
 
   private ProductSyncOptions syncOptions;
+
+  private Product product;
 
   /**
    * Delete all product related test data from the target project. Then creates for the target CTP
@@ -67,7 +74,7 @@ public class ProductSyncMultipleVariantsIT {
                 PRODUCT_KEY_1_MULTIPLE_VARIANTS_RESOURCE_PATH, productType.toReference())
             .build();
 
-    executeBlocking(CTP_TARGET_CLIENT.execute(ProductCreateCommand.of(productDraft)));
+    product = executeBlocking(CTP_TARGET_CLIENT.execute(ProductCreateCommand.of(productDraft)));
   }
 
   @AfterAll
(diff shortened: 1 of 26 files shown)


src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
```diff
@@ -838,6 +848,45 @@ public static List<UpdateAction<Product>> buildChangeMasterVariantUpdateAction(
         });
   }
 
+  private static List<UpdateAction<Product>> buildSetAttributeInAllVariantsAction(
```
salander85, line 851:
```suggestion
  private static List<UpdateAction<Product>> buildSetAttributeInAllVariantsUpdateAction(
```

salander85, line 851: A comment would be nice here.

src/test/java/com/commercetools/sync/products/utils/ProductUpdateActionUtilsTest.java
```diff
@@ -98,7 +98,10 @@ void buildVariantsUpdateActions_updatesVariants() {
     final Map<String, AttributeMetaData> attributesMetaData = new HashMap<>();
     final AttributeMetaData priceInfo =
         AttributeMetaData.of(AttributeDefinitionBuilder.of("priceInfo", null, null).build());
+    final AttributeMetaData sizeAttributeMetaData =
```
salander85, line 101: why is this change necessary here?

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncMultipleVariantsIT.java
```diff
@@ -0,0 +1,117 @@
...
+    final ProductDraft productDraft =
+        createProductDraftBuilder(
+                PRODUCT_KEY_2_MULTIPLE_VARIANTS_RESOURCE_PATH, productType.toReference())
+            .build();
+
+    final ProductSync productSync = new ProductSync(syncOptions);
+    final ProductSyncStatistics syncStatistics =
+        executeBlocking(productSync.sync(singletonList(productDraft)));
+
+    AssertionsForStatistics.assertThat(syncStatistics).hasValues(1, 0, 1, 0, 0);
```
salander85, line 90: Check the attributes / sku of master variant here.

salander85 (changes requested):

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/918/files#diff-c23a4d1680c961a86cbb2611de8a67213974d2b8bfd49f999013e04e39de40f7R497-R499

lojzatran:
@salander85 please review this commit: https://github.com/commercetools/commercetools-sync-java/pull/918/commits/27213aa3e5fab80b9553d5150ac9dd6955c809f4. It fixes the other problem to make the test pass.

src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncMultipleVariantsIT.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/918/commits/7912b61cfb31fd078bab2032e6bb9206ae525950

src/main/java/com/commercetools/sync/products/utils/ProductUpdateActionUtils.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/918/commits/7625c7146d7fdd2f61342c8aa87a1cfbb3cf6aaa

butenkor:
> A: The matching of variant is NOT done by SKU, but by key. In this case, the keys of the variants are different. However, their SKUs are the same. Therefore we have this issue.

I believe that in that case you need to apply `-temp` suffix to both SKU and the key and later undo it. 

> A: No. It is replaced by another variant that has the same SKU, but different key. As we always remove the temporary variant afterwards, there is no need to rename back.

By replacing the original master variant it might mean that you lose the data it had which was set by another process. User could use beforeUpdateCallback to ensure that the data is not unset (by discarding particular update actions)

lojzatran:
> I believe that in that case you need to apply `-temp` suffix to both SKU and the key and later undo it.

I don't think we need to add suffix to the key. The keys are completely different in draft and in the CTP product. That's why the sync cannot find existing variant and it creates `AddVariant` update action instead of just updating the attributes.

salander85 (approved):

PR #921, opened 2023-03-21
Update ci workflow

lojzatran:
#### Summary
I disabled running CI (tests) when creating a new tag on master branch. It is redundant.

Additionally I changed the README files to v9.2.3 so that it matches the current release.


lojzatran:
CI workflow doesn't work as expected.

PR #928, opened 2023-04-01
Java sdk v2 product batch validator

lojzatran:
#### Summary

I migrated unit test class for product batch validator. The implementation was already migrated before. Additionally I also did some renaming.


src/main/java/com/commercetools/sync/sdk2/customobjects/helpers/CustomObjectCompositeIdentifier.java
```diff
@@ -0,0 +1,139 @@
...
+          + "between the container and key.";
+
+  private CustomObjectCompositeIdentifier(
+      @Nonnull final String key, @Nonnull final String container) {
+    this.key = key;
+    this.container = container;
+  }
+
+  /**
+   * Given a {@link io.sphere.sdk.customobjects.CustomObjectDraft}, creates a {@link
```
salander85, line 33:
```suggestion
   * Given a {@link  com.commercetools.api.models.custom_object.CustomObjectDraft}, creates a {@link
```

salander85: Please check and replace in all comments.

src/main/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidator.java
```diff
@@ -192,8 +191,7 @@ private void collectReferencedKeysInAttributes(
         getReferencedKeysWithReferenceTypeId(variantDraft, ProductTypeReferenceImpl.PRODUCT_TYPE));
```
salander85, line 191: why not using the interface here:
```suggestion
        getReferencedKeysWithReferenceTypeId(variantDraft, ProductTypeReference.PRODUCT_TYPE));
```

src/main/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidator.java
```diff
@@ -192,8 +191,7 @@ private void collectReferencedKeysInAttributes(
...
 
     referencedKeys.customObjectCompositeIdentifiers.addAll(
-        getReferencedKeysWithReferenceTypeId(
-            variantDraft, CustomObjectReferenceImpl.KEY_VALUE_DOCUMENT));
+        getReferencedKeysWithReferenceTypeId(variantDraft, CustomObject.referenceTypeId().name()));
```
salander85, line 194:
```suggestion
        getReferencedKeysWithReferenceTypeId(variantDraft, CustomObjectReference.KEY_VALUE_DOCUMENT));
```

src/test/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidatorTest.java
```diff
@@ -0,0 +1,540 @@
...
+  }
+
+  @Test
+  void getReferencedProductKeys_WithNullAttributes_ShouldReturnEmptySet() {
+    assertThat(getReferencedProductKeys(ProductVariantDraftBuilder.of().build())).isEmpty();
+  }
+
+  @Test
+  void getReferencedProductKeys_WithNoAttributes_ShouldReturnEmptySet() {
+    assertThat(getReferencedProductKeys(ProductVariantDraftBuilder.of().attributes().build()))
```
salander85, line 74: Not sure if this test is the same as the one before? What about adding `emptyList()` of attributes instead?

src/test/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidatorTest.java
```diff
@@ -0,0 +1,540 @@
...
+            "foo", getProductReferenceWithId("foo"), getProductReferenceWithId("bar"));
+    final Attribute productReferenceAttribute =
+        AttributeBuilder.of().name("foo").value(getProductReferenceWithId("foo")).build();
+
+    final List<Attribute> attributes =
+        asList(null, productReferenceAttribute, productReferenceSetAttribute);
+
+    final ProductVariantDraft variantDraft =
+        ProductVariantDraftBuilder.of().attributes(attributes).build();
+    assertThat(getReferencedProductKeys(variantDraft)).containsExactlyInAnyOrder("foo", "bar");
```
salander85, line 99: I wonder if this test is correct. 
You mock a ProductReference with ID calling `getProductReferenceWithId("bar")` and then you assert that `getReferencedProductKeys(variantDraft)` contains the IDs but not the keys...

salander85 (approved):

src/test/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidatorTest.java
lojzatran: Yes it is the same: https://github.com/commercetools/commercetools-sync-java/blob/89a4a0f194fb0e1666ffc9720c7cbf845f84c640/src/test/java/com/commercetools/sync/products/helpers/ProductBatchValidatorTest.java#L77

src/test/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidatorTest.java
lojzatran: The test is correct. Indeed the naming is a bit weird. The old test is the same: https://github.com/commercetools/commercetools-sync-java/blob/89a4a0f194fb0e1666ffc9720c7cbf845f84c640/src/test/java/com/commercetools/sync/products/helpers/ProductBatchValidatorTest.java#L91

It seems like the problem is in the method [`getReferencedProductKeys`](https://github.com/commercetools/commercetools-sync-java/blob/89a4a0f194fb0e1666ffc9720c7cbf845f84c640/src/main/java/com/commercetools/sync/products/helpers/ProductBatchValidator.java#L248) - this method gets the [referenced product IDs](https://github.com/commercetools/commercetools-sync-java/blob/89a4a0f194fb0e1666ffc9720c7cbf845f84c640/src/main/java/com/commercetools/sync/products/helpers/ProductBatchValidator.java#L282), not the keys. So it seems like a wrong naming here.

As I was not involved in the development of this tool, I don't why it is named like this. However, this is a public method, so we should not change the name as this could be a breaking change for a customer.

src/main/java/com/commercetools/sync/sdk2/customobjects/helpers/CustomObjectCompositeIdentifier.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/928/commits/a1c4729eff2306e52b861bc81501348bbbe42a54

src/main/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidator.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/928/commits/67914a7483999a27f7d4e91b491c1b02bf299be9

src/main/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidator.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/928/commits/67914a7483999a27f7d4e91b491c1b02bf299be9

src/test/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidatorTest.java
salander85: Hm, wonder if `ProductVariantDraftBuilder.of().attributes().build()` produces the same as `ProductVariantDraftBuilder.of().build()` ? Then it would be better to do `ProductVariantDraftBuilder.of().attributes(emptyList()).build()`?

salander85 (approved):

src/test/java/com/commercetools/sync/sdk2/products/helpers/ProductBatchValidatorTest.java
lojzatran: I'm not sure what do you mean here. The old test and the new test are exactly the same. Do you mean that `ProductVariantDraftBuilder` is now internally changed and we should also adapt our tests?

PR #936, opened 2023-04-11
Add VariantReferenceResolutionUtils and its tests

lojzatran:
**Summary**

I added VariantReferenceResolutionUtils and related tests. This is the second step to migrate ProductReferenceResolver.


salander85 (approved):
Needs to be merged with base https://github.com/commercetools/commercetools-sync-java/tree/java-sdk-v2-product-sync-migration and differences should be checked!!

PR #942, opened 2023-04-18
Migrate WaitingToBeResolved class and tests

lojzatran:
#### Summary

Migrate 1 class and its tests. They're necessary for ProductSync.


src/main/java/com/commercetools/sync/sdk2/commons/BaseSync.java
```diff
@@ -27,7 +29,7 @@ protected BaseSync(@Nonnull final U statistics, @Nonnull final V syncOptions) {
    *     instance of {@code U} which is a subclass of {@link BaseSyncStatistics} representing the
```
salander85, line 29: Please also update the generics in the comments.

salander85 (changes requested):

src/main/java/com/commercetools/sync/sdk2/commons/BaseSync.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/942/commits/394405f3780b2c758a49607183f3e63176938c99

salander85 (approved):

PR #947, opened 2023-04-20
Migrate product sync and ITs to java sdk v2

lojzatran:
#### Summary

Main changes: https://github.com/commercetools/commercetools-sync-java/pull/947/files#diff-98a7b2c5b9b1ea946cc5fc3ded767a6a5395f36a81b18a8017f24bea79fd0e14,  https://github.com/commercetools/commercetools-sync-java/pull/947/files#diff-d8e12c9e202ba3b3ac24469897c21c425021fa437481fddc3004f8f62927672f

## Errors fixed in this PR
1) Fixed this error:
```
{
  "statusCode" : 400,
  "message" : "Request body does not contain valid JSON.",
  "errors" : [ {
    "code" : "InvalidJsonInput",
    "message" : "Request body does not contain valid JSON.",
    "detailedErrorMessage" : "masterVariant -> prices -> channel: ResourceIdentifier requires an 'id' xor a 'key' but was id:3313a01a-fdad-49ff-9cda-1f4e8e0b4764 and key:missingKey"
  } ]
}
```
The problem was when using ResourceIdentifier we can set either `key` or `id`, but not both. https://github.com/commercetools/commercetools-sync-java/pull/947/commits/dd113f17c57ca1b809d52a85d0ab3d7b96102efe

2) When `categoryOrderHints` is an empty object in the product, it is represented as null in Java object. Therefore we have to check to avoid NPE.

3) I renamed the method `createObjectFromResource` to `readObjectFromResource`. Even though the name `createObjectFromResource` might be slightly better, it was impossible for me to find this method by name when I'm migrating the old code. In the old sdk we have a method called `readObjectFromResource`. So whenever I migrate the code, I expect a method with the same name exists for sdk v2. Lesson learnt: always keep the same name for the common methods so that they're easy to find.

---

## Problem I could not fix

In the old test there was an integration test that verifies that [in the method `ProductSync.fetchAndUpdate()`](https://github.com/commercetools/commercetools-sync-java/blob/1c0aee38fac10a1bcf34d16f6749b226900fead1/src/main/java/com/commercetools/sync/sdk2/customers/CustomerSync.java#L290) whenever sync fails to update with CT error, the sync would fail with a correct error message etc. This was achieved in the past mocking the [execute method with ProductUpdate params](https://github.com/commercetools/commercetools-sync-java/blob/35c0df8928b0ecd68d8fc4c84d7e1da8d42fc747/src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncIT.java#L586). The other methods that fetches from CT were not mocked and were really calling CT. This is not possible anymore with the new sdk v2. To update a product now we have to call sth like `products().update().execute()`. So that means we have to mock `products()`, `update()` and `execute()`. However, when we mock `products()`, we also return the mock for all previous fetches that uses this endpoint in the sync (e.g. `products().withKey().get().execute()`), so we have to mock them too. So these kind of tests where you mock only the calls you need and the rest are real method calls are not possible anymore.

---

The rest I had to adapt to ProductSync.java class and its integration tests working. In tests I changed create methods to ensure methods because they were failing too much with Concurrency Exception. Afterwards the tests were not failing anymore.


salander85 (approved):
LGTM!

PR #954, opened 2023-04-26
Migrate ITs for customergroup service

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-63


src/integration-test/java/com/commercetools/sync/integration/sdk2/services/impl/CustomerGroupServiceImplIT.java
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Deletes customer group from the target CTP projects, then it populates the project with test
+   * data.
+   */
+  @BeforeEach
+  void setup() {
+    deleteCustomerGroups(CTP_TARGET_CLIENT);
+    warnings = new ArrayList<>();
+    oldCustomerGroup =
+        createCustomerGroup(CTP_TARGET_CLIENT, CUSTOMER_GROUP_NAME, CUSTOMER_GROUP_KEY);
```
lojzatran, line 36: Please use `ensureCustomerGroup()` instead and try not to delete custom groups after every test suit run. If we could eliminate these deletions, tests would become more stable.

lojzatran (approved):

PR #964, opened 2023-05-04
DEVX-75 add generic to update action utils

lojzatran:
#### Summary
JIRA: https://commercetools.atlassian.net/browse/DEVX-75

I added correct generics to the common asset util so that we could reuse it in category as well.


salander85 (approved):
Good work

PR #972, opened 2023-05-15
Migrate CategoryReferenceResolver and unittest

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-81


lojzatran (changes requested):
There is another class that was not migrated. Please check and migrate it/create a task to migrate it later: 

https://github.com/commercetools/commercetools-sync-java/blob/master/src/test/java/com/commercetools/sync/products/helpers/productreferenceresolver/CategoryReferenceResolverTest.java

lojzatran (approved):

PR #981, opened 2023-05-19
DEVX-88 Migrate ProductTypeTransformUtils

lojzatran:
#### Summary

JIRA: https://commercetools.atlassian.net/browse/DEVX-88

### DIFF
--- a/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/ProductTypeTransformUtils.java
+++ b/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/ProductTypeTransformUtils.java
@@ -1,0 +1,118 @@
+package com.commercetools.sync.sdk2.producttypes.utils;
+
+import static com.commercetools.sync.sdk2.producttypes.utils.ProductTypeReferenceResolutionUtils.mapToProductTypeDrafts;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
+
+import com.commercetools.api.client.ProjectApiRoot;
+import com.commercetools.api.models.product_type.AttributeDefinition;
+import com.commercetools.api.models.product_type.AttributeNestedType;
+import com.commercetools.api.models.product_type.AttributeSetType;
+import com.commercetools.api.models.product_type.AttributeType;
+import com.commercetools.api.models.product_type.ProductType;
+import com.commercetools.api.models.product_type.ProductTypeDraft;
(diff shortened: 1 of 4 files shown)


src/main/java/com/commercetools/sync/sdk2/producttypes/utils/ProductTypeTransformUtils.java
```diff
@@ -0,0 +1,118 @@
...
+
+      return loadNestedProductTypeReferenceKeys(productTypes)
+          .thenApply(ignore -> mapToProductTypeDrafts(productTypes, referenceIdToKeyCache));
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> loadNestedProductTypeReferenceKeys(
+        @Nonnull final List<ProductType> productTypes) {
+
+      final Set<String> setOfTypeIds = new HashSet<>(collectNestedReferenceIds(productTypes));
```
salander85, line 81: What about the direct references? `AttributeReferenceType`

src/test/java/com/commercetools/sync/sdk2/producttypes/utils/ProductTypeTransformUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    // assertion
+    assertThat(productTypeDrafts)
+        .containsExactly(
+            ResourceToDraftConverters.toProductTypeDraft(productTypeFoo),
+            ResourceToDraftConverters.toProductTypeDraft(productTypeBar));
+  }
+
+  @Test
+  void transform_ShouldReplaceProductTypeNestedAttributeReferenceIdsWithKeys() {
```
salander85, line 104: Add test to check the transformation of direct references.

salander85 (changes requested):

src/main/java/com/commercetools/sync/sdk2/producttypes/utils/ProductTypeTransformUtils.java
lojzatran: The direct references are tested also in the test `transform_ShouldReplaceProductTypeNestedAttributeReferenceIdsWithKeys`. The test is testing set of the direct references, so in the end they're testing what you wrote here. But for better clarity I also added a test that does not have set of references, but only a direct reference: https://github.com/commercetools/commercetools-sync-java/pull/981/commits/89f1f3969caed914b8aacd34c88b98a2f7ae1ea8

salander85 (approved):

PR #983, opened 2023-05-25
DEVX-89 migrate enum value update action utils classes

lojzatran:
#### Summary

JIRA: https://commercetools.atlassian.net/browse/DEVX-90, https://commercetools.atlassian.net/browse/DEVX-89


lojzatran:
I will try to increase the coverage as there are some codes that were not tested before.

salander85 (approved):

PR #987, opened 2023-05-31
Add helper to convert  customField values to JsonNode

salander85:
**Problem**: To calculate updateActions for custom-fields the fields values of a Resource and a Draft needs to be comparable. As in jvm-sdk-v1 all fields values have been of type JsonNode, with java-sdk-v2 has changed to Object type. 
**Solution**: Convert all fields values to JsonNode and compare like before.

JIRA: https://commercetools.atlassian.net/browse/DEVX-169

### DIFF
--- a/src/main/java/com/commercetools/sync/sdk2/commons/utils/CustomUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/sdk2/commons/utils/CustomUpdateActionUtils.java
@@ -331,7 +331,6 @@ public final class CustomUpdateActionUtils {
 
     final String oldCustomTypeId = oldCustomFields.getType().getId();
 
-    // Convert Object to JSONNode
     final Map<String, Object> oldCustomFieldsJsonMap = oldCustomFields.getFields().values();
     final String newCustomTypeId = newCustomFields.getType().getId();
     final FieldContainer fields = newCustomFields.getFields();
@@ -499,9 +498,10 @@ public final class CustomUpdateActionUtils {
     return newCustomFields.keySet().stream()
         .filter(
             newCustomFieldName -> {
-              final Object newCustomFieldValue =
+              // Convert Object to JSONNode
+              final JsonNode newCustomFieldValue =
                   convertFieldValuesToJsonNode(newCustomFields.get(newCustomFieldName));
-              final Object oldCustomFieldValue =
+              final JsonNode oldCustomFieldValue =
                   convertFieldValuesToJsonNode(oldCustomFields.get(newCustomFieldName));
               return !isNullJsonValue(newCustomFieldValue)
                   && !Objects.equals(newCustomFieldValue, oldCustomFieldValue);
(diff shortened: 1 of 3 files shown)


src/main/java/com/commercetools/sync/sdk2/commons/utils/CustomUpdateActionUtils.java
```diff
@@ -329,7 +331,7 @@ List<ResourceUpdateActionT> buildNonNullCustomFieldsUpdateActions(
 
     final String oldCustomTypeId = oldCustomFields.getType().getId();
 
-    // Object is JsonNode
+    // Convert Object to JSONNode
```
lojzatran, line 334: Where do you convert? Maybe move this comment closer to the place where you do the conversion?

src/main/java/com/commercetools/sync/sdk2/commons/utils/CustomUpdateActionUtils.java
```diff
@@ -497,9 +499,10 @@ List<ResourceUpdateActionT> buildNewOrModifiedCustomFieldsUpdateActions(
     return newCustomFields.keySet().stream()
         .filter(
             newCustomFieldName -> {
-              final Object newCustomFieldValue = newCustomFields.get(newCustomFieldName);
-              final Object oldCustomFieldValue = oldCustomFields.get(newCustomFieldName);
-              // TODO: This comparison doesn't work when newFieldValue is passed as JSON string.
+              final Object newCustomFieldValue =
```
lojzatran, line 502:
```suggestion
              final JsonNode newCustomFieldValue =
```

lojzatran: Also for `isNullJsonValue()` you can accept `JsonNode` instead of `Object` and remove the type checking from there.

lojzatran (changes requested):
Is there a test that checks the old field is of e.g. type `Boolean=true` and the new field is of type `JsonNode.instance.boolean(true)` and that they are equal?

salander85:
> Is there a test that checks the old field is of e.g. type `Boolean=true` and the new field is of type `JsonNode.instance.boolean(true)` and that they are equal?

There are some tests already, hidden in some ITs. I will add more to make it clear.

src/main/java/com/commercetools/sync/sdk2/commons/utils/CustomUpdateActionUtils.java
salander85: [a5de830](https://github.com/commercetools/commercetools-sync-java/pull/987/commits/a5de8303e2ce7084a05ce7343a1dd0baac0528ff)

src/main/java/com/commercetools/sync/sdk2/commons/utils/CustomUpdateActionUtils.java
salander85: [a5de830](https://github.com/commercetools/commercetools-sync-java/pull/987/commits/a5de8303e2ce7084a05ce7343a1dd0baac0528ff)

lojzatran:
Good job 🥳

lojzatran (approved):

PR #989, opened 2023-06-02
DEVX-172 Migrate ProductTypeActionUtils and all related classes

lojzatran:
#### Summary

JIRA: https://commercetools.atlassian.net/browse/DEVX-172

#### Issues:
1. In unit tests we had an invalid case when we want to [change attribute constraint from SAME_FOR_ALL to COMBINATION UNIQUE](https://github.com/commercetools/commercetools-sync-java/blob/c05bb4c47b11ca4357c099dc21dfa08e55d6cf8b/src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java#L356). This is not allowed by the platform and the SDK does not have a method for this. Thus there is no way to compile such code in JVM SDK v2 and I had to update the tests to verify that it can be changed only to NONE, otherwise it throws an error.
2. There is no default values when creating resources (e.g. product types or attributes) like in the sdk v1. Now the SDK v2 throws error when null is supplied. Therefore the following test is not valid anymore. https://github.com/commercetools/commercetools-sync-java/blob/c05bb4c47b11ca4357c099dc21dfa08e55d6cf8b/src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java#L442
3. Some tests checks if attribute label is changed, but there are couple of tests just with different attribute type ([String](https://github.com/commercetools/commercetools-sync-java/blob/c05bb4c47b11ca4357c099dc21dfa08e55d6cf8b/src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java#L510), [set of strings](https://github.com/commercetools/commercetools-sync-java/blob/c05bb4c47b11ca4357c099dc21dfa08e55d6cf8b/src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java#L532), [set of set of strings](https://github.com/commercetools/commercetools-sync-java/blob/c05bb4c47b11ca4357c099dc21dfa08e55d6cf8b/src/test/java/com/commercetools/sync/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java#L559) etc). Since the attribute type makes no difference for the attribute label change, I removed those redundant tests.

### DIFF
--- a/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
@@ -353,9 +353,16 @@ final class AttributeDefinitionUpdateActionUtils {
               newAttributeConstraint.name()));
     }
 
+    if (AttributeConstraintEnum.NONE.equals(oldAttributeConstraint)
+        && newAttributeConstraint == null) {
+      // attribute constraint is optional and by default attribute constraint is NONE:
+      // https://docs.commercetools.com/api/projects/productTypes#ctp:api:type:AttributeDefinitionDraft
+      return Optional.empty();
+    }
+
     return buildUpdateAction(
-        oldAttributeDefinition.getAttributeConstraint(),
-        newAttributeDefinition.getAttributeConstraint(),
+        oldAttributeConstraint,
+        newAttributeConstraint,
         () ->
             ProductTypeChangeAttributeConstraintActionBuilder.of()
                 .attributeName(oldAttributeDefinition.getName())
--- a/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
@@ -52,7 +52,7 @@ final class AttributeDefinitionUpdateActionUtils {
   static List<ProductTypeUpdateAction> buildActions(
       @Nonnull final AttributeDefinition oldAttributeDefinition,
       @Nonnull final AttributeDefinitionDraft newAttributeDefinitionDraft)
-      throws BuildUpdateActionException {
+      throws UnsupportedOperationException {
 
     final List<ProductTypeUpdateAction> updateActions =
         filterEmptyOptionals(
@@ -338,7 +338,7 @@ final class AttributeDefinitionUpdateActionUtils {
   static Optional<ProductTypeUpdateAction> buildChangeAttributeConstraintUpdateAction(
       @Nonnull final AttributeDefinition oldAttributeDefinition,
       @Nonnull final AttributeDefinitionDraft newAttributeDefinition)
-      throws BuildUpdateActionException {
+      throws UnsupportedOperationException {
 
     final AttributeConstraintEnum newAttributeConstraint =
         newAttributeDefinition.getAttributeConstraint();
@@ -347,7 +347,7 @@ final class AttributeDefinitionUpdateActionUtils {
     if (newAttributeConstraint != null
         && !newAttributeConstraint.equals(oldAttributeConstraint)
         && !AttributeConstraintEnum.NONE.equals(newAttributeConstraint)) {
-      throw new BuildUpdateActionException(
+      throw new UnsupportedOperationException(
           format(
               "Invalid AttributeConstraint update to %s. Only following updates are allowed: SameForAll to None and Unique to None.",
               newAttributeConstraint.name()));
--- a/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
+++ b/src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
@@ -1,0 +1,367 @@
+package com.commercetools.sync.sdk2.producttypes.utils;
+
+import static com.commercetools.sync.sdk2.commons.utils.CommonTypeUpdateActionUtils.buildUpdateAction;
+import static com.commercetools.sync.sdk2.commons.utils.OptionalUtils.filterEmptyOptionals;
+import static com.commercetools.sync.sdk2.producttypes.utils.LocalizedEnumValueUpdateActionUtils.buildLocalizedEnumValuesUpdateActions;
+import static com.commercetools.sync.sdk2.producttypes.utils.PlainEnumValueUpdateActionUtils.buildEnumValuesUpdateActions;
+import static java.lang.String.format;
+import static java.util.Optional.ofNullable;
+
+import com.commercetools.api.models.product_type.AttributeConstraintEnum;
+import com.commercetools.api.models.product_type.AttributeConstraintEnumDraft;
+import com.commercetools.api.models.product_type.AttributeDefinition;
+import com.commercetools.api.models.product_type.AttributeDefinitionDraft;
+import com.commercetools.api.models.product_type.AttributeEnumType;
+import com.commercetools.api.models.product_type.AttributeLocalizedEnumType;
+import com.commercetools.api.models.product_type.AttributePlainEnumValue;
+import com.commercetools.api.models.product_type.AttributeSetType;
+import com.commercetools.api.models.product_type.AttributeType;
+import com.commercetools.api.models.product_type.ProductTypeChangeAttributeConstraintActionBuilder;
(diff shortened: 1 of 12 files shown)


src/test/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,840 @@
...
+  static void setup() {
+    old =
+        AttributeDefinitionBuilder.of()
+            .name("attributeName1")
+            .label(ofEnglish("label1"))
+            .type(AttributeTypeBuilder::textBuilder)
+            .isRequired(false)
+            .attributeConstraint(AttributeConstraintEnum.SAME_FOR_ALL)
+            .inputTip(ofEnglish("inputTip1"))
+            .inputHint(TextInputHint.TextInputHintEnum.SINGLE_LINE)
```
salander85, line 70: This doesn't exist??
```suggestion
            .inputHint(TextInputHint.SINGLE_LINE)
```

src/test/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,840 @@
...
+      assertThat(e)
+          .hasMessage(
+              "Invalid AttributeConstraint update to COMBINATION_UNIQUE. Only following updates are allowed: SameForAll to None and Unique to None.");
+    }
+  }
+
+  @Test
+  void buildActions_WithNewDifferentValues_ShouldReturnActions()
+      throws UnsupportedOperationException {
+    final List<ProductTypeUpdateAction> result = buildActions(old, newDifferent);
```
salander85, line 456: As i see, here and below test is the only usage of `old, newDifferent and newSame`. So, the BeforeAll isn't necessary and instantiating it here would make it easier to read.

src/test/java/com/commercetools/sync/sdk2/producttypes/utils/producttypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java
```diff
@@ -0,0 +1,1516 @@
...
+      buildAttributesUpdateActions_WithNullNewAttributesAndNoOldAttributes_ShouldNotBuildActions() {
+    final ProductType oldProductType = mock(ProductType.class);
+    when(oldProductType.getAttributes()).thenReturn(emptyList());
+    when(oldProductType.getKey()).thenReturn("product_type_key_1");
+
+    final List<ProductTypeUpdateAction> updateActions =
+        buildAttributesUpdateActions(
+            oldProductType,
+            ProductTypeDraftBuilder.of()
+                .key("key")
```
salander85, line 144: Shouldn't this be "product_type_key_1"?

salander85 (changes requested):

src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
```diff
@@ -0,0 +1,374 @@
...
+      return Optional.empty();
+    }
+
+    return buildUpdateAction(
+        oldAttributeConstraint,
+        newAttributeConstraint,
+        () ->
+            ProductTypeChangeAttributeConstraintActionBuilder.of()
+                .attributeName(oldAttributeDefinition.getName())
+                .newValue(AttributeConstraintEnumDraft.NONE)
```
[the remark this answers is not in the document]
lojzatran: I don't want to put the logic of the platform validation here. The only reason I have this check here is because for `changeAttributeConstraint` action I have to supply `AttributeConstraintEnumDraft` class and this class contains only 1 enum - `AttributeConstraintEnumDraft.NONE`. 

**Longer explanation:**
When customer supplies drafts to us, they supply with class `AttributeConstraintEnum`. This class contains 4 enums - NONE, UNIQUE, COMBINATION_UNIQUE and SAME_FOR_ALL. However, when we generate `changeAttributeConstraint` inside java-sync, we have to supply `AttributeConstraintEnumDraft.NONE` only as there is no other enums available for `AttributeConstraintEnumDraft`. Therefore I have to check in this method that user does NOT supply anything else than `AttributeConstraintEnum.NONE`. But the goal was NOT to simulate the platform check here.
If customer supplies `oldAttributeConstraint eq CombinationUnique -> change to NONE` - we would generate an action and it's up to the platform to deal with it.

src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
```diff
@@ -0,0 +1,374 @@
...
+   * Returns an optional containing the attribute type if is an {@link AttributeEnumType} or if the
+   * {@link AttributeType} is a {@link AttributeSetType} with an {@link AttributeEnumType} as a
+   * subtype, it returns this subtype in the optional. Otherwise, an empty optional.
+   *
+   * @param attributeType the attribute type.
+   * @return an optional containing the attribute type if is an {@link AttributeEnumType} or if the
+   *     {@link AttributeType} is a {@link AttributeSetType} with an {@link AttributeEnumType} as a
+   *     subtype, it returns this subtype in the optional. Otherwise, an empty optional.
+   */
+  private static Optional<AttributeEnumType> getAttributeEnumType(
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/989/commits/2c21623910d4b88bdc9c9fb4bf22ad1300b873c1

src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
```diff
@@ -0,0 +1,374 @@
...
+   * AttributeLocalizedEnumType} as a subtype, it returns this subtype in the optional. Otherwise,
+   * an empty optional.
+   *
+   * @param attributeType the attribute type.
+   * @return an optional containing the attribute type if is an {@link AttributeLocalizedEnumType}
+   *     or if the {@link AttributeType} is a {@link AttributeSetType} with an {@link
+   *     AttributeLocalizedEnumType} as a subtype, it returns this subtype in the optional.
+   *     Otherwise, an empty optional.
+   */
+  private static Optional<AttributeLocalizedEnumType> getLocalizedEnumAttributeType(
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/989/commits/2c21623910d4b88bdc9c9fb4bf22ad1300b873c1

src/test/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/989/commits/75b1adf883bf06918a01b0fda85d961c94d8e5da

src/test/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtilsTest.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/989/commits/664dfb5e47b8220a5bc39a2057d542c7a5de5e8f

src/test/java/com/commercetools/sync/sdk2/producttypes/utils/producttypeactionutils/BuildAttributeDefinitionUpdateActionsTest.java
lojzatran: https://github.com/commercetools/commercetools-sync-java/pull/989/commits/6b97efefa1d103aca2f9bbd274ccc99e0d84c77a

src/main/java/com/commercetools/sync/sdk2/producttypes/utils/AttributeDefinitionUpdateActionUtils.java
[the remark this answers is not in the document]
salander85: got it. thanks for the explanation.

PR #995, opened 2023-06-14
Migrate missing helpers and utilities

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-108

**Issue:**
In java-sdk-v2 the `CustomObjectDraft` value and actual `CustomObject` value can be of different type as these fields are of type `Object`. In v1 the value-field has been of type `JsonNode`, so this method isn't valid anymore and had to be adjusted.

https://github.com/commercetools/commercetools-sync-java/blob/master/src/main/java/com/commercetools/sync/customobjects/utils/CustomObjectSyncUtils.java#L19

### DIFF
--- a/src/main/java/com/commercetools/sync/sdk2/customobjects/helpers/CustomObjectBatchValidator.java
+++ b/src/main/java/com/commercetools/sync/sdk2/customobjects/helpers/CustomObjectBatchValidator.java
@@ -62,10 +62,9 @@ public class CustomObjectBatchValidator
 
     if (customObjectDraft == null) {
       handleError(CUSTOM_OBJECT_DRAFT_IS_NULL);
+      return false;
     } else {
       return true;
     }
-
-    return false;
   }
 }
--- a/src/main/java/com/commercetools/sync/sdk2/customobjects/helpers/CustomObjectBatchValidator.java
+++ b/src/main/java/com/commercetools/sync/sdk2/customobjects/helpers/CustomObjectBatchValidator.java
@@ -1,0 +1,71 @@
+package com.commercetools.sync.sdk2.customobjects.helpers;
+
+import static java.util.stream.Collectors.toSet;
+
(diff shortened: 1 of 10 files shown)


src/main/java/com/commercetools/sync/sdk2/customobjects/helpers/CustomObjectBatchValidator.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private boolean isValidCustomObjectDraft(@Nullable final CustomObjectDraft customObjectDraft) {
+
+    if (customObjectDraft == null) {
+      handleError(CUSTOM_OBJECT_DRAFT_IS_NULL);
+    } else {
+      return true;
+    }
+
+    return false;
```
lojzatran, line 69:
```suggestion
    if (customObjectDraft == null) {
      handleError(CUSTOM_OBJECT_DRAFT_IS_NULL);
      return false;
    } else {
      return true;
    }
```

src/test/java/com/commercetools/sync/sdk2/customobjects/utils/CustomObjectSyncUtilsTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+    assertThat(CustomObjectSyncUtils.hasIdenticalValue(oldCustomObject, newCustomObjectdraft))
+        .isFalse();
+  }
+
+  @Test
+  void
+      hasIdenticalValue_WithSameNestedJsonNode_WithSameAttributeOrderInNestedJson_ShouldBeIdentical() {
+
+    JsonNode oldNestedJson =
```
lojzatran, line 191:
```suggestion
    final JsonNode oldNestedJson =
```

lojzatran (approved):
👍

PR #1000, opened 2023-06-20
Create doc for custom-object-sync

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-110

### DIFF
--- a/docs/sdk2/usage/CUSTOM_OBJECT_SYNC.md
+++ b/docs/sdk2/usage/CUSTOM_OBJECT_SYNC.md
@@ -1,0 +1,257 @@
+# Custom Object Sync
+
+Module used for importing/syncing CustomObject into a commercetools project. 
+It also provides utilities for correlating a custom object to a given custom object draft based on the 
+comparison of a [CustomObject](https://docs.commercetools.com/api/projects/custom-objects#customobject) 
+against a [CustomObjectDraft](https://docs.commercetools.com/api/projects/custom-objects#customobjectdraft).
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Usage](#usage)
+  - [Prerequisites](#prerequisites)
+    - [ProjectApiRoot](#projectapiroot)
+    - [Required Fields](#required-fields)
+    - [SyncOptions](#syncoptions)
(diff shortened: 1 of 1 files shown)


docs/sdk2/usage/CUSTOM_OBJECT_SYNC.md
```diff
@@ -0,0 +1,257 @@
...
+
+##### beforeUpdateCallback
+In theory, `CustomObjectSyncOptions` provides callback before update operation. User can customize their own callback and inject
+into sync options. However, in the actual case, `beforeUpdateCallback`is not triggered in the custom object sync process. When
+the new custom object draft has the same key and container as an existing custom object but different in custom object values, 
+the sync process automatically performs the create/update operation. The value of a corresponding custom object in the target project is overwritten. This approach is different from other resources and no update action is involved.
+
+No example is applicable.
+
+##### beforeCreateCallback
```
lojzatran, line 115: This callback also applies for updates, right? If so, please mention it here as well.

docs/sdk2/usage/CUSTOM_OBJECT_SYNC.md
```diff
@@ -0,0 +1,257 @@
...
+ final Logger logger = LoggerFactory.getLogger(CustomObjectSync.class);
+ final CustomObjectSyncOptions customObjectSyncOptions = CustomObjectSyncOptionsBuilder
+         .of(projectApiRoot)
+         .warningCallback((syncException, draft, customObject) -> 
+            logger.warn(new SyncException("My customized message"), syncException)).build();
+````
+
+##### beforeUpdateCallback
+In theory, `CustomObjectSyncOptions` provides callback before update operation. User can customize their own callback and inject
+into sync options. However, in the actual case, `beforeUpdateCallback`is not triggered in the custom object sync process. When
```
lojzatran, line 109: Maybe give a link to the docs that there are no updates available for custom objects: https://docs.commercetools.com/api/projects/custom-objects#create-or-update-customobject

lojzatran (approved):

docs/sdk2/usage/CUSTOM_OBJECT_SYNC.md
salander85: It calls the same API endpoint, but the sync calculates if an update or creation is needed. This callback is only applied when the given customObject doesn't already exist in the project. So, from this perspective it's only in case of custom object creation.

PR #1023, opened 2023-07-04
Migrate helpers for InventorySync

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-133

### DIFF
--- a/src/test/java/com/commercetools/sync/sdk2/inventories/helpers/InventoryReferenceResolverTest.java
+++ b/src/test/java/com/commercetools/sync/sdk2/inventories/helpers/InventoryReferenceResolverTest.java
@@ -1,0 +1,287 @@
+package com.commercetools.sync.sdk2.inventories.helpers;
+
+import com.commercetools.api.client.error.BadRequestException;
+import com.commercetools.api.models.channel.*;
+import com.commercetools.api.models.type.CustomFieldsDraftBuilder;
+import com.commercetools.api.models.type.TypeResourceIdentifierBuilder;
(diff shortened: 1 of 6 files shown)


src/test/java/com/commercetools/sync/sdk2/inventories/helpers/InventoryReferenceResolverTest.java
```diff
@@ -0,0 +1,290 @@
...
+                          BLANK_KEY_VALUE_ON_RESOURCE_IDENTIFIER));
+              return null;
+            })
+        .toCompletableFuture()
+        .join();
+  }
+
+  @Test
+  void
+      resolveSupplyChannelReference_WithResolvedSupplyChannelReference_ShouldNotResolveChannelReference() {
```
lojzatran, line 266: How do you verify that the test does not resolve channel reference? It could resolve channel reference as well as the assertions are exactly the same as for `resolveSupplyChannelReference_WithNonExistingChannelAndEnsureChannel_ShouldResolveSupplyChannelReference`

lojzatran (changes requested):

src/test/java/com/commercetools/sync/sdk2/inventories/helpers/InventoryReferenceResolverTest.java
salander85: You're right, this test doesn't have value. I removed it: [2750321](https://github.com/commercetools/commercetools-sync-java/pull/1023/commits/2750321bafaad23b63e0a69ef62cf5f03281265b)

lojzatran (approved):

PR #1067, opened 2023-08-01
Migrate channel helper and tests

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-213


src/test/java/com/commercetools/sync/sdk2/commons/utils/ChannelCustomUpdateActionUtilsTest.java
```diff
@@ -0,0 +1,80 @@
...
+            action -> {
+              assertThat(action.getType()).isNull();
+              assertThat(action.getFields()).isNull();
+            });
+  }
+
+  @Test
+  void buildSetCustomFieldAction_WithChannelResource_ShouldBuildChannelUpdateAction() {
+    final String customFieldName = "name";
+    final JsonNode customFieldValue = JsonNodeFactory.instance.textNode("foo");
```
lojzatran, line 66:
```suggestion
    final String customFieldValue = "foo";
```

lojzatran (changes requested):
It's weird that `ChannelCustomActionBuilder.java`is not used anywhere in channel sync. As a standalone class I do not see much value. What is the purpose then? If it has no use, I think it's better to remove it and write it to the migration guide.

salander85:
> It's weird that `ChannelCustomActionBuilder.java`is not used anywhere in channel sync. As a standalone class I do not see much value. What is the purpose then? If it has no use, I think it's better to remove it and write it to the migration guide.

@lojzatran So, i reverted the changes and add a note to the migration guide.

PR #1072, opened 2023-08-10
DEVX-219 migrate leftover tests part 1

lojzatran:
#### Summary

JIRA: https://commercetools.atlassian.net/browse/DEVX-219


src/test/java/com/commercetools/sync/sdk2/commons/utils/ProductAssetCustomUpdateActionUtilsTest.java
```diff
@@ -59,7 +59,7 @@ void buildRemoveCustomTypeAction_WithProductAsset_ShouldBuildProductUpdateAction
     ProductSetAssetCustomTypeAction productSetAssetCustomTypeAction =
         (ProductSetAssetCustomTypeAction) updateAction;
     assertThat(productSetAssetCustomTypeAction.getAssetId()).isEqualTo(null);
-    assertThat(productSetAssetCustomTypeAction.getAssetKey()).isEqualTo(null);
+    assertThat(productSetAssetCustomTypeAction.getAssetKey()).isEqualTo(assetKey);
```
lojzatran, line 62: This test is wrong because it has no assetId nor assetKey. According to the docs, one of these must be present: https://docs.commercetools.com/api/projects/products#set-asset-custom-type. So I made the fix and also fixed where the action is being built.

PR #1083, opened 2023-08-28
Migrate tests for nested attribute references on products

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-222

### DIFF
--- a/src/integration-test/java/com/commercetools/sync/integration/commons/utils/ProductTypeITUtils.java
+++ b/src/integration-test/java/com/commercetools/sync/integration/commons/utils/ProductTypeITUtils.java
@@ -528,6 +528,17 @@ public final class ProductTypeITUtils {
               assertThat(enumValue.getLabel()).isEqualTo(enumValueDraft.getLabel());
             });
   }
+    @Nonnull
+    public static Map<String, Object> createNestedAttributeValueReferences(
+            @Nonnull final String attributeName, @Nonnull final Object referenceValue) {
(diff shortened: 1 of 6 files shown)


src/integration-test/java/com/commercetools/sync/integration/commons/utils/ProductTypeITUtils.java
```diff
@@ -529,5 +511,17 @@ private static void assertLocalizedEnumsValuesAreEqual(
             });
   }
 
+  @Nonnull
+  public static Attribute createNestedAttributeValueReferences(
+      @Nonnull final String attributeName, @Nonnull final Object referenceValue) {
+    return AttributeBuilder.of().name(attributeName).value(referenceValue).build();
```
lojzatran, line 517: I see you changed [in this PR](https://github.com/commercetools/commercetools-sync-java/pull/1083/commits/ce0189aabe11d7bf2facb8659528772ab2aa001a) from Map to AttributeBuilder. Is it the only way we support? If so, we should document this.

salander85: Ok, i found an issue with this change. So, i have to make it work for Map too.

lojzatran: If it's too hard to support both ways, you could also document this as a restriction and support only AttributeBuilder.

PR #1088, opened 2023-09-05
Reference resolution of attributes references

salander85:
JIRA: https://commercetools.atlassian.net/browse/DEVX-215

Contains the code corrections incl. test adjustments for reference resolution of attribute references.
Includes migrated IT for product-sync:
```
src/integration-test/java/com/commercetools/sync/integration/externalsource/products/ProductSyncWithReferencedProductsInAnyOrderIT.java

```


src/main/java/com/commercetools/sync/sdk2/commons/utils/ResourceIdentifierUtils.java
```diff
@@ -23,5 +29,28 @@ public static boolean isReferenceOfType(
         && referenceValue.getTypeId().getJsonName().equals(referenceTypeId);
   }
 
+  /**
+   * Given a {@link JsonNode} {@code referenceValue} which is the JSON representation of CTP
+   * Reference object, this method checks if it is has a {@code typeId} with the value equal to
```
lojzatran, line 34:
```suggestion
   * Reference object, this method checks if it has a {@code typeId} with the value equal to
```

src/main/java/com/commercetools/sync/sdk2/products/utils/AttributeUtils.java
```diff
@@ -1,95 +1,28 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import io.vrap.rmf.base.client.utils.json.JsonUtils;
 import java.util.List;
-import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 
 public final class AttributeUtils {
 
   /**
    * Given an attribute this method checks if its value is of type "reference" or "set" of
```
lojzatran, line 14: How does it know if it's a set or not?

src/main/java/com/commercetools/sync/sdk2/products/utils/AttributeUtils.java
```diff
@@ -1,95 +1,28 @@
...
-          nestedSet.stream().flatMap(Collection::stream).collect(Collectors.toList());
-    } else {
-      return singletonList(attribute);
-    }
-    return flattenedAttributes.stream()
-        .map(attr -> flatMapNestedAttributes(attr))
-        .flatMap(Collection::stream)
-        .collect(Collectors.toList());
+  @Nonnull
+  public static List<JsonNode> getAttributeReferencesAsJson(@Nonnull final Attribute attribute) {
```
lojzatran, line 22: So returning Reference was not a correct way?

lojzatran (changes requested):

src/main/java/com/commercetools/sync/sdk2/products/utils/AttributeUtils.java
salander85: I've splitted the method to make it more clear what's done: [7308732](https://github.com/commercetools/commercetools-sync-java/pull/1088/commits/73087329061ab9452135ae77b863b42240389082)
The converted value has to be set in the attribute to be able to replace the references in async streams.

lojzatran: Maybe I didn't ask clearly. I was wondering why Reference is not used anymore? What is the problem that you changed everywhere to JsonNode instead?

salander85: The problem is, that as attribute value is typed with Object. Coming from different sources, it can be Map, JsonNode, String, Reference ... To overcome the complexity checkin all those different types i've changed to JsonNode.

lojzatran (approved):

PR #1116, opened 2023-11-14
Fix transitive dependency

lojzatran:
#### Summary

Make commercetools-java-sdk-v2 available as transitive dependency.


salander85 (approved):
Good work.

PR #1124, opened 2023-12-04
Fix NPE in Inventory entry reference resolution

salander85:
#### Summary
https://commercetools.atlassian.net/browse/DEVX-275

#### Description
To avoid NPE thrown by InventoryEntryDraftBuilder i've added a check for required fields and created empty draft in case of missing required (which will fail in sync() with proper error message because it'll be checked by BatchValidator).


lojzatran (approved):
Good job 🥇

PR #1210, opened 2025-03-31
fix: dont report not found object as error when delete

lojzatran:
#### Summary

When we are attempting to delete custom objects and the server returns 404 Not Found, this shouldn‘t be considered as an error as it is already deleted.

Fixes: https://github.com/commercetools/commercetools-project-sync/issues/619


andihalberkamp:
lgtm :-)

src/main/java/com/commercetools/sync/services/impl/UnresolvedReferencesServiceImpl.java
```diff
@@ -154,6 +155,12 @@ public CompletionStage<Optional<WaitingToBeResolvedT>> delete(
                 return Optional.of(
                     OBJECT_MAPPER.convertValue(resource.getBody().getValue(), clazz));
               } else {
+                if (exception instanceof NotFoundException) {
+                  // if we try to delete and the custom object is already gone, then everything is
+                  // fine
```
andihalberkamp, line 160: Maybe remove the while line here

andihalberkamp (approved):

